PR #226, opened 2022-03-24
fix(simulated/series): fill pools taking into account decimals

abylaw:
Closes GEN-772

### DIFF
--- a/pkg/deployments/deploy/simulated/06-series.js
+++ b/pkg/deployments/deploy/simulated/06-series.js
@@ -173,7 +173,7 @@ module.exports = async function () {
 
       const _scale = await adapter.callStatic.scale();
       const scale = _scale.div(1e12).toNumber() / 1e6
-      const _principalInTarget = principalPriceInTarget.div(ethers.BigNumber(10).pow(decimals - 4)) / (10 ** Math.abs(4 - decimals))
+      const _principalInTarget = principalPriceInTarget.div(ethers.BigNumber.from(10).pow(decimals - 4)) / (10 ** Math.abs(4 - decimals))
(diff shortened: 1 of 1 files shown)


jparklev (approved):
ty! looks like the build is failing with this, but otherwise lgtm

Screen Shot 2022-03-24 at 8 37 10 AM

pkg/deployments/deploy/simulated/06-series.js
```diff
@@ -76,10 +73,14 @@ module.exports = async function () {
       const { abi: tokenAbi } = await deployments.getArtifact("Token");
       const pt = new ethers.Contract(ptAddress, tokenAbi, signer);
       const yt = new ethers.Contract(ytAddress, tokenAbi, signer);
+      const decimals = await target.decimals()
+      const one = ethers.BigNumber.from(1).mul(ethers.BigNumber.from(10).pow(decimals))
+      const fourtyThousand =  ethers.BigNumber.from(40_000).mul(ethers.BigNumber.from(10).pow(decimals))
+      const oneMillion = ethers.BigNumber.from(1_000_000).mul(ethers.BigNumber.from(10).pow(decimals))
```
fedealconada, line 79: fun to see the different ways we can implement this hehe --> on the Capital Seeder I've used parseUnits-> https://github.com/sense-finance/sense-v1-autotasks/blob/3b73c3744c83e4d798e5fcd7a96d3991c2571578/utils/data.ts#L134

fedealconada (approved):
LGTM! Thanks for the PR!

pkg/deployments/deploy/simulated/06-series.js
abylaw: oh that looks quite clean! i changed it to your approach hehe

PR #232, opened 2022-03-28
GEN-788: fix(MockAdapter): add scaling on wrap/unwrap fns

fedealconada:
- Fix MockAdapter wrap/unwrap functions
- change some missing renamings (e.g yields -> YTs)
- add remove liquidity test


justfile
```diff
@@ -107,9 +108,15 @@ turbo-test-local-no-fuzz *cmds="": && _timer
...
 
-turbo-test-local-unusual-decimal-val *cmds="": && _timer
+turbo-test-local-greater-decimal-val *cmds="": && _timer
 	cd {{ invocation_directory() }}; export FORGE_MOCK_TARGET_DECIMALS={{ HEX_8 }}; \
-		export FORGE_MOCK_UNDERLYING_DECIMALS={{ HEX_12 }}; forge test \
+		export FORGE_MOCK_UNDERLYING_DECIMALS={{ HEX_6 }}; forge test \
+	    {{ lib-paths-from-pkg-deps }} -vvv --force --root {{ invocation_directory() }} \
+		--optimize --optimize-runs 20 --ffi -m "^test(M(a[^i]|[^a])|[^M])" {{ cmds }}
+
+turbo-test-local-lower-decimal-val *cmds="": && _timer
```
jparklev, line 58: Nice idea 👍 Is this what was causing us to miss the bug in the mock adapter?

pkg/core/src/tests/Periphery.t.sol
```diff
@@ -35,16 +35,16 @@ contract PeripheryTest is TestHelper {
...
     function testSponsorSeries() public {
         uint256 maturity = getValidMaturity(2021, 10);
-        (address pt, address yield) = sponsorSampleSeries(address(alice), maturity);
+        (address pt, address yt) = sponsorSampleSeries(address(alice), maturity);
 
-        // check pt and yield deployed
+        // check pt and yt deployed
         assertTrue(pt != address(0));
-        assertTrue(yield != address(0));
+        assertTrue(yt != address(0));
```
jparklev, line 43: image

pkg/core/src/tests/Periphery.t.sol
```diff
@@ -1243,6 +1258,58 @@ contract PeripheryTest is TestHelper {
...
+            minAmountsOut,
+            0,
+            false
+        );
+
+        uint256 uBalAfter = ERC20(adapter.underlying()).balanceOf(address(bob));
+        assertEq(ERC20(balancerVault.yieldSpacePool()).balanceOf(address(bob)), 0);
+        assertEq(ptBal, ptToBeIssued);
+        assertEq(uBalBefore + underlyingBal, uBalAfter);
+        assertEq(underlyingBal, targetToBeAdded.fmul(lscale) * SCALING_FACTOR);
```
jparklev: Could also check that the target `bob` had before is equal (or close) in value to the tokens he has now (like `targetBeforeAdd` = `currentUnderlyingBal.fdiv(scale)` + `currentPtBal.fmul(targetPerPt)` or something)

pkg/core/src/tests/test-helpers/mocks/MockAdapter.sol
```diff
@@ -88,19 +88,34 @@ contract MockAdapter is CropAdapter {
...
         underlying.transferFrom(msg.sender, address(this), uBal);
-        uint256 mintAmount = uBal.fdivUp(lscale.value, FixedMath.WAD);
+        uint256 mintAmount = uBal.fdivUp(lscale.value);
+        uint256 scalingFactor = 10 **
+            (
+                target.decimals() > underlying.decimals()
+                    ? target.decimals() - underlying.decimals()
+                    : underlying.decimals() - target.decimals()
+            );
+        mintAmount = mintAmount / scalingFactor;
```
jparklev: So we're going from underlying decimals to target decimals here, right? Shouldn't it be:
```
mintAmount * (10^(targetDecimals - underlyingDecimals))

// Which is equivalent to
if (underlyingDecimals <= targetDecimals) {
    mintAmount * (10^(targetDecimals - underlyingDecimals))
} else {
    mintAmount / (10^(underlyingDecimals - targetDecimals))
}
```

Whereas I think `mintAmount = mintAmount / scalingFactor` will always remove decimals, it'll never "upscale"

jparklev: (similar comment for a few of the other conversions)

jparklev (commented):
Nice changes overall and I like that we're improving test coverage as we're going. Just question or two on tests and a comment about conversions

pkg/core/src/tests/test-helpers/mocks/MockAdapter.sol
fedealconada: yeah, you're right! it should check both sides. actually, that was my original plan and why I've added the `lower` `greater` scripts butt I guess then I was biased by our own code haha.
I've just added some tests for unwrap and wrap functions and fix what you've mentioned as well, gonna push in a min :)

pkg/core/src/tests/Periphery.t.sol
fedealconada: Yep, totally. I don't understand why we would add `+ currentPtBal.fmul(targetPerPt)`. The target before I've added liquidity should be ~equal in value to the underlying balance after removing liquidity, right?

fedealconada: nvm, i confused PT with LP, yeah, makes sense!

pkg/core/src/tests/test-helpers/mocks/MockAdapter.sol
fedealconada: just pushed!

pkg/core/src/tests/Periphery.t.sol
jparklev: 👍

pkg/core/src/tests/test-helpers/mocks/MockAdapter.sol
jparklev: Just reviewed the last commits and they LGTM!

justfile
fedealconada: > Nice idea 👍 Is this what was causing us to miss the bug in the mock adapter?

Partly, but the main thing is that we didn't have proper tests to check that wrap and unwrap work as expected. Because the tests were simulating the exact same "formula" than the code uses, so that's why tests were passing. With the addition of those 2 little tests I've added, YES we could have caught the error.

fedealconada:
just appeased linter errors and ready to merge then? @jparklev

jparklev:
> just appeased linter errors and ready to merge then? @jparklev

Sure!

pkg/core/src/tests/test-helpers/mocks/MockAdapter.sol
```diff
@@ -58,7 +59,11 @@ contract MockAdapter is CropAdapter {
...
         )
-    {}
+    {
+        uint256 tDecimals = MockTarget(_target).decimals();
+        uint256 uDecimals = MockTarget(MockTarget(target).underlying()).decimals();
+        scalingFactor = 10**(tDecimals > uDecimals ? tDecimals - uDecimals : uDecimals - tDecimals);
+    }
 
     function scale() external virtual override returns (uint256 _value) {
         if (value > 0) {
```
jparklev, line 69: One thing i've been thinking @fedealconada is that we should simplify the logic here for the mock adapter a little bit. Will submit a little PR if I have the chance to show what I was thinking

PR #201, opened 2022-03-07
add adapter for Targets implementing 4626

jparklev:

PR #233, opened 2022-04-05
chore: bump solmate and pin it to e7deddff

jparklev:
This gets us the [ERC4626.sol](https://github.com/Rari-Capital/solmate/blob/main/src/mixins/ERC4626.sol) mixin, which we use in https://github.com/sense-finance/sense-v1/pull/201


pkg/core/src/tests/Divider.t.sol
```diff
@@ -32,8 +32,8 @@ contract Dividers is TestHelper {
         uint256 maturity = getValidMaturity(2021, 10);
         try alice.doSponsorSeries(address(adapter), maturity) {
             fail();
-        } catch (bytes memory error) {
-            assertEq0(error, arithmeticError);
+        } catch Error(string memory error) {
+            assertEq(error, "TRANSFER_FROM_FAILED");
```
jparklev, line 36: New revert message for solmate's `SafeTransferLib`

pkg/core/src/tests/CAdapter.tm.sol
```diff
@@ -11,7 +11,7 @@ import { CAdapter, CTokenLike, PriceOracleLike } from "../adapters/compound/CAda
 import { BaseAdapter } from "../adapters/BaseAdapter.sol";
 
 import { Assets } from "./test-helpers/Assets.sol";
-import { DSTest } from "./test-helpers/DSTest.sol";
+import { DSTest } from "./test-helpers/test.sol";
```
jparklev, line 14: Changed the name of `DS-Test.sol` to match what foundry expects from the `DS-Test` library, which makes integration with solmate's latest version work (since solmate reads from `test.sol` and we can't remap `DSTest.sol` to `test.sol` on our side). This feels like too much of a monkey patch and we should find a better approach before merging

fedealconada (approved):
lgtm!

jparklev:
OK, just reviewed the commits between the previous solmate version and the commit we're pinning now – changes look small, well-tested, and well-reviewed for the contracts we use. That said, I think
* Whenever another major, audited version is released, we should re-pin to that
* We should consider switching to using Solmate's fixed point math directly now that it has more conveniences like what we're used to (WAD math defaults, denom change passthroughs, gas optimizations)

PR #234, opened 2022-04-08
GEN-821: add fuse adapter

fedealconada:
## Motivation

Sense wants to add the ability for users to deploy Fuse adapters.

## Solution

Created both the Fuse Factory (`FFactory.sol`) and the Fuse Adapter (`FAdapter.sol`).
Even though Fuse uses Compound contracts, we had to take into consideration some differences when implementing these contracts:

- Each Fuse pool uses a different comptroller so `periphery.deployAdapter` should now be able to receive this data. We have decided to take the same approach as in `_flashBorrowAndSwap` where the method receives a `bytes memory data` that it needs to decode. In the case of the Fuse Factory, this will be the comptroller address. Potentially, for future adapters of different kind, it could mean something else that that implementation needs.
- Fuse pools can offer multiple reward tokens: different and multiple reward tokens for each market (target) of each pool. We've created both a `CropsFactory` and a `CropsAdapter` which support multiple reward tokens. Adapters rewarding only one token could opt just to inherit the previous `CropAdapter` which will be more gas efficient than the `CropsAdapter` one. 

Some more context about the multiple rewards adapters:

Crops Adapter
---------------
This contract includes a `address[] memory _rewardTokens` which stores the reward tokens so, when `notify` is called, we can update all the tokens accordingly (behaviour is exactly the same as in the `CropAdapter` except that now we iterate over all the reward tokens).

As the `rewardTokens` could change over time (a protocol starts/stops rewarding some token) we've added a `setRewardsTokens` method to simply replace that array. 

Crops Factory
--------------
Since `cropsAdapter.setRewardsTokens()` is an auth'd method, if the adapter was deployed via a factory, we added a `setRewardTokens` on this contract which acts passthrough to be able to modify the rewards tokens array.

Fuse Adapter
-------------
This contract defines yet another method called `setRewardsTokens` which receives a `_rewardsDistributorsList` apart from the `_rewardsTokens`.  This is because, when calling `_claimRewards` (which is a method that any adapter inheriting from `CropsAdapter ` **must** override) we need iterate over this rewards distributors which are the ones that have the `claimRewards()` function that actually make the claim of rewards on Fuse.

Fuse Factory
-------------
Because of what's mentioned above, we implemented here a another yet another `setRewardTokens` which is a passthrough to be able to call `fadapter.setRewardTokens()`

## Implementer Checklist

- [x]  [OPTIONAL] Integrate the feature into the deployment scripts


pkg/core/src/Periphery.sol
```diff
@@ -120,7 +120,6 @@ contract Periphery is Trust, IERC3156FlashBorrower {
     /// @param target Target to onboard
     function deployAdapter(address f, address target) external returns (address adapter) {
         if (!factories[f]) revert Errors.FactoryNotSupported();
-        if (!AdapterFactory(f).exists(target)) revert Errors.TargetNotSupported();
```
fedealconada, line 123: Removing this from here since I think it's more a responsibility of the Factory to check for this.

pkg/core/src/adapters/BaseAdapter.sol
```diff
@@ -24,82 +24,58 @@ abstract contract BaseAdapter is IERC3156FlashLender {
...
-    /// principal that's set aside for Yield Tokens (e.g. 0.1e18 means that 10% of the principal is reserved).
-    /// @notice If `0`, it means no principal is set aside for Yield Tokens
-    uint256 public immutable tilt;
-
-    /// @notice The number this function returns will be used to determine its access by checking for binary
-    /// digits using the following scheme: <onRedeem(y/n)><collect(y/n)><combine(y/n)><issue(y/n)>
-    /// (e.g. 0101 enables `collect` and `issue`, but not `combine`)
-    uint256 public immutable level;
+    /// @notice adapter params
+    AdapterParams public adapterParams;
```
fedealconada, line 37: We are now passing adapterParams instead of all the params separately (similar to what we do with the Factory)

pkg/core/src/adapters/BaseAdapter.sol
```diff
@@ -195,6 +171,57 @@ abstract contract BaseAdapter is IERC3156FlashLender {
             uint256
         )
     {
-        return (target, stake, stakeSize);
+        return (adapterParams.target, adapterParams.stake, adapterParams.stakeSize);
+    }
+
+    // TODO: decide if we want to add all the following getters
```
fedealconada: I've added all this getters but, as you can see on the TODOs, there are a bunch that are only used on the tests. Still I was wondering if: either we add the ones that are being used on the contracts (not tests) or if you should add them all. We could also not have any of them and just get all of the adapter params but it would be inefficient (gas wise).

pkg/core/src/adapters/BaseFactory.sol
```diff
@@ -18,9 +18,6 @@ abstract contract BaseFactory {
     /// @notice Sense core Divider address
     address public immutable divider;
 
-    /// @notice Protocol's data contract address
-    address public immutable protocol;
```
fedealconada, line 22: Removed `protocol` as factories might need more than one protocol address, like in the case of Fuse that you can pass different comptroller addresses.

pkg/core/src/adapters/BaseFactory.sol
```diff
@@ -40,26 +37,25 @@ abstract contract BaseFactory {
...
+    function deployAdapter(address _target) external virtual returns (address adapter) {
+        revert Errors.NotImplemented();
+    }
 
-    /// @notice Target validity check that must be overriden by child contracts
-    function exists(address _target) external virtual returns (bool);
+    /// @notice Deploys both an adapter and a target wrapper for the given _target and specifies a comptroller
+    /// @param _target Address of the Target token
+    /// @param _protocol Address of the comptroller contract
+    function deployAdapter(address _target, address _protocol) external virtual returns (address adapter) {
```
fedealconada: We now have two `deployAdapter()` functions, one that only receives the target and it will use the hardcoded protocol address (like the `CAdapter` which uses COMPTROLLER) and another one that receives a 2nd param `_protocol` to cover for cases like in Fuse that it needs different comptroller addresses.

fedealconada: Alternatively, we could have only one `deployAdapter` function which receives 2 params: `address target` and `bytes memory data` (same pattern as `Periphery._flashBorrowAndSwap`) that the parent Adapter should `abi.decode`. Not sure about the gas cost of this alternative but one downside is that for some protocol like Compound in which we are always using the same address (which is already hardcoded), every time we have to call `deployAdapter` we would need to send something like `deployAdapter(cDAI_address, "")` (note the 2nd param). No big deal, but just commenting here.

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -10,52 +10,30 @@ import { Divider } from "../Divider.sol";
...
-    uint256 public rewardBal; // last recorded balance of reward token
     uint256 public totalTarget;
     mapping(address => uint256) public tBalance;
-    mapping(address => uint256) public rewarded; // reward token per user
+
+    address[] public rewardTokens;
+    mapping(address => uint256) public shares; // accumulated reward token per collected target per reward token
+    mapping(address => uint256) public rewardedBalances; // last recorded balance of reward token per reward token
+    mapping(address => mapping(address => uint256)) public rewarded; // rewarded token per token per user
+    // TODO: simplify mappings into only one as they share the same key?
```
fedealconada: I should probably do this one

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -10,52 +10,30 @@ import { Divider } from "../Divider.sol";
...
-            _mode,
-            _tilt,
-            _level
-        )
-    {
-        reward = _reward;
+        BaseAdapter.AdapterParams memory _adapterParams,
+        address[] memory _rewardTokens
+    ) Trust(msg.sender) BaseAdapter(_divider, _adapterParams) {
+        rewardTokens = _rewardTokens;
```
fedealconada: This contract now supports having multiple reward tokens. The contract has now a `address[] rewardTokens` array which can be initialised from the parent adapter (e.g `CAdapter`). As a protocol can decide to stop rewarding some given token, we've added the `setRewardsTokens` which allows the owner of the Adapter to modify the `rewardTokens` array. As  `setRewardsTokens` requires trust, we have also made the `CropAdapter` to inherit from `Trust.sol`

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -75,34 +53,43 @@ abstract contract CropAdapter is BaseAdapter {
...
 
     /// @notice Some protocols don't airdrop reward tokens, instead users must claim them.
     /// This method may be overriden by child contracts to claim a protocol's rewards
-    function _claimReward() internal virtual {
+    function _claimRewards() internal virtual {
         return;
     }
 
+    function setRewardTokens(address[] memory _rewardTokens) public requiresTrust {
+        _claimRewards();
```
fedealconada: We make this call so as to make sure that, in the case that the new array contains less reward tokens (we are removing some), we do a last claim rewards on those that will be removed.

pkg/core/src/adapters/CropFactory.sol
```diff
@@ -2,17 +2,25 @@
...
-        FactoryParams memory _factoryParams,
-        address _reward
-    ) BaseFactory(_divider, _protocol, _factoryParams) {
-        reward = _reward;
+    constructor(address _divider, FactoryParams memory _factoryParams)
+        Trust(msg.sender)
+        BaseFactory(_divider, _factoryParams)
+    {}
+
+    // TODO: do we want this to be a function on the Periphery rather than here?
```
fedealconada: thoughts?

pkg/core/src/adapters/compound/CAdapter.sol
```diff
@@ -91,72 +91,51 @@ contract CAdapter is CropAdapter {
...
-    {
-        isCETH = _target == CETH;
-        ERC20(underlying).approve(_target, type(uint256).max);
-        uDecimals = CTokenLike(underlying).decimals();
+        AdapterParams memory _adapterParams,
+        address[] memory _rewardTokens
+    ) CropAdapter(_divider, _adapterParams, _rewardTokens) {
+        // Sanity check: underlying passed must equal to target's underlying
+        address underlying = _adapterParams.target == CETH ? WETH : CTokenLike(_adapterParams.target).underlying();
+        if (_adapterParams.underlying != underlying) revert Errors.InvalidParam(); // TODO: revert or just override with the correrct one?
```
fedealconada: We are doing this sanity check to make sure that the underlying passed in `adapterParams` equals the one of `target.underlying()`. If not, we revert but we could also just replace with the correct one and not revert.
Alternatively, we could just not do this check?

pkg/core/src/adapters/fuse/FAdapter.sol
```diff
@@ -0,0 +1,208 @@
...
+    uint8 public uDecimals;
+
+    constructor(
+        address _divider,
+        address _comptroller,
+        AdapterParams memory _adapterParams,
+        address[] memory _rewardTokens
+    ) CropAdapter(_divider, _adapterParams, _rewardTokens) {
+        // Sanity check: underlying passed must equal to target's underlying
+        // TODO: do we need this sanity check?
```
fedealconada: Same as comment in `CAdapter`

pkg/core/src/adapters/fuse/FAdapter.sol
```diff
@@ -0,0 +1,208 @@
...
+        return _to18Decimals(exRate);
+    }
+
+    function scaleStored() external view override returns (uint256) {
+        // return FTokenLike(adapterParams.target).exchangeRateStored();
+        uint256 exRate = FTokenLike(adapterParams.target).exchangeRateStored();
+        return _to18Decimals(exRate);
+    }
+
+    function _claimRewards() internal virtual override {
```
fedealconada, line 107: When we call `_claimRewards()` it will iterate over all the `rewardDistributors` (gotten from `comptroller.getRewardDistributors()`) and then, on each one of them, we need to check if that distributor is supported for the adapter's target and, if so, call `claimRewards`.

Also note that, in the case that a target now starts to receive a new reward token, we would need to update the `rewardTokens` array that exist on `CropAdapter.sol`.  

I mentioned that we are adding `address[] rewardTokens` array into the `CropAdapter`. We could, alternatively, directly call `comptroller.getRewardDistributors()` (same as what we are doing when we initialise the `FAdapter`). 
- Upside: we always have the updated reward tokens list. 
- Downside: it would be more gas expensive AND if a reward token stops to be supported there could be some amount of tokens that are never distributed to the user. I think the later is the biggest issue. IMHO, this alterrnative is a no go but wanted to comment on it.

pkg/core/src/adapters/fuse/FFactory.sol
```diff
@@ -0,0 +1,68 @@
...
+
+    address public constant WETH = 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2;
+    address public constant FUSE_POOL_DIRECTORY = 0x835482FE0532f169024d5E9410199369aAD5C77E;
+
+    constructor(address _divider, FactoryParams memory _factoryParams) CropFactory(_divider, _factoryParams) {}
+
+    function deployAdapter(address _target, address _comptroller) external override returns (address adapter) {
+        /// Sanity checks
+        /// TODO: any better way to verify this?
+        if (!FusePoolLensLike(FUSE_POOL_DIRECTORY).poolExists(_comptroller)) revert Errors.InvalidParam();
```
fedealconada: Actually, I think we can remove this check. If comptroller does not exist, the next line ([33](https://github.com/sense-finance/sense-v1/pull/234/files#diff-e4b60b47abe1f12334fcb564327f939753e57b45b10469f09eecf8be427a05b1R33)) would revert. Right?

pkg/core/src/adapters/fuse/FFactory.sol
```diff
@@ -0,0 +1,68 @@
...
+        for (uint256 i = 0; i < rewardsDistributors.length; i++) {
+            (, uint32 lastUpdatedTimestamp) = RewardsDistributorLike(rewardsDistributors[i]).marketState(_target);
+            if (lastUpdatedTimestamp > 0)
+                rewardTokens[i] = RewardsDistributorLike(rewardsDistributors[i]).rewardToken();
+        }
+
+        // TODO: check _comptroller is valid
+        BaseAdapter.AdapterParams memory adapterParams = BaseAdapter.AdapterParams({
+            target: _target,
+            underlying: underlying == address(0) ? WETH : underlying, // TODO: are we good witht this check right?
```
fedealconada: This check assumes that an underlying which has the zero address is because it's target is some Fuse ETH token, so we set the underlying to be WETH.

pkg/core/src/tests/Adapter.t.sol
```diff
@@ -299,6 +276,76 @@ contract Adapters is TestHelper {
         assertClose(reward.balanceOf(address(alice)), 96 * 1e18);
     }
 
+    // test multiple rewards
+
+    // TODO:
```
fedealconada: Still have to complete some of these tests below

pkg/core/src/tests/FAdapter.tm.sol
```diff
@@ -0,0 +1,254 @@
...
+        uint256 tBalanceAfter = ERC20(Assets.f18ETH).balanceOf(address(this));
+        uint256 uBalanceAfter = ERC20(Assets.WETH).balanceOf(address(this));
+
+        assertEq(uBalanceAfter, 0);
+        assertEq(tBalanceBefore + wrapped, tBalanceAfter);
+    }
+
+    event ClaimRewards(address indexed owner, uint256 indexed amount);
+
+    // function testMainnetNotify() public {
```
fedealconada: Have to finish this test!

pkg/core/src/tests/test-helpers/interfaces/IAdapter.sol
```diff
@@ -1,30 +0,0 @@
-// SPDX-License-Identifier: AGPL-3.0-only
-pragma solidity 0.8.11;
-
-abstract contract IAdapter {
```
fedealconada, line 4: I'm removing all these interfaces, WDYT? They are not being use.

jparklev: Yes, fine with me to remove them

pkg/core/src/adapters/fuse/FFactory.sol
jparklev: Is that the convention for fuse – if the underlying of the fToken is `0`, then it's taken as `WETH`? Is there documentation for this anywhere?

pkg/core/src/adapters/fuse/FFactory.sol
jparklev: Right, it should revert if calling a contract that doesn't exist, though I think someone could deploy a dummy comptroller that does have a `markets` function. To guard against that case, I think we should have this `poolExists` check. Could we do a mainnet test with valid vs. invalid comptrollers? :)

pkg/core/src/adapters/fuse/FAdapter.sol
jparklev: Yea, thanks for expanding on the context. I agree that it makes sense to use the `rewardTokens` array in crop adapter. Would it make any sense to use it in this `_claimRewards` functions too? (rather than calling `getRewardDistributors`)

pkg/core/src/adapters/fuse/FAdapter.sol
jparklev: I don't think this is necessary. This and the `CAdapter` will mostly be deployed from a factory, right? which won't get this wrong

pkg/core/src/adapters/CropFactory.sol
jparklev: I think I like it here better. It feels like the Periphery shouldn't have to worry about this functionality, which is specific to crop adapters

pkg/core/src/adapters/BaseFactory.sol
jparklev: What if we had each factory only work for one comptroller (so you'd have to deploy a new factory for each different pool). This would enforce this "one protocol per factory" rule – what do you think?

pkg/core/src/adapters/BaseAdapter.sol
jparklev: Yea, better to save gas where we can. I think getters for the functions used publicly, and direct access via the mapping for pieces of data we only need for tests. Does that make sense?

pkg/core/src/adapters/BaseAdapter.sol
jparklev: Could this variable be made immutable? Also do you have an idea of what the gas cost change is with this?

pkg/core/src/adapters/BaseAdapter.sol
```diff
@@ -24,82 +24,58 @@ abstract contract BaseAdapter is IERC3156FlashLender {
...
+        /// @notice Amount to stake at issuance
+        uint256 stakeSize;
+        /// @notice Min maturity (seconds after block.timstamp)
+        uint256 minm;
+        /// @notice Max maturity (seconds after block.timstamp)
+        uint256 maxm;
+        /// @notice 0 for monthly, 1 for weekly
+        uint256 mode;
+        /// @notice Issuance fee
+        uint256 ifee;
```
jparklev: if you run the tests with word packing (i.e. making `ifee` &  `tilt` uint128), does that save any gas now that we're using a struct?

pkg/core/src/adapters/BaseAdapter.sol
```diff
@@ -24,82 +24,58 @@ abstract contract BaseAdapter is IERC3156FlashLender {
...
-    /// (e.g. 0101 enables `collect` and `issue`, but not `combine`)
-    uint256 public immutable level;
+    /// @notice adapter params
+    AdapterParams public adapterParams;
+
+    /* ========== DATA STRUCTURES ========== */
+
+    struct AdapterParams {
+        /// @notice Target token to divide
+        address target;
```
jparklev: I'm not opposed to storing these as contract interfaces anymore FWIW (ie `ERC20 target`) in case you think that'd be better here now

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -75,34 +53,43 @@ abstract contract CropAdapter is BaseAdapter {
             }
         }
 
-        rewarded[_usr] = tBalance[_usr].fmulUp(share, FixedMath.RAY);
+        for (uint256 i = 0; i < rewardTokens.length; i++) {
+            rewarded[rewardTokens[i]][_usr] = tBalance[_usr].fmulUp(shares[rewardTokens[i]], FixedMath.RAY);
```
jparklev: I feel like this adds overhead to the case where there is just one reward token. Should we split out a simpler crop join adapter that's like it was before? Or is it not that much more gas?

pkg/core/src/tests/FFactory.tm.sol
```diff
@@ -0,0 +1,140 @@
...
+        assertEq(FAdapter(adapter).rewardTokens(1), Assets.CRV);
+        assertEq(FAdapter(adapter).rewardTokens(2), address(0));
+
+        uint256 scale = FAdapter(adapter).scale();
+        assertTrue(scale > 0);
+    }
+
+    function testMainnetCantDeployAdapterWithoutComptroller() public {
+        divider.setPeriphery(address(this));
+        try factory.deployAdapter(Assets.f18DAI) {
```
jparklev: FWIW now that we're using foundry exclusively, we could use the `expectRevert` cheat code for these

pkg/core/src/adapters/BaseAdapter.sol
fedealconada: I'll have to check, but I did it this way because keeping all the params separately + the new `address[] rewardTokens` array was triggering the "stack too deep" problem

pkg/core/src/adapters/BaseAdapter.sol
fedealconada: me neither haha, but, should we have it maybe on a separate task to do this on all other contracts to keep consistency? or we just do it here on these adapterParams with `target` and `underlying`?

pkg/core/src/adapters/BaseAdapter.sol
fedealconada: makes sense to me!

pkg/core/src/adapters/BaseFactory.sol
fedealconada: I think we said we didn't want to have 1 factory per each fuse pool but just one that would cover all fuse pools (existing and future ones). We can discuss this on today's call :)

pkg/core/src/adapters/CropAdapter.sol
fedealconada: yeah, it adds some overhead, let me check the gas difference, but I wouldn't be opposed to have a simple crop adapter.

pkg/core/src/adapters/fuse/FAdapter.sol
fedealconada: exactly!

pkg/core/src/adapters/fuse/FAdapter.sol
```diff
@@ -0,0 +1,208 @@
...
+        // TODO: do we need this sanity check?
+        address underlying = (_adapterParams.underlying == address(0) || _adapterParams.underlying == WETH)
+            ? WETH
+            : FTokenLike(_adapterParams.target).underlying();
+        if (_adapterParams.underlying != underlying) revert Errors.InvalidParam(); // TODO: revert or just override with the correrct one?
+
+        comptroller = _comptroller;
+
+        // Initialise rewardTokens by calling getRewardsDistributors() -> rewardToken(). Skip array contains addresses
+        if (_rewardTokens.length > 0) {
```
fedealconada: if we are gonna be having the `FAdapters` deployed via the `FFactory`, we don't need this check `if (_rewardTokens.length > 0) {` and we will always want `rewardTokens` to be initialised by calling  `getRewardsDistributors()` -> `rewardToken()`, right?

pkg/core/src/adapters/fuse/FAdapter.sol
fedealconada: well, the problem here is that `claimRewards()` is a function of the `RewardsDistributor`. So, we can't really iterate over `rewardTokens` array. We could store the distributors on the `FAdapter` and then iterate over that array when `_claimRewards()` is called. We discuss this on today's call as well. 

Context: 
- One `Comptroller` can have 0 to multiple `RewardsDistributor` 
- One `RewardsDistributor` has one Reward Token

pkg/core/src/adapters/fuse/FFactory.sol
fedealconada: I did [these](https://github.com/sense-finance/sense-v1/pull/234/files#diff-fe51a5957219329eec388fcbc5d337789761af4c70d999734e2aed46e406a8afR123) ones

pkg/core/src/adapters/fuse/FFactory.sol
fedealconada: well, i'm not 100% sure, what i've done was to go through all existing comptrollers and i've checked that for each token's underlying that was zero address, the target was some ETH token, i can reach out rari's community and ask there

fedealconada: i've found out that there's an `isCEther()` lol so maybe we could just do 
`underlying: FTokenLike(_target).isCEther() ? WETH : underlying,`

jparklev: Perfect, sounds good

pkg/core/src/adapters/compound/CAdapter.sol
jparklev: Yea, this will mostly be passed in from the factory, which will do things correctly, right? So probably leaning away from this check

pkg/core/src/adapters/BaseAdapter.sol
jparklev: I'm not opposed to changing things incrementally as we touch different parts of the codebase, then maybe having a sweeping issue later. Does that sound reasonable?

fedealconada: sounds good!

pkg/core/src/adapters/BaseFactory.sol
fedealconada: For the record, we've decided to go with the [alternative](https://github.com/sense-finance/sense-v1/pull/234/files#r848656346)

pkg/core/src/adapters/CropAdapter.sol
fedealconada: we've decided to have 2 adapters, a `CropAdapter` which only supports one reward token and a `CropsAdapter` (note the `s`) which supports multiple reward tokens.

pkg/core/src/adapters/fuse/FAdapter.sol
fedealconada: For the record, we've decided to have a `mapping (address => address) public rewardsDistributors; // rewards distributors for reward token` on the `FAdapter`. This way, when `claimRewards()`, we can iterate over the `rewardTokens` array and for each one of them, we will get which is the rewards distributor address (using the mapping) and then we can call `rewardsDistributor.claimRewards()`

pkg/core/src/adapters/BaseAdapter.sol
fedealconada: i've been having issues to run the gas snapshot diff, but just FYI @jparklev :
`testAdapterHasParams`: `3112152 -> 3415001 (+302849)`
`testInitSeriesWeekly`: `5396122 -> 5730584 (+334462)`
`testFuzzIssue`: `3904724 -> 3923587 (+18863)`

pkg/core/src/adapters/BaseAdapter.sol
fedealconada: these are the results with word packing:
`testAdapterHasParams`: `3415001 -> 3414281 (-720)`
`testInitSeriesWeekly`: `5730584-> 5686641 (-43943)`
`testFuzzIssue`: `3923587 -> 3919651 (-3936)`

this is achieved by doing the following changes:
on `BaseFactory.sol`
image

on `BaseAdapter.sol`
image

wdyt @jparklev ?

fedealconada: image

jparklev: So it looks like it costs more to deploy, and less for normal operation with those changes? I see in the factory params that `ifee`, `mode` and `tilt` don't add up to 256 bits – do we get any gains if they do? LGTM though, i'm supportive of this change

fedealconada: well.. I could make one of them bigger so as to fit 256 but I don';t think we will have a number exceeding the 128 for ifee, 16 for mode or 64 for tilt, right? and there's no gain gas wise

pkg/core/src/adapters/BaseFactory.sol
```diff
@@ -31,35 +28,26 @@ abstract contract BaseFactory {
...
+    constructor(address _divider, FactoryParams memory _factoryParams) {
         divider = _divider;
-        protocol = _protocol;
         factoryParams = _factoryParams;
     }
 
     /* ========== REQUIRED DEPLOY ========== */
 
-    /// @notice Deploys both an adapter and a target wrapper for the given _target
+    /// @notice Deploys both an adapter and a target wrapper for the given _target and specifies a comptroller
```
jparklev: Isn't the comptroller specific to compound and fuse?

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -81,7 +56,7 @@ abstract contract CropAdapter is BaseAdapter {
     /// @notice Distributes rewarded tokens to users proportionally based on their `tBalance`
     /// @param _usr User to distribute reward tokens to
     function _distribute(address _usr) internal {
-        _claimReward();
+        _claimRewards();
```
jparklev: What's the reason to add `s` to this one?

pkg/core/src/adapters/CropsFactory.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.11;
+
+// Internal references
+import { CropsAdapter } from "./CropsAdapter.sol";
+import { BaseFactory } from "./BaseFactory.sol";
+import { Trust } from "@sense-finance/v1-utils/src/Trust.sol";
+
+abstract contract CropsFactory is Trust, BaseFactory {
+    // address[] public rewardTokens;
```
jparklev: Vestige?

jparklev (commented):
Just did a quick scan, coming back later for a full review. This one was bigger than we realized, but I'm glad we're doing it now. Also, where was that `roll` and `scale` failing?

pkg/core/src/adapters/BaseFactory.sol
fedealconada: true, removing comment!

pkg/core/src/adapters/CropAdapter.sol
fedealconada: vestige from when I was removing the CropAdapter

fedealconada:
> Just did a quick scan, coming back later for a full review. This one was bigger than we realized, but I'm glad we're doing it now. Also, where was that `roll` and `scale` failing?

[Here](testMainnetNotify)

And the issue with the `ClaimRewards` event is [here](https://github.com/sense-finance/sense-v1/pull/234/files#diff-76bd75ca37488f203c24f8295e1db78cf4b671a8ef13f7cb5a3ff2e45b2c64dcR289)

pkg/core/src/tests/adapters/FAdapter.tm.sol
```diff
@@ -0,0 +1,325 @@
...
+        uint256 maturity = DateTimeFull.timestampFromDateTime(
+            month == 12 ? year + 1 : year,
+            month == 12 ? 1 : (month + 1),
+            1,
+            0,
+            0,
+            0
+        );
+
+        address prankedAddress = 0x8FdD0CF22012a5FEcDbF77eF30d9e9834DC1bf0A;
```
jparklev: Could we add this address to our `AddressBook`? (currently the library named `Assets`)

pkg/core/src/tests/adapters/FAdapter.tm.sol
```diff
@@ -204,10 +204,9 @@ contract FAdapters is FAdapterTestHelper {
         assertEq(tBalanceBefore + wrapped, tBalanceAfter);
     }
 
-    event ClaimRewards(address indexed owner, uint256 indexed amount);
-    event Transfer(address indexed from, address indexed to, uint256 amount);
+    event ClaimRewards(address indexed owner, uint256 amount);
```
jparklev, line 251: @fedealconada you might be mad when you see why this was failing... (second arg isn't indexed)

fedealconada: ALT

jparklev:
> roll

What's the link for the roll one? (takes me to `https://github.com/sense-finance/sense-v1/pull/testMainnetNotify`)

pkg/core/src/tests/adapters/FAdapter.tm.sol
```diff
@@ -0,0 +1,322 @@
...
+
+    event ClaimRewards(address indexed owner, uint256 amount);
+
+    function testMainnetNotifyFAdapter() public {
+        // At block 14603884, Convex Tribe Pool has 4 rewards distributors with CVX, CRV, LDO and FXS reward tokens
+        // asset --> rewards:
+        // FRAX3CRV --> CVX and CRV
+        // cvxFXSFXS-f --> CVX, CRV and FXS
+        // CVX --> no rewards
+        // hevm.roll(14603884); // rolling to a previous block makes the .scale() call to fail.
```
jparklev: When I do this locally I see:
Screen Shot 2022-04-19 at 2 26 31 PM

My suspicion is that this is because one of the compound contracts isn't prepared to deal with a block number going backwards. It only happened when I went far enough back (a few blocks didn't do it, I assume b/c a block number hadn't been stored recently) and it didn't happen when I jumped forward – are your findings the same?

pkg/core/src/adapters/BaseFactory.sol
```diff
@@ -31,35 +28,26 @@ abstract contract BaseFactory {
 
     struct FactoryParams {
         address oracle; // oracle address
-        uint256 ifee; // issuance fee
         address stake; // token to stake at issuance
         uint256 stakeSize; // amount to stake at issuance
         uint256 minm; // min maturity (seconds after block.timstamp)
         uint256 maxm; // max maturity (seconds after block.timstamp)
+        uint128 ifee; // issuance fee
         uint16 mode; // 0 for monthly, 1 for weekly
```
jparklev, line 36: Did you already try making `mode` a `uint64`?

pkg/core/src/adapters/CropsAdapter.sol
```diff
@@ -0,0 +1,106 @@
...
+abstract contract CropsAdapter is Trust, BaseAdapter {
+    using SafeTransferLib for ERC20;
+    using FixedMath for uint256;
+
+    /// @notice Program state
+    uint256 public totalTarget;
+    mapping(address => uint256) public tBalance;
+
+    address[] public rewardTokens;
+    mapping(address => CropMeta) public data; // rewarded token per token per user
```
jparklev: This is the same comment as line 31 – should they be different?

pkg/core/src/adapters/CropsAdapter.sol
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    function setRewardTokens(address[] memory _rewardTokens) public requiresTrust {
+        _claimRewards();
+        rewardTokens = _rewardTokens;
+    }
+
+    /* ========== MODIFIERS ========== */
+
+    modifier onlyDivider() {
```
jparklev: Since it's only used once, maybe we don't need the modifier here

pkg/core/src/adapters/CropsAdapter.sol
```diff
@@ -0,0 +1,106 @@
...
+
+    /// @notice Program state
+    uint256 public totalTarget;
+    mapping(address => uint256) public tBalance;
+
+    address[] public rewardTokens;
+    mapping(address => CropMeta) public data; // rewarded token per token per user
+
+    struct CropMeta {
+        // Accumulated reward token per collected target per reward token
```
jparklev: Do we need the final `per reward token` for these comments? That could just be implicit from the mapping, right?

pkg/core/src/adapters/CropsAdapter.sol
```diff
@@ -0,0 +1,106 @@
...
+    using FixedMath for uint256;
+
+    /// @notice Program state
+    uint256 public totalTarget;
+    mapping(address => uint256) public tBalance;
+
+    address[] public rewardTokens;
+    mapping(address => CropMeta) public data; // rewarded token per token per user
+
+    struct CropMeta {
```
jparklev: Might be in favor of naming this just `Crop`

pkg/core/src/adapters/fuse/FAdapter.sol
```diff
@@ -0,0 +1,205 @@
...
+import { Errors } from "@sense-finance/v1-utils/src/libs/Errors.sol";
+import { CropsAdapter } from "../CropsAdapter.sol";
+
+interface WETHLike {
+    function deposit() external payable;
+
+    function withdraw(uint256 wad) external;
+}
+
+interface FTokenLike {
```
jparklev: Perhaps we could use the CTokenLike from the `CAdapter` (since f tokens are c token _like_)

pkg/core/src/adapters/fuse/FAdapter.sol
```diff
@@ -0,0 +1,205 @@
...
+    /// @notice Get the price of an underlying asset.
+    /// @param target The target asset to get the underlying price of.
+    /// @return The underlying asset price in ETH as a mantissa (scaled by 1e18).
+    /// Zero means the price is unavailable.
+    function getUnderlyingPrice(address target) external view returns (uint256);
+
+    function price(address underlying) external view returns (uint256);
+}
+
+/// @notice Adapter contract for cTokens
```
jparklev:
```suggestion
/// @notice Adapter contract for fTokens
```

pkg/core/src/adapters/fuse/FAdapter.sol
```diff
@@ -0,0 +1,205 @@
...
+    constructor(
+        address _divider,
+        address _comptroller,
+        AdapterParams memory _adapterParams,
+        address[] memory _rewardTokens,
+        address[] memory _rewardsDistributors
+    ) CropsAdapter(_divider, _adapterParams, _rewardTokens) {
+        rewardTokens = _rewardTokens;
+        comptroller = _comptroller;
+        isFETH = _adapterParams.underlying == WETH; // TODO: is this okay?
```
jparklev: We could also use `isCEther` here

pkg/core/src/adapters/fuse/FAdapter.sol
```diff
@@ -0,0 +1,205 @@
...
+        ERC20(_adapterParams.underlying).approve(_adapterParams.target, type(uint256).max);
+        uDecimals = FTokenLike(_adapterParams.underlying).decimals();
+
+        // initialise rewardsDistributors mapping
```
jparklev:
```suggestion
        // Initialize rewardsDistributors mapping
```
(nit, both are correct really)

pkg/core/src/adapters/fuse/FAdapter.sol
```diff
@@ -0,0 +1,205 @@
...
+        for (uint256 i = 0; i < _rewardTokens.length; i++) {
+            rewardsDistributors[_rewardTokens[i]] = _rewardsDistributors[i];
```
jparklev: Confused me a little that one `rewardsDistributors` is an array and the other is mapping (maybe just a naming thing)

pkg/core/src/adapters/fuse/FAdapter.sol
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    function scaleStored() external view override returns (uint256) {
+        uint256 exRate = FTokenLike(adapterParams.target).exchangeRateStored();
+        return _to18Decimals(exRate);
+    }
+
+    function _claimRewards() internal virtual override {
+        for (uint256 i = 0; i < rewardTokens.length; i++) {
+            RewardsDistributorLike(rewardsDistributors[rewardTokens[i]]).claimRewards(address(this));
```
jparklev: I think we should use the [scoped claimRewards function](https://github.com/Rari-Capital/compound-protocol/blob/e33902633808aa4c594fa56049a64c5c7c0a7cff/contracts/RewardsDistributorDelegate.sol#L351)

pkg/core/src/adapters/fuse/FFactory.sol
```diff
@@ -0,0 +1,96 @@
...
+    function deployAdapter(address _target, bytes memory data) external override returns (address adapter) {
+        address comptroller = abi.decode(data, (address));
+
+        /// Sanity checks
+        /// TODO: any better way to verify this?
+        if (!FusePoolLensLike(FUSE_POOL_DIRECTORY).poolExists(comptroller)) revert Errors.InvalidParam();
+        (bool isListed, ) = FComptrollerLike(comptroller).markets(_target);
+        if (!isListed) revert Errors.TargetNotSupported();
+
+        // Initialise rewardTokens by calling getRewardsDistributors() -> rewardToken()
```
jparklev:
```suggestion
        // Initialize rewardTokens by calling getRewardsDistributors() -> rewardToken()
```

pkg/core/src/adapters/fuse/FFactory.sol
```diff
@@ -0,0 +1,96 @@
...
+    address public constant WETH = 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2;
+    address public constant FUSE_POOL_DIRECTORY = 0x835482FE0532f169024d5E9410199369aAD5C77E;
+
+    constructor(address _divider, FactoryParams memory _factoryParams) CropsFactory(_divider, _factoryParams) {}
+
+    function deployAdapter(address _target, bytes memory data) external override returns (address adapter) {
+        address comptroller = abi.decode(data, (address));
+
+        /// Sanity checks
+        /// TODO: any better way to verify this?
```
jparklev: Do you think this way isn't good?

pkg/core/src/adapters/fuse/FFactory.sol
```diff
@@ -0,0 +1,96 @@
...
+        if (!FusePoolLensLike(FUSE_POOL_DIRECTORY).poolExists(comptroller)) revert Errors.InvalidParam();
+        (bool isListed, ) = FComptrollerLike(comptroller).markets(_target);
+        if (!isListed) revert Errors.TargetNotSupported();
+
+        // Initialise rewardTokens by calling getRewardsDistributors() -> rewardToken()
+        address[] memory rewardsDistributors = FComptrollerLike(comptroller).getRewardsDistributors();
+        address[] memory rewardTokens = new address[](rewardsDistributors.length);
+        address[] memory targetRewardsDistributors = new address[](rewardsDistributors.length);
+
+        uint256 valid;
```
jparklev: Is this supposed to be `valId` or "valid"?

pkg/core/src/adapters/fuse/FFactory.sol
```diff
@@ -0,0 +1,96 @@
...
+        );
+    }
+
+    // TODO: do we want this function to receive an array of adapters? I don't think there will be a use case for it since, if
+    // we pass several adapters, it would mean that all those adapters (targets) share the same reward tokens, which would be odd...
+    /// @notice Replace existing reward tokens array with a new one for a given adapter
+    /// @param _adapter address of adapter to update
+    /// @param _rewardTokens array of rewards tokens addresses
+    /// @param _adapter array of adapters to update the rewards tokens on
+    function setRewardTokens(
```
jparklev, line 86: It looks like there's a `setRewardTokens` on the CropsFactory this inherits as well – do we need both? do we need to call both independently?

pkg/core/src/tests/adapters/Adapter.t.sol
```diff
@@ -0,0 +1,98 @@
...
+    function getUnderlyingPrice() external view override returns (uint256) {
+        return 1e18;
+    }
+
+    function doSetAdapter(Divider d, address _adapter) public {
+        d.setAdapter(_adapter, true);
+    }
+}
+
+contract Adapters is TestHelper {
```
jparklev, line 45: Is this a BaseAdapter test suite?

pkg/core/src/tests/adapters/CAdapter.tm.sol
```diff
@@ -1,28 +1,28 @@
...
+import { Assets } from "../test-helpers/Assets.sol";
+import { DSTest } from "../test-helpers/test.sol";
+import { Hevm } from "../test-helpers/Hevm.sol";
+import { DateTimeFull } from "../test-helpers/DateTimeFull.sol";
+import { User } from "../test-helpers/User.sol";
+import { LiquidityHelper } from "../test-helpers/LiquidityHelper.sol";
 
 interface CropAdapterLike {
-    function _claimReward() external;
+    function _claimRewards() external;
```
jparklev:
```suggestion
    function _claimReward() external;
```

pkg/core/src/adapters/BaseAdapter.sol
jparklev: You might not love this suggestion... But, since we're adding a decent amount of gas with this struct, what if we made the _most frequently accessed_ storage variables immutable, and put everything else in the `AdapterParams`? i.e. maybe make `target` & `ifee` immutable, and keep everything else in the struct

jparklev (commented):
- [x]  [FIRST TIME ONLY] Review the [[solcurity](https://github.com/Rari-Capital/solcurity)](https://github.com/Rari-Capital/solcurity) standard
- [x]  Read the PR descriptions and the linear ticket
- [x]  Do a high-level review of the code to confirm that the implementation does what it’s supposed to do
- [x]  Go line-by-line through the implementation and check state transitions – since we don’t have a coverage engine for foundry yet, you can imagine how one might check that each different code path is covered
- [x]  Look at the feature and take some time to think about what tests *you* would write, then confirm that all of those cases are covered by the tests that are present
- [x]  Confirm that the implementer has completed the implementer checklist correctly

pkg/core/src/tests/adapters/FAdapter.tm.sol
```diff
@@ -0,0 +1,322 @@
...
+        assertEq(f18EthAdapter.scale(), wethOut);
+        // Sanity check
+        assertEq(ERC20(Assets.WETH).decimals(), 18);
+
+        // Scale is in 18 decimals when the Underlying has a non-standard number of decimals (6 for USDC) ----
+
+        giveTokens(Assets.f18USDC, ONE_FTOKEN, hevm);
+        ERC20(Assets.f18USDC).approve(address(f18UsdcAdapter), ONE_FTOKEN);
+
+        uint256 usdcOut = f18UsdcAdapter.unwrapTarget(ONE_FTOKEN);
```
jparklev, line 376: Could we try this with an ftoken in a different pool as well? (like the tribe convex pool)

pkg/core/src/tests/adapters/FAdapter.tm.sol
```diff
@@ -0,0 +1,322 @@
...
+        assertEq(f18DaiAdapter.scale(), scale);
+    }
+
+    function testMainnetGetUnderlyingPrice() public {
+        PriceOracleLike oracle = PriceOracleLike(Assets.RARI_ORACLE);
+        uint256 price = oracle.price(Assets.DAI);
+        assertEq(f18DaiAdapter.getUnderlyingPrice(), price);
+    }
+
+    function testMainnetUnwrapTarget() public {
```
jparklev, line 165: Could we make sure there's a wrap -> unwrap test in here (like [this](https://github.com/sense-finance/sense-v1/blob/7acd94be5cd7eb2594cfee4814abfe2d228135be/pkg/core/src/tests/ERC4626Adapter.t.sol#L57) one). Mainly just checking, if we wrap then unwrap, do we end up with the same balances we started with

pkg/core/src/tests/adapters/FAdapter.tm.sol
```diff
@@ -0,0 +1,322 @@
...
+
+        uint256 tBalanceAfter = ERC20(Assets.f18DAI).balanceOf(address(this));
+        uint256 uBalanceAfter = ERC20(Assets.DAI).balanceOf(address(this));
+
+        assertEq(uBalanceAfter, 0);
+        assertEq(tBalanceBefore + wrapped, tBalanceAfter);
+    }
+
+    // test with f18ETH
+    function testMainnetFETHAdapterScale() public {
```
jparklev, line 202: Could we add a test like [this](https://github.com/sense-finance/sense-v1/blob/7acd94be5cd7eb2594cfee4814abfe2d228135be/pkg/core/src/tests/ERC4626Adapter.t.sol#L113) that checks that scale corresponds to the rate from `unwrapTarget`?

fedealconada: Isn't that covered on [this](https://github.com/sense-finance/sense-v1/pull/234/files/c3c439d0dcb6b64863083e96c366d3099d811889#diff-76bd75ca37488f203c24f8295e1db78cf4b671a8ef13f7cb5a3ff2e45b2c64dcR296) one?

jparklev: Yes, I think it is :)

pkg/core/src/tests/adapters/FAdapter.tm.sol
fedealconada: Yeah, my findings are the same as yours...  :/

pkg/core/src/tests/adapters/FAdapter.tm.sol
fedealconada: I can, though this address is a very specific to this test... it's just an address of user that has some `f156FRAX3CRV`

fedealconada: we actually don't need this address, gonna remove it!

pkg/core/src/adapters/fuse/FAdapter.sol
fedealconada: I could rename it to "distributors"?

pkg/core/src/adapters/fuse/FAdapter.sol
fedealconada: I tried that but it seems that it does not exist for fuse targets... e.g take a look at the [FXS rewards distributor](https://etherscan.io/address/0x30E9A1Bc6A6a478fC32F9ac900C6530Ad3A1616F#writeContract)

pkg/core/src/adapters/fuse/FFactory.sol
fedealconada: Nah, it was the first thing that came to my mind but i think its fine :)

pkg/core/src/adapters/fuse/FFactory.sol
fedealconada: lol, it's just `valid` but i think it's not the best name, maybe just `counter` would be better...

fedealconada: or `idx`

pkg/core/src/adapters/fuse/FFactory.sol
fedealconada: - The one in `CropsFactory` allows one to set the `rewardsTokens`.
- In the case of the Fuse Adapters, as they have the `rewardsDistributors`, whenever we want to modify the `rewardsTokens` we would likely want to also modify the `rewardsDistributors`.
  - This is why, the `FFactory` has that `setRewardsTokens` that receives both the `rewardsTokens` and the `rewardsDistributors`.
  - `FFactory` will not be making use of the `CropsFactory.setRewardsTokens`
  - In the future, there could be some other type of Factory that does not have this concept of `RewardsDistributor` and just deals with the `rewardsTokens`. In that case, it will use the `CropsFactory.setRewardsTokens`.

pkg/core/src/tests/adapters/Adapter.t.sol
fedealconada: Well... `adapter`is a `MockAdapter` which inherits from `CropAdapter` but the test suite is just testing BaseAdapter things.. This is probably something to fix on the technical debt ticket, wdyt?

pkg/core/src/adapters/BaseFactory.sol
fedealconada: this is the change from `uint16` to `uint64`:

fedealconada: this is the change from uint16 to uint64:
image

pkg/core/src/adapters/fuse/FFactory.sol
fedealconada: i've changed it t `idx`, let me know if you prefer other naming

pkg/core/src/tests/adapters/FAdapter.tm.sol
fedealconada: added [here](https://github.com/sense-finance/sense-v1/pull/234/files#diff-76bd75ca37488f203c24f8295e1db78cf4b671a8ef13f7cb5a3ff2e45b2c64dcR353)

pkg/core/src/tests/adapters/FAdapter.tm.sol
fedealconada: removed!

pkg/core/src/tests/adapters/FAdapter.tm.sol
fedealconada: [here](https://github.com/sense-finance/sense-v1/pull/234/files#diff-76bd75ca37488f203c24f8295e1db78cf4b671a8ef13f7cb5a3ff2e45b2c64dcR339)

pkg/core/src/adapters/BaseAdapter.sol
fedealconada: haha, no problem! i think it makes sense. For someone reading the contracts, it would not be immediately easy to identify why we have some things on the struct on some otherss outside, but I'm down. What about `target`, `ifee` and also `underlying`?

fedealconada: Screenshot 2022-04-20 at 16 30 08

WDYT @jparklev ?

pkg/core/src/adapters/BaseFactory.sol
jparklev: 👍 better as is then

pkg/core/src/adapters/fuse/FAdapter.sol
jparklev: or even just `_rewardsDistributorsList` or something

pkg/core/src/tests/adapters/FAdapter.tm.sol
jparklev: Oh, awesome!

pkg/core/src/tests/adapters/FAdapter.tm.sol
jparklev: cool, that doesn't seem like a problem though, right? What's the use case?

pkg/core/src/adapters/fuse/FAdapter.sol
jparklev: Wow, you're right. And I see that non-rewards tokens actually might not have a `claimRewards` method at all? Like the proxy contract [for `f18-DAI`](https://etherscan.io/address/0x8E4E0257A4759559B4B1AC087fe8d80c63f20D19#writeProxyContract). Going to try to reconcile this with the github

pkg/core/src/adapters/fuse/FFactory.sol
jparklev: nah, that works!

pkg/core/src/tests/adapters/Adapter.t.sol
jparklev: yea, we can add a quality work piece for this. In the meantime, do we need this test file?

pkg/core/src/adapters/BaseAdapter.sol
jparklev: Beautiful (also how big is your monitor, holy cow)

pkg/core/src/tests/adapters/FAdapter.tm.sol
jparklev: Looks great

pkg/core/src/tests/adapters/FAdapter.tm.sol
jparklev: Looks great

pkg/core/src/tests/adapters/FAdapter.tm.sol
jparklev: Looks good. I'm starting to realize fuse upgraded their ftoken contract at some point (f18 tokens are different from f156), so good to test both

pkg/core/src/adapters/fuse/FFactory.sol
jparklev: Yea, I see. I appreciate the context here. It feels very slightly weird to me, but I can't think of a better solution atm ~

pkg/core/src/adapters/compound/CFactory.sol
```diff
@@ -4,44 +4,50 @@ pragma solidity 0.8.11;
...
-                factoryParams.maxm,
-                factoryParams.mode,
-                factoryParams.tilt,
-                DEFAULT_LEVEL,
-                reward
-            )
-        );
+        BaseAdapter.AdapterParams memory adapterParams = BaseAdapter.AdapterParams({
+            target: _target,
+            underlying: _target == CETH ? WETH : CTokenLike(_target).underlying(),
```
jparklev: Is there an `isCEther` like in the f tokens?

pkg/core/src/adapters/fuse/FAdapter.sol
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    function setRewardTokens(address[] memory _rewardTokens, address[] memory _rewardsDistributors)
+        public
+        virtual
+        requiresTrust
+    {
+        super.setRewardTokens(_rewardTokens);
+        for (uint256 i = 0; i < _rewardTokens.length; i++) {
+            rewardsDistributors[_rewardsDistributors[i]] = _rewardTokens[i];
```
jparklev: I'm a little confused about the `rewardsDistributors` mapping, here it's like `rewardsDistributors[_rewardsDistributors[i]] = _rewardTokens[i]` and on line 87 it's like
`rewardsDistributors[_rewardTokens[i]] = _rewardsDistributors[i]`

pkg/core/src/tests/Periphery.tm.sol
```diff
@@ -94,11 +97,18 @@ contract PeripheryTestHelper is DSTest, LiquidityHelper {
...
+        cfactory = new CFactory(address(divider), factoryParams, Assets.COMP);
 
-        divider.setIsTrusted(address(factory), true);
-        divider.setIsTrusted(address(factory), true);
-        periphery.setFactory(address(factory), true);
+        // Deploy fuse factory
+        ffactory = new FFactory(address(divider), factoryParams);
+
+        divider.setIsTrusted(address(cfactory), true);
+        divider.setIsTrusted(address(cfactory), true);
```
jparklev: duplicate calls to `setIsTrusted`? (here and for the ffactory?)

pkg/core/src/tests/adapters/FAdapter.tm.sol
```diff
@@ -0,0 +1,409 @@
...
+
+        adapterParams.target = Assets.f156USDC;
+        adapterParams.underlying = CTokenLike(Assets.f156USDC).underlying();
+        f156UsdcAdapter = new FAdapter(
+            address(divider),
+            Assets.TRIBE_CONVEX,
+            adapterParams,
+            new address[](0),
+            new address[](0)
+        ); // Fuse adapter
```
jparklev:
```suggestion
        ); // Fuse 256 USDC adapter
```

pkg/core/src/adapters/compound/CFactory.sol
fedealconada: I checked and didn't find anything...

pkg/core/src/tests/adapters/FAdapter.tm.sol
fedealconada: ohh, good one!

pkg/core/src/adapters/BaseAdapter.sol
fedealconada: haha, i wanted to show as much as i could so i did ctrl- lol

pkg/core/src/tests/adapters/Adapter.t.sol
fedealconada: i didn't want to remove this file since it was already there, i'd say to keep it there and then, during the quality work, we decide what to do, soudns good?

pkg/core/src/tests/adapters/FAdapter.tm.sol
fedealconada: Since you've now pushed the change to pin a block number, we are now guaranteed that this test will always pass. In the case that, in the future, we decide to update that pinned block and, in the new block, the rewards distributors for that pool have changed, this test will fail. It wouldn't fail if we could have use the `.roll`. But I think we are safe, we just have to have that in mind if one day this test starts maybe because of that reason, in which case we could update the test and that's it.

pkg/core/src/adapters/fuse/FAdapter.sol
fedealconada: Right! To reconcile with what?

pkg/core/src/adapters/fuse/FAdapter.sol
fedealconada: good catch! that's actually a bug, i'm pushing the fix + tests

jparklev:
Looks like the build is failing @fedealconada?

pkg/core/src/Periphery.sol
```diff
@@ -118,10 +118,13 @@ contract Periphery is Trust, IERC3156FlashBorrower {
     /// @dev Called by external address, deploy a new Adapter via an Adapter Factory
     /// @param f Factory to use
     /// @param target Target to onboard
-    function deployAdapter(address f, address target) external returns (address adapter) {
```
KentonPrescott, line 121: Suggest adding a comment describing the `data` parameter

pkg/core/src/adapters/BaseAdapter.sol
```diff
@@ -71,35 +71,23 @@ abstract contract BaseAdapter is IERC3156FlashLender {
...
         ifee = _ifee;
-        stake = _stake;
-        stakeSize = _stakeSize;
-        minm = _minm;
-        maxm = _maxm;
-        mode = _mode;
-        tilt = _tilt;
+
+        // Sanity check
+        if (_adapterParams.minm >= _adapterParams.maxm) revert Errors.InvalidMaturityOffsets();
```
KentonPrescott: Why are there no sanity checks given on other parameters, such as `mode`, `tilt`, and `ifee`?

pkg/core/src/adapters/CropsAdapter.sol
```diff
@@ -0,0 +1,103 @@
...
+
+abstract contract CropsAdapter is Trust, BaseAdapter {
+    using SafeTransferLib for ERC20;
+    using FixedMath for uint256;
+
+    /// @notice Program state
+    uint256 public totalTarget;
+    mapping(address => uint256) public tBalance;
+
+    address[] public rewardTokens;
```
KentonPrescott: Suggest adding comments that describe `totalTarget`, `tBalance`, and `rewardTokens`.

pkg/core/src/adapters/CropsAdapter.sol
```diff
@@ -0,0 +1,103 @@
...
+    mapping(address => Crop) public data;
+
+    struct Crop {
+        // Accumulated reward token per collected target
+        uint256 shares;
+        // Last recorded balance of reward token
```
KentonPrescott: Suggest specifying the owner of the balance. `address(this)` is not immediately clear from the comment.

pkg/core/src/adapters/CropsAdapter.sol
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    /// @notice Some protocols don't airdrop reward tokens, instead users must claim them.
+    /// This method may be overriden by child contracts to claim a protocol's rewards
+    function _claimRewards() internal virtual {
+        return;
+    }
+
+    function setRewardTokens(address[] memory _rewardTokens) public requiresTrust {
```
KentonPrescott, line 99: Couple points re: `setRewardTokens()`:
1. Suggest adding a Natspec'd description
2. Why shouldn't we emit a log in this function?

pkg/core/src/adapters/CropsFactory.sol
```diff
@@ -0,0 +1,23 @@
...
+abstract contract CropsFactory is Trust, BaseFactory {
+    constructor(address _divider, FactoryParams memory _factoryParams)
+        Trust(msg.sender)
+        BaseFactory(_divider, _factoryParams)
+    {}
+
+    /// @notice Replace existing reward tokens array with a new one and update adapters passed
+    /// @param _rewardTokens array of rewards tokens addresses
+    /// @param _adapters array of adapters to update the rewards tokens on
+    function setRewardTokens(address[] memory _rewardTokens, address[] memory _adapters) public requiresTrust {
```
KentonPrescott, line 18: Should we consider emitting an event here?

KentonPrescott (approved):
LGTM. Left a few misc suggestions & assumed that the interface points to Fuse & Fuse pools are correct.

pkg/core/src/adapters/BaseAdapter.sol
fedealconada: This validation was already here, I only adapted it to read the params from `_adapterParams`. We kind of decided not to be validating every input param (not only on this function but in general). I guess that the question could be if we could remove this check since we dont have checks for other params. @jparklev

pkg/core/src/adapters/CropsAdapter.sol
fedealconada: adding event!

pkg/core/src/adapters/CropsFactory.sol
fedealconada: tbh, i feel this one is not needed since we have the one in `adapter.setRewardTokens`

jparklev: very slight agree that we could do w.o., but ok either way

pkg/core/src/adapters/BaseAdapter.sol
jparklev: Right, we assume the parameters used to instantiate are correct and that any sanity checks are done out of bounds (our system has to be resilient against bad params b/c adapters can be created by anyone). I think i'm ok removing this check as well

pkg/core/src/adapters/fuse/FAdapter.sol
```diff
@@ -0,0 +1,191 @@
...
+        requiresTrust
+    {
+        super.setRewardTokens(_rewardTokens);
+        for (uint256 i = 0; i < _rewardTokens.length; i++) {
+            rewardsDistributorsList[_rewardTokens[i]] = _rewardsDistributorsList[i];
+        }
+        emit RewardsDistributorsChanged(_rewardsDistributorsList);
+    }
+
+    /* ========== LOGS ========== */
```
jparklev, line 192:
```suggestion
    /* ========== LOGS ========== */
    
```

pkg/core/src/adapters/fuse/FAdapter.sol
```diff
@@ -0,0 +1,191 @@
...
+    event RewardsDistributorsChanged(address[] indexed rewardsDistributorsList);
+
```
jparklev, line 195:
```suggestion
/* ========== FALLBACK ========== */

```

pkg/core/src/adapters/CropsAdapter.sol
```diff
@@ -0,0 +1,108 @@
...
+        address _target,
+        address _underlying,
+        uint128 _ifee,
+        BaseAdapter.AdapterParams memory _adapterParams,
+        address[] memory _rewardTokens
+    ) Trust(msg.sender) BaseAdapter(_divider, _target, _underlying, _ifee, _adapterParams) {
+        rewardTokens = _rewardTokens;
+    }
+
+    function notify(
```
jparklev, line 46: One thought on `notify`, what if we restricted to run a max of once per block, would that save gas? As it is now, I think notify is called multiple times for some transactions (buying YTs calls issue, which notifies, and transfers YTs, which collects and notifies again). Or maybe even once per hour or day. What do you think? (and same for CropAdapter)

pkg/core/src/tests/adapters/Adapter.t.sol
jparklev: yea, sgtm

pkg/core/src/tests/factories/FFactory.tm.sol
```diff
@@ -0,0 +1,165 @@
...
+        assertEq(FAdapter(adapter).divider(), address(divider));
+        assertEq(FAdapter(adapter).name(), "Tribe Convex Pool Curve.fi Factory USD Metapool: Frax Adapter");
+        assertEq(FAdapter(adapter).symbol(), "fFRAX3CRV-f-156-adapter");
+        assertEq(FAdapter(adapter).rewardTokens(0), Assets.CVX);
+        assertEq(FAdapter(adapter).rewardTokens(1), Assets.CRV);
+        assertEq(FAdapter(adapter).rewardTokens(2), address(0));
+        assertEq(FAdapter(adapter).rewardTokens(3), address(0));
+
+        uint256 scale = FAdapter(adapter).scale();
+        assertTrue(scale > 0);
```
jparklev:
```suggestion
        assertGt(scale, 0);
```

pkg/core/src/tests/factories/FFactory.tm.sol
```diff
@@ -0,0 +1,165 @@
...
+        address f = factory.deployAdapter(Assets.f156FRAX3CRV, abi.encode(Assets.TRIBE_CONVEX));
+        FAdapter adapter = FAdapter(payable(f));
+        assertTrue(address(adapter) != address(0));
+        assertEq(FAdapter(adapter).target(), address(Assets.f156FRAX3CRV));
+        assertEq(FAdapter(adapter).divider(), address(divider));
+        assertEq(FAdapter(adapter).name(), "Tribe Convex Pool Curve.fi Factory USD Metapool: Frax Adapter");
+        assertEq(FAdapter(adapter).symbol(), "fFRAX3CRV-f-156-adapter");
+        assertEq(FAdapter(adapter).rewardTokens(0), Assets.CVX);
+        assertEq(FAdapter(adapter).rewardTokens(1), Assets.CRV);
+        assertEq(FAdapter(adapter).rewardTokens(2), address(0));
```
jparklev, line 107: The comment says the pool has "4 rewards distributors", but there are only two reward tokens?

pkg/core/src/tests/factories/FFactory.tm.sol
```diff
@@ -0,0 +1,165 @@
...
+
+        factory.setRewardTokens(f, rewardTokens, rewardsDistributors);
+
+        assertEq(adapter.rewardTokens(0), Assets.LDO);
+        assertEq(adapter.rewardTokens(1), Assets.FXS);
+        assertEq(adapter.rewardsDistributorsList(Assets.LDO), Assets.REWARDS_DISTRIBUTOR_LDO);
+        assertEq(adapter.rewardsDistributorsList(Assets.FXS), Assets.REWARDS_DISTRIBUTOR_FXS);
+    }
+
+    function testMainnetCantSetRewardsTokens() public {
```
jparklev: We could also make tests like these fuzz tests. ie `testFuzzMainnetCantSetRewardsTokens(address lad)` and do a `if (lad == address(this)) return`

jparklev (approved):
A few final thoughts and questions. Then LGTM!

pkg/core/src/tests/factories/FFactory.tm.sol
fedealconada: It has 4 reward distributors but `fFRAX3CRV-f-156` only receives CVX and CRV rewards

pkg/core/src/adapters/CropsAdapter.sol
fedealconada: You're right, but I feel like it would be a bit complicated to reconcile those calls into just one since, for example in the case you mentioned, we would be calling the notify with different values, right?

jparklev: Yes, that's right – I realize what I was actually trying to guard against is too many calls to `_claimRewards`, so perhaps we could put that check there? Could even do it on the FAdapter/CAdapter themselves)

pkg/core/src/adapters/compound/CAdapter.sol
```diff
@@ -66,6 +66,9 @@ interface ComptrollerLike {
         );
 
     function oracle() external returns (address);
+
+    /// @notice The COMP accrued but not yet transferred to each user
+    function compAccrued(address usr) external returns (uint256);
```
jparklev: unused function?

pkg/core/src/adapters/compound/CAdapter.sol
```diff
@@ -89,6 +92,8 @@ contract CAdapter is CropAdapter {
     bool public immutable isCETH;
     uint8 public immutable uDecimals;
 
+    uint256 internal lastCalledBlock;
```
jparklev: What do u think about a name like `lastRewardedBlock` or something similar?

pkg/core/src/adapters/fuse/FAdapter.sol
```diff
@@ -68,6 +68,8 @@ contract FAdapter is CropsAdapter {
     bool public isFETH;
```
jparklev: Is there a reason these aren't immutable here, but they are on the cAdapter?

jparklev (approved):
Small things, lgt merge after resolving. Nice tests :)

PR #235, opened 2022-04-11
feat(ci): add analyze job with slither

jparklev:
blocked until slither adds foundry support


jparklev:
Archiving this since it seems slither won't work out of the box for us and coverage is a bigger gain in the short-term.

PR #236, opened 2022-04-12
Periphery v1.3: Add basic structure for YT buying with off-chain calcs

jparklev:
## Motivation

Allow users to buy YTs with Target in a gas-efficient way

## Solution

Add a `targetToBorrow` parameter to the `swap___ForYTs` functions, which will now be used in the following operation:
1. Pull Target from the user (or pull underlying and wrap)
2. Flash borrow `targetToBorrow` amount of Target
3. Issue PTs/YTs with both the Target pulled from the user and the Target borrowed
4. Sell all the newly issued PTs for Target
5. Repay the flashloan
6. Send the new YTs and any remaining Target to the user

In order to cover the new flashborrowing path, I split `_flashBorrowAndSwap` into `_flashBorrowAndSwapToYTs` & `_flashBorrowAndSwapFromYTs`, encode a flag indicating the swap type in the data, then switch on the type in the `onFlashLoan` callback. Note that `minAccepted` is now `minOut` (aligns with Balancer's usage), and for the YT buying path in `onFlashLoan`, acts as a limit to how much Target (out of what the user originally passed in) can be returned to the user

## Slippage and Front Running

@KentonPrescott @alanckeegan we talked about front-running, and how the state of the pool can change after we calculate the amount of Target to borrow on the frontend, leading to an uncertain amount of Target being returned. We threw around 5% as a cushion so that, if the pool balances (reserves) change in either direction, the amount of Target returned would end up in most cases between 0 and 10%. I a tests with the 5% optimal return balance, and then tried two cases, somebody buying PTs and somebody selling PTs before the TX gets mined. Here's how it looked:

|     | Target In | Target Returned | PT Reserves |  Target Reserves | PT Price |
| ----------- | ----------- | ---------- | ---------- | ---------- | ------- |
| Pre-calculated      |  5     | 0.25 (5%) |  500 | 521.7 | 0.911 |
| Case 1) After a PT buy   | 5    |  0.58 (11.7%) | 450 | 567.6|  0.921 |
| Case 2) After a PT sell  |  5 | 0.042 (0.8%) | 530 | 494.5 | 0.905 |

As a reminder, it's fine if too much Target is returned (in so far as it's good UX), but the transaction will completely revert if the state of the pool is such that less than `0` Target would be returned

## Implementer Checklist
- [x]  [FIRST TIME ONLY] Review the [solcurity](https://github.com/Rari-Capital/solcurity) standard
- [x]   Create a short "Motivation" section for the PR
- [x]  [OPTIONAL] Integrate the feature into the deployment scripts
- [x]  [OPTIONAL] Add new sanity checks to the deployment scripts

_note: b/c space is fairly complicated and we've run into trouble trying to mock out all the features before, I opted to make all of the tests "integration" tests, in that they all use the real space deployment on mainnet_

resolves GEN-802


pkg/core/src/Periphery.sol
```diff
@@ -665,58 +675,83 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
-            divider.pt(adapter, maturity),
-            amount,
-            pool.getPoolId(),
-            ytBalIn - acceptableError
-        );
+        if (ytToTarget) {
+            ERC20 target = ERC20(Adapter(adapter).target());
 
-        // We take the lowest of the two balances, as long as they're within a margin of acceptable error.
-        if (ptBal >= ytBalIn + acceptableError && ptBal <= ytBalIn - acceptableError)
```
jparklev, line 733: I think this check would never trigger? ie `ptBal` can never be bigger than `ytBalIn + acceptableError` and smaller than `ytBalIn - acceptableError`

jparklev:
Something is up with the rpc url here on CI. Will look into it tomorrow, but went ahead and added you as a reviewer pre tests and spec to get your eyes on the approach @fedealconada

pkg/core/src/Periphery.sol
```diff
@@ -166,35 +166,40 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
-        uint256 minAccepted
+        uint256 underlyingIn,
+        uint256 targetToBorrow,
+        uint256 minOut
     ) external returns (uint256 ytBal) {
         ERC20 underlying = ERC20(Adapter(adapter).underlying());
-        underlying.safeTransferFrom(msg.sender, address(this), uBal); // pull target
-        uint256 tBal = Adapter(adapter).wrapUnderlying(uBal); // wrap underlying into target
-        ytBal = _swapTargetForYTs(adapter, maturity, tBal, minAccepted);
+        underlying.safeTransferFrom(msg.sender, address(this), underlyingIn); // Pull Unerlying
```
fedealconada: typo in "Unerlying"

pkg/core/src/Periphery.sol
```diff
@@ -665,58 +669,82 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
+        bool result = Adapter(adapter).flashLoan(this, Adapter(adapter).target(), amountToBorrow, data);
+        if (!result) revert Errors.FlashBorrowFailed();
+    }
+
     /// @dev ERC-3156 Flash loan callback
     function onFlashLoan(
         address initiator,
         address, /* token */
-        uint256 amount,
+        uint256 amountBorrrowed,
```
fedealconada, line 725: this was `amount` as per the `ERC-3156` specification, I'm good with both but just wanted to comment on this

jparklev:
All of the new tests are in `pkg/core/src/tests/Periphery.tm.sol`

pkg/core/src/tests/Divider.t.sol
```diff
@@ -783,9 +783,7 @@ contract Dividers is TestHelper {
     }
 
     function testFuzzCombine(uint128 tBal) public {
-        if (tBal < 100) {
-            return;
-        }
+        hevm.assume(tBal > 10);
```
fedealconada, line 754: ♥

pkg/core/src/tests/Periphery.tm.sol
```diff
@@ -11,133 +11,463 @@ import { ERC20 } from "@rari-capital/solmate/src/tokens/ERC20.sol";
...
+        assertEq(targetBalPre - targetBalPost + targetReturned, TARGET_IN);
+        assertEq(ytBalPost - ytBalPre, ytsOut);
+        // Check that the YTs returned are the result of issuing from the borrowed Target + transferred Target
+        assertEq(ytsOut, (TARGET_IN + TARGET_TO_BORROW).fmul(1e18 - mockAdapter.ifee()));
+
+        // Check that we got less than 0.000001 Target back
+        assertTrue(targetReturned < 0.000001e18);
+    }
+
+    // Pattern similar to https://github.com/FrankieIsLost/gradual-dutch-auction/src/test/ContinuousGDA.t.sol#L113
```
fedealconada:
```suggestion
    // Pattern similar to https://github.com/FrankieIsLost/gradual-dutch-auction/blob/master/src/test/ContinuousGDA.t.sol#L113
```

yt-buying-lib/index.js
```diff
@@ -0,0 +1,105 @@
+import Decimal from "decimal.js";
+import optimjs from "optimization-js";
+import SETTINGS from "./settings.js";
+
+class SpaceFluxer {
```
fedealconada, line 5: nice name

fedealconada:
- [x]  [FIRST TIME ONLY] Review the [[solcurity](https://github.com/Rari-Capital/solcurity)](https://github.com/Rari-Capital/solcurity) standard
- [x]  Read the PR descriptions and the linear ticket
- [x]  Do a high-level review of the code to confirm that the implementation does what it’s supposed to do
- [x]  Go line-by-line through the implementation and check state transitions – since we don’t have a coverage engine for foundry yet, you can imagine how one might check that each different code path is covered
- [x]  Look at the feature and take some time to think about what tests *you* would write, then confirm that all of those cases are covered by the tests that are present
- [x]  Confirm that the implementer has completed the implementer checklist correctly

fedealconada (approved):
lgtm! nice good test suite :)

- [x]  [FIRST TIME ONLY] Review the [[solcurity](https://github.com/Rari-Capital/solcurity)](https://github.com/Rari-Capital/solcurity) standard
- [x]  Read the PR descriptions and the linear ticket
- [x]  Do a high-level review of the code to confirm that the implementation does what it’s supposed to do
- [x]  Go line-by-line through the implementation and check state transitions – since we don’t have a coverage engine for foundry yet, you can imagine how one might check that each different code path is covered
- [x]  Look at the feature and take some time to think about what tests *you* would write, then confirm that all of those cases are covered by the tests that are present
- [x]  Confirm that the implementer has completed the implementer checklist correctly

alexfertel:
> and in for the YT buying side functions as a limit to how much Target can be returned to the user

Haven't had a good look at the code yet, but my mind couldn't parse the previous sentence very well, would you mind rephrasing it?

jparklev:
> > and in for the YT buying side functions as a limit to how much Target can be returned to the user
> 
> Haven't had a good look at the code yet, but my mind couldn't parse the previous sentence very well, would you mind rephrasing it?

Just rephrased to 
> and for the YT buying path in `onFlashLoan`, acts as a limit to how much Target (out of what the user originally passed in) can be returned to the user
is that clearer?

alexfertel:
> > > and in for the YT buying side functions as a limit to how much Target can be returned to the user
> > 
> > 
> > Haven't had a good look at the code yet, but my mind couldn't parse the previous sentence very well, would you mind rephrasing it?
> 
> Just rephrased to
> 
> > and for the YT buying path in `onFlashLoan`, acts as a limit to how much Target (out of what the user originally passed in) can be returned to the user
> > is that clearer?

awesome! arigato!

pkg/core/src/Periphery.sol
```diff
@@ -679,58 +685,82 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
-        bool result = Adapter(adapter).flashLoan(this, address(target), amount, data);
+        uint256 decimals = target.decimals();
+        uint256 acceptableError = decimals < 9 ? 1 : PRICE_ESTIMATE_ACCEPTABLE_ERROR / 10**(18 - decimals);
+        bytes memory data = abi.encode(adapter, uint256(maturity), ytBalIn, ytBalIn - acceptableError, true);
+        bool result = Adapter(adapter).flashLoan(this, address(target), amountToBorrow, data);
         tBal = target.balanceOf(address(this));
         if (!result) revert Errors.FlashBorrowFailed();
     }
 
+    function _flashBorrowAndSwapToYTs(
```
alexfertel, line 706: Would we want explanation comments for this function like the ones for the `From` counterpart?

```
    /// @notice Initiates a flash loan of Target, swaps target amount to PTs and combines
    /// @param adapter adapter
    /// @param maturity maturity
    /// @param ytBalIn YT amount the user has sent in
    /// @param amountToBorrow target amount to borrow
    /// @return tBal amount of Target obtained from a sale of YTs
```

pkg/core/src/Periphery.sol
```diff
@@ -679,58 +685,82 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
-        uint256 amount
+        uint256 amountToBorrow
     ) internal returns (uint256 tBal) {
         ERC20 target = ERC20(Adapter(adapter).target());
-        bytes memory data = abi.encode(adapter, maturity, ytBalIn);
-        bool result = Adapter(adapter).flashLoan(this, address(target), amount, data);
+        uint256 decimals = target.decimals();
+        uint256 acceptableError = decimals < 9 ? 1 : PRICE_ESTIMATE_ACCEPTABLE_ERROR / 10**(18 - decimals);
+        bytes memory data = abi.encode(adapter, uint256(maturity), ytBalIn, ytBalIn - acceptableError, true);
+        bool result = Adapter(adapter).flashLoan(this, address(target), amountToBorrow, data);
```
alexfertel, line 692: Probably something I'm missing, but how do we know the Adapter has enough target to loan `amountToBorrow`?

pkg/core/src/Periphery.sol
```diff
@@ -505,17 +512,16 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
-        tBal = _swap(divider.pt(adapter, maturity), Adapter(adapter).target(), issued, pool.getPoolId(), minAccepted);
-
-        // transfer yts & target to user
-        ERC20(Adapter(adapter).target()).safeTransfer(msg.sender, tBal);
-        ERC20(divider.yt(adapter, maturity)).safeTransfer(msg.sender, issued);
+        // Transfer YTs & Target to user
+        ERC20 target = ERC20(Adapter(adapter).target());
+        ERC20 yt = ERC20(divider.yt(adapter, maturity));
+        target.safeTransfer(msg.sender, targetBal = target.balanceOf(address(this)));
+        yt.safeTransfer(msg.sender, ytBal = yt.balanceOf(address(this)));
```
alexfertel: Just to confirm, this means that the Periphery always stays with 0 balance of target and PT/YT after a buy swap?

yt-buying-lib/index.js
```diff
@@ -0,0 +1,105 @@
...
+console.log("\nArgs:", { targetToJoin, ptsToSwapIn, targetInForYTs, optimalTargetReturned });
+
+// Check price
+const ptPrice = spaceFluxer._swapPTsForTarget(new Decimal("0.00001")).times(100000);
+console.log("\nPT Price:", ptPrice);
+
+// Optimize YT buy
+spaceFluxer.setScale(CURRENT_SCALE);
+const buyYTTargetToBorrow = spaceFluxer.buyYTs(new Decimal(targetInForYTs), new Decimal(optimalTargetReturned));
+console.log("\nOptimal amount to borrow:", optimjs.minimize_Powell(buyYTTargetToBorrow, [0.5]).argument[0]);
```
alexfertel, line 105: Let me know if I scoped properly what needs to be done on the FE:

* Rework SpaceFluxer so that ptReserves, targetReserves, and supply can be modified such that they reflect the state of the pool we're going to buy YTs from and rename `ttl` to `maturity`.
* Add it as a lib file and update our package.json
* Init a SpaceFluxer with the pool params.
* Update the calls to `swap__ForYTs` so that the `amountToBorrow` is the result of `optimjs.minimize_Powell(buyYTTargetToBorrow, [0.5]).argument[0]`.

yt-buying-lib/index.js
```diff
@@ -0,0 +1,105 @@
...
+const [targetToJoin, ptsToSwapIn, targetInForYTs, _optimalTargetReturned] = process.argv.slice(2);
+if ([targetToJoin, ptsToSwapIn, targetInForYTs].includes(undefined)) {
+  throw new Error(
+    `Missing at least one of the three required positional arguments:
+
+    <targetToJoin> <ptsToSwapIn> <targetInForYTs> <[optional] optimalTargetReturned>
+    `,
+  );
+}
+const optimalTargetReturned = _optimalTargetReturned || 0;
```
alexfertel, line 86: When would it be different than `0`? Isn't `0` the optimal?

pkg/core/src/tests/test-helpers/mocks/MockAdapter.sol
```diff
@@ -65,28 +65,26 @@ contract MockAdapter is CropAdapter {
...
-        if (_value != lscale.value) {
-            // update value only if different than the previous
-            lscale.value = _value;
+    function scale() external virtual override returns (uint256 _scale) {
+        if (scaleOverride > 0) {
+            _scale = scaleOverride;
+            lscale.value = scaleOverride;
             lscale.timestamp = block.timestamp;
+        } else {
+            uint256 gps = GROWTH_PER_SECOND.fmul(99 * (10**(18 - 2)));
```
alexfertel, line 53: Is it too costly to extract this to a constant? I'm not sure what this number means. Also, why `18 - 2` and not `16`? (I assume it's cause this 18 shows it has to do smth with decimals?)

alexfertel (approved):
Ok, to the best of my ability with solidity I went over every line. Aside from some questions about the integration with the frontend, I think I understood what we are doing and it looks good to me. Great job!

- [x]  [FIRST TIME ONLY] Review the [[solcurity](https://github.com/Rari-Capital/solcurity)](https://github.com/Rari-Capital/solcurity) standard
- [x]  Read the PR descriptions and the linear ticket
- [x]  Do a high-level review of the code to confirm that the implementation does what it’s supposed to do
- [x] Go line-by-line through the implementation and check state transitions – since we don’t have a coverage engine for foundry yet, you can imagine how one might check that each different code path is covered
- [x]  Look at the feature and take some time to think about what tests *you* would write, then confirm that all of those cases are covered by the tests that are present
- [x]  Confirm that the implementer has completed the implementer checklist correctly

pkg/core/src/Periphery.sol
jparklev: Sure, not a bad idea to add that in

pkg/core/src/Periphery.sol
jparklev: We don't at this step! The tx will just revert overall if it doesn't (so something we can be aware of on the FE potentially)

pkg/core/src/Periphery.sol
jparklev: _exactly_

yt-buying-lib/index.js
jparklev: Sounds correct to me. Quick things:
* The fluxer uses Decimal.js, so will need to finally bring that in :sadge:
* Should be `targetToBorrow` for the public `swap__ForYTs` functions #pendant
* `minimize_Powell`'s second arg is a _guess/starting point_ for the minimization function. I've never had it fail on me when starting from `0.5`, but there's some potential we'll need to do something a little smarter there in the future

yt-buying-lib/index.js
jparklev: Aha, great question!! :mad_scientist: Checkout the "Slippage and Front Running" section of the PR description – we'll probably actually use a heuristic like "5% of the user's passed in target" as the optimal target return, to make sure the tx succeeds, even if other txs are mined that change the exchange rate before the user's

pkg/core/src/tests/test-helpers/mocks/MockAdapter.sol
jparklev: Yea, that's my understanding but possible a better q for mr @fedealconada

jparklev:
Waiting to merge until #234 is merged – following the deployment plan mentioned here https://github.com/sense-finance/sense-v1/pull/240#issuecomment-1111074279

pkg/core/src/Periphery.sol
jparklev: done https://github.com/sense-finance/sense-v1/pull/236/commits/7b25a3468a46c94d7aec27acf38944a62153b3e2

pkg/core/src/Periphery.sol
```diff
@@ -679,58 +685,88 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
+    /// @param targetIn Target amount the user has sent in
+    /// @param amountToBorrow Target amount to borrow
+    /// @return minOut minimum amount of Target accepted out for the issued PTs
+    function _flashBorrowAndSwapToYTs(
+        address adapter,
+        uint256 maturity,
+        uint256 targetIn,
+        uint256 amountToBorrow,
+        uint256 minOut
+    ) internal returns (uint256 tBal) {
```
jparklev: @fedealconada @alexfertel looks like we missed this unused return value. Might be a hole in our process, maybe we should add a checklist item like "check that all parameters and return values are used"? What do y'all think?

alexfertel: Hmmm, shouldn't that be covered on a linter rule? 🤔 I actually saw this, but I thought it was me not knowing something about solidity 😞 Should have mentioned it!

Maybe we should add something to the list like: "Make sure new linter/compiler warnings are not added"?

jparklev: > I actually saw this, but I thought it was me not knowing something about solidity 😞 Should have mentioned it!

Totally! Next time 

> Maybe we should add something to the list like: "Make sure new linter/compiler warnings are not added"?

Yea, we fail the CI if the linter errors, but maybe it only warns for cases like these? If so, we should look at making our linter more restrictive, would be great if we could automate the process and get a failing CI for cases like this (also slither might help us achieve something similar in #235)

pkg/core/src/tests/test-helpers/mocks/MockAdapter.sol
fedealconada: > Is it too costly to extract this to a constant?

It would actually be better to have it in a constant bu we might have not done that because this is just a Mock contract... 

> I'm not sure what this number means

To be reaaally honest, I don't remember how did we get to this number, but the baseline is that we grab the `GROWTH_PER_SECOND` which represents a 25% APY (arbitrary number) and, together with a time variable, we calculate a scale value. So, we have a dynamic scale value that changes over time to make this function more "realistic".

>  Also, why 18 - 2 and not 16

Yeah, right.

pkg/core/src/Periphery.sol
fedealconada: I missed it as well! Yeah, probably making linter more restrictive will do the trick.

yt-buying-lib/README.md
```diff
@@ -0,0 +1,12 @@
+# YT Buying JS Library 
+
+## Development
+
+This repo uses an npm package that cannot be resolved with `yarn`, so the dependencies must be installed with `npm.
```
alexfertel, line 5:
```suggestion
This repo uses an npm package that cannot be resolved with `yarn`, so the dependencies must be installed with `npm`.
```

I'm wondering if I'm doing something differently. I just ran `yarn add decimal.js optimization-js` and it worked fine. `optimization-js` has 336 deps tho...

jparklev: Interesting, I wonder if the yarn issue I was seeing here w `optimization-js` was b/c of the yarn workspace setup with the rest of the repo? Regardless, if it's working with yarn on your side, that's great!

> optimization-js has 336 deps tho...

Hmph. looks ok size-wise thought, right? :)
Screen Shot 2022-04-28 at 9 23 25 AM

pkg/core/src/tests/test-helpers/mocks/MockAdapter.sol
jparklev: Will leave it untouched for now since it's a testing contract, but we can circle back to the mock adapter in quality work down the line

pkg/core/src/Periphery.sol
jparklev: Hmm these linting rule options aren't that great tbh https://github.com/protofire/solhint/blob/master/docs/rules.md#best-practise-rules

PR #238, opened 2022-04-19
GEN-820: swappable periphery params

jparklev:
## Motivation

To make future migrations easier by not requiring us to redeploy the Periphery every time we change the Space Factory or Pool Manager

## Solution

Remove the immutable modifier from the `poolManager` and `spaceFactory` storage variables, then add setters that let trusted actors update the values

## Implementer Checklist
- [x]  [FIRST TIME ONLY] Review the [solcurity](https://github.com/Rari-Capital/solcurity) standard
- [x]   Create a short "Motivation" section for the PR

### DIFF
--- a/pkg/core/src/Periphery.sol
+++ b/pkg/core/src/Periphery.sol
@@ -44,15 +44,15 @@ contract Periphery is Trust, IERC3156FlashBorrower {
     Divider public immutable divider;
 
     /// @notice Sense core Divider address
-    PoolManager public immutable poolManager;
-
-    /// @notice Sense core Divider address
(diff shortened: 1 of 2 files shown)


jparklev:
Here's how our gas is looking after moving from immutable to mutable Screen Shot 2022-04-19 at 12 24 24 AM

jparklev:
Here's how our gas is looking after moving from immutable to mutable Screen Shot 2022-04-19 at 12 24 24 AM

fedealconada (approved):
clean PR! lgtm!

PR #242, opened 2022-04-27
GEN-904 deploy Fuse Factory

fedealconada:
## Solution

Add tasks that deploys the new Fuse Factory to mainnet and verifies factory contract on Etherscan.

## Implementer Checklist

N/A


fedealconada:
Blocked by the deployment of the new Periphery cc @jparklev

fedealconada:
friendly reminder @jparklev @abylaw

pkg/deployments/tasks/20220531-fuse-factory/index.js
```diff
@@ -0,0 +1,127 @@
...
+    // if mainnet, verify on etherscan
+    if (chainId === "1") {
+      console.log("\n-------------------------------------------------------");
+      console.log("Trying to verify contract on Etherscan...");
+      try {
+        const hola = await hre.run("verify:verify", {
+          address: factoryAddress,
+          constructorArguments: [divider.address, factoryParams],
+        });
+        console.log('HOLA', hola);
```
jparklev: What's the return value for `verify`? Also should we teak the log message? :p

pkg/deployments/tasks/20220531-fuse-factory/index.js
```diff
@@ -0,0 +1,127 @@
...
+      }
+    }
+
+    if (chainId === "1") {
+      console.log("\n-------------------------------------------------------");
+      console.log("\nACTIONS TO BE DONE ON DEFENDER: ");
+      console.log("\n1. Set factory on Periphery");
+      console.log("\n2. Deploy adapters on Periphery (if needed)");
+      console.log("\n3. For each adapter, set corresponding guard on Divider");
+      console.log("\n4. For each adapter, sponsor Series on Periphery");
```
jparklev, line 116: Nice addition!

pkg/deployments/tasks/20220531-fuse-factory/input.js
```diff
@@ -0,0 +1,32 @@
...
+    targets: F_FACTORY_TARGETS
+  },
+];
+
+module.exports = {
+  mainnet: {
+    divider: "0x86bA3E96Be68563E41c2f5769F1AF9fAf758e6E0",
+    periphery: "0xFff11417a58781D3C72083CB45EF54d79Cd02437", 
+    factories: MAINNET_FACTORIES,
+    adapters: [],
```
jparklev: unused, i think. Also the adapters to deploy are set by the `F_FACTORY_TARGETS` var, right?

pkg/deployments/tasks/20220531-fuse-factory/input.js
```diff
@@ -0,0 +1,32 @@
+const { WETH_TOKEN, MASTER_ORACLE, F18DAI_TOKEN, OLYMPUS_POOL_PARTY } = require("../../hardhat.addresses");
+const ethers = require("ethers");
+
+// TODO: decide which adapter targets to deploy
```
jparklev, line 4: Yea, can confirm w @KentonPrescott and @alanckeegan before we actually run these! (same w the other params)

pkg/deployments/tasks/20220531-fuse-factory/input.js
```diff
@@ -0,0 +1,32 @@
...
+const F_FACTORY_TARGETS = [
+  { name: "f18DAI", series: [], address: F18DAI_TOKEN.get("1"), comptroller: OLYMPUS_POOL_PARTY.get("1") },
+];
+
+const MAINNET_FACTORIES = [
```
jparklev, line 9: We're only planning on one factory for fuse afaik, so we could make this a single object – wdyt?

jparklev (approved):
tiny comments, but all looks good

pkg/deployments/tasks/20220531-fuse-factory/index.js
fedealconada: haha, i'm always forgetting a weird console.log somwhere, have to improve that...
it returns nothing, i was checking exactly that because nothing was said on the docs

pkg/deployments/tasks/20220531-fuse-factory/input.js
fedealconada: yeah, removing it! the `adapaters` there would be the adapters we want to deploy directly (without factory) which we are not using for this task and `F_FACTORY_TARGETS` are the ones to be deployed via this fuse factory

pkg/deployments/tasks/20220531-fuse-factory/input.js
fedealconada: yeah, i think that we may want to decide how we want to do this tasks forward. My idea of having that as an array (same reason why I kept the `adapters: []`) was kind of thinking on that we always have the same "structure" and then we fill in what we need. I have not been 100% consistent though but that was the idea...

pkg/deployments/tasks/20220531-fuse-factory/input.js
fedealconada: yep

pkg/deployments/tasks/20220531-fuse-factory/index.js
abylaw: +1

pkg/deployments/tasks/20220531-fuse-factory/index.js
```diff
@@ -0,0 +1,129 @@
...
+      try {
+        await hre.run("verify:verify", {
+          address: factoryAddress,
+          constructorArguments: [divider.address, factoryParams],
+        });
+      } catch (e) {
+        console.log(e);
+        console.log("We couldn't verify the contract on Etherscan, you may try manually.");
+      }
+    }
```
abylaw, line 67: maybe a bit early for this, but might be nice to pull this into a `util` function (e.g. `verifyContractOnEtherscan`) for other tasks to consume.

pkg/deployments/tasks/20220531-fuse-factory/input.js
abylaw: > My idea of having that as an array (same reason why I kept the adapters: []) was kind of thinking on that we always have the same "structure" and then we fill in what we need.

I'm assuming this is to try and reuse some of our tasks, right? (correct me if I'm wrong)

I think we want two things out of our tasks right now, which are:
1. a reliable way to reproduce a state on a blockchain
2. reusable parts so we don't have to write a task from scratch each time

So maybe it would make sense to separate our code into:

1. migrations (meant to be run once or idempotently, and running all migrations should bring a chain to the same "state"), 
2. reusable tasks which can be run with different params  (e.g., deploy adapter factory, deprecate adapter, etc..., shared code that interacts with contracts. we could also give them versions e.g. `spaceFactory_v_1_4.)

\+ other utils not tied to interacting/deploying contracts, like `verifyContractOnEtherscan`

that way, migrations can be fairly simple:
```
const { mainnet } = require("./input");

const factory = await deployAdapterFactory(mainnet.FACTORY)
await verifyContractOnEtherscan(factory)
```

wdygt?

abylaw (approved):
LGTM :) One small suggestion & a larger discussion on how we should structure our tasks, but which shouldn't block merging.

pkg/deployments/tasks/20220531-fuse-factory/input.js
fedealconada: yeah, i think this could be a good direction, definitely something we can improve on! gonna create a QW task and ping to this comment so we can re-structure tasks at some point to make them even easier to create and code to be more re-usable :)

PR #244, opened 2022-05-05
chore(core/Periphery): add yt buying event

jparklev:
## Motivation

The existing `Swapped` event doesn't give us enough information to establish a cost basis for users' YTs. See the comment here: https://github.com/sense-finance/sense-subgraph/pull/17/files#r861448409

## Solution

Add a `YTPurchased` event that provides both the initial target in and the target returned

## Implementer Checklist

N/A


fedealconada (approved):
LGTM!

PR #245, opened 2022-05-12
refactor(core/adapters): make maturity bounds function virtual

jparklev:
## Motivation

Enable the [auto roller](https://github.com/sense-finance/auto-roller) to prevent new Series from being initialized

## Solution

Make the maturity bounds function virtual so that custom adapters can be written that have custom logic for setting the maturity bounds. This allows them to _disable_ new series by setting the `maxm` to 0, then conditionally changing it when it's ready to allow new Series again

## Implementer Checklist

N/A


fedealconada (approved):
LGTM!

PR #246, opened 2022-05-16
GEN-1043: add OZ relayer address as a Divider trusted address on Goerli

fedealconada:
## Motivation

When deploying to Goerli, we remove the deployer address from the trusted addresses for the Divider but we are not adding any other address.

Therefore, there's no way to pause the divider. We can call the emergency stop but that will not pause the divider (it will only set the permissionless mode to false and pause all the adapters passed in the args). 

## Solution

Set an address, different to the deployer's address, as trusted for the Divider on Goerli.  

## Implementer Checklist

N/A

### DIFF
--- a/pkg/deployments/deploy/prod/07-multisig_setup.js
+++ b/pkg/deployments/deploy/prod/07-multisig_setup.js
@@ -6,8 +6,14 @@ module.exports = async function () {
   const deployerSigner = await ethers.getSigner(deployer);
   const chainId = await getChainId();
 
-  if (!SENSE_MULTISIG.has(chainId)) throw Error("No sense multisig found");
-  const multisig = chainId === "5" ? OZ_RELAYER.get("5") : SENSE_MULTISIG.get(chainId);
+  let signer;
+  if (chainId === "1") {
+    if (!SENSE_MULTISIG.has("1")) throw Error("No sense multisig found for mainnet");
+    signer = SENSE_MULTISIG.get(chainId);
+  } else {
+    if (!OZ_RELAYER.get(chainId)) throw Error(`No OZ relayer found for chain ID ${chainId}`);
+    signer = OZ_RELAYER.get("5");
+  }
 
   let tokenHandler = await ethers.getContract("TokenHandler", deployerSigner);
   let divider = await ethers.getContract("Divider", deployerSigner);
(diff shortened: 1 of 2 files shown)


pkg/deployments/deploy/prod/07-multisig_setup.js
```diff
@@ -7,7 +7,7 @@ module.exports = async function () {
   const chainId = await getChainId();
 
   if (!SENSE_MULTISIG.has(chainId)) throw Error("No sense multisig found");
```
jparklev: Shouldn't this fail if there's not address for this chainid? Which address are we setting as trusted on these contracts atm?

pkg/deployments/deploy/prod/07-multisig_setup.js
```diff
@@ -7,7 +7,7 @@ module.exports = async function () {
...
-  const multisig = SENSE_MULTISIG.get(chainId);
+  const multisig = chainId === "5" ? OZ_RELAYER.get(chainId) : SENSE_MULTISIG.get(chainId);
```
jparklev: This approach looks good to me, but think we shouldn't do the `SENSE_MULTISIG.has(chainId)` for chainid 5 if we're doing this :)

pkg/deployments/deploy/prod/07-multisig_setup.js
fedealconada: So, the "real" issue was that, SENSE_MULTISIG for goerli matches the deployer address so:

```
  log("Trust the multisig address on the divider");
  if (!(await divider.isTrusted(multisig))) {
    await (await divider.setIsTrusted(multisig, true)).wait();
  }
  ```
  
because of if check, we are not adding the multisig as trusted because it is already trusted. But then, at the end of this script we remove the deployer address as a trusted address.

File: pkg/deployments/deploy/prod/07-multisig_setup.js (hunk 2 above)
Comment (fedealconada) [reply pkg/deployments/deploy/prod/07-multisig_setup.js]: Not sure I follow...

**Comment by fedealconada** [comment]
was a bit procrastinated but check at the new changes @jparklev @alexfertel

**Comment by fedealconada** [comment]
gentle bump @jparklev @alexfertel

File: pkg/deployments/deploy/prod/07-multisig_setup.js (hunk 1 above)
Comment (jparklev) [reply pkg/deployments/deploy/prod/07-multisig_setup.js]: > SENSE_MULTISIG for goerli matches the deployer address

What do you think about changing this? Like maybe not having a multisig address for goerli at all, which wouldn't be a problem since we use the `OZ_RELAYER` for chain id 5

File: pkg/deployments/deploy/prod/07-multisig_setup.js [hunk 3]
```diff
@@ -1,65 +1,112 @@
...
 
   log("Trust the multisig address on the periphery");
-  await (await periphery.setIsTrusted(multisig, true)).wait();
+  if (!(await periphery.isTrusted(multisig))) {
+    await (await periphery.setIsTrusted(multisig, true)).wait();
+  }
 
   log("Trust the multisig address on the emergency stop");
-  await (await emergencyStop.setIsTrusted(multisig, true)).wait();
+  if (!(await emergencyStop.isTrusted(multisig))) {
```
jparklev: These checks are to avoid duplicate transactions?

pkg/deployments/deploy/prod/07-multisig_setup.js
```diff
@@ -1,65 +1,112 @@
...
 
-  log("Untrust deployer on the token handler");
-  await (await tokenHandler.setIsTrusted(deployer, false)).wait();
+    log("\n-------------------------------------------------------");
+    log("\nRemove deployer address from trusted on contracts");
 
-  log("Untrust deployer on the divider");
-  await (await divider.setIsTrusted(deployer, false)).wait();
+    log("Untrust deployer on the token handler");
+    await (await tokenHandler.setIsTrusted(deployer, false)).wait();
```
jparklev: should we do like a `if (await tokenHandler.isTrusted(deployer)) {` for these to be consistent with the other simulated script?

jparklev (approved):
nits and a question about naming (multisig vs relayer)

pkg/deployments/deploy/prod/07-multisig_setup.js
fedealconada: yep, if we are running the script and a transaction fails (which sometimes happen) then, when re-running, if we dont have this if conditionals, it would execute again these functions

fedealconada: we had the same for the sim scripts but I guess we forgot to add them on these ones

pkg/deployments/deploy/prod/07-multisig_setup.js
fedealconada: totally! i've just missed those :) ty!

pkg/deployments/deploy/prod/07-multisig_setup.js
fedealconada: something like changing that into this? cc @jparklev 
```
  let signer;
  if (chainId === "1") {
    if (!SENSE_MULTISIG.has("1")) throw Error("No sense multisig found for mainnet");
    signer = SENSE_MULTISIG.get(chainId);
  } else {
    if (!OZ_RELAYER.get(chainId)) throw Error(`No OZ relayer found for chain ID ${chainId}`);
    signer = OZ_RELAYER.get("5");
  }
  ```

alexfertel (approved):
LGTM!

fedealconada:
approved @jparklev ?

pkg/deployments/deploy/prod/07-multisig_setup.js
jparklev: Yep!

jparklev:
Yea, good with me!

PR #247, opened 2022-05-17
feat: add task to deploy long wstETH adapter

fedealconada:
## Motivation

We want to allow Sense to have long term wstETH Series

## Solution

Deploy a wstETH adapter with corresponding config so as to allow series of max 10 years maturity.

## Implementer Checklist


pkg/deployments/tasks/20220517-long-wsteth-adapter/index.js
```diff
@@ -0,0 +1,96 @@
...
+    const adptr = new ethers.Contract(adapterAddress, adapterAbi, signer);
+    const scale = await adptr.callStatic.scale();
+    console.log(`-> scale: ${scale.toString()}`);
+    console.log(`${contractName} adapterAddress: ${adapterAddress}`);
+
+    const { address: targetAddress } = target;
+    const { abi: tokenAbi } = await deployments.getArtifact("Token");
+    const targetContract = new ethers.Contract(targetAddress, tokenAbi, signer);
+
+    console.log("\nEnable the Periphery to move the deployer's Target for liquidity provisions");
```
jparklev: Are we provisioning liquidity from the deployer?

jparklev (approved):
Nice, overall straightforward and LGTM. You tried it on a hardhat fork and everything looked OK?

fedealconada:
> Nice, overall straightforward and LGTM. You tried it on a hardhat fork and everything looked OK?

Yep, fork runs ok!

pkg/deployments/tasks/20220517-long-wsteth-adapter/index.js
fedealconada: No, we are provisioning liquidity from relayer through the capital seeder

pkg/deployments/tasks/20220517-long-wsteth-adapter/input.js
```diff
@@ -0,0 +1,53 @@
...
+module.exports = {
+  mainnet: {
+    divider: "0x86bA3E96Be68563E41c2f5769F1AF9fAf758e6E0",
+    periphery: "0xFff11417a58781D3C72083CB45EF54d79Cd02437",
+    poolManager: "0xf01eb98de53ed964ac3f786b80ed8ce33f05f417",
+    vault: "0xBA12222222228d8Ba445958a75a0704d566BF2C8",
+    spaceFactory: "0x984682770f1EED90C00cd57B06b151EC12e7c51C",
+    adapters: MAINNET_ADAPTERS,
+  },
+};
```
abylaw, line 53: Suggest pulling these out somewhere so that we can easily change the addresses (assuming other tasks also use these addresses) if we release a new version the `periphery`/`poolManager`/`spaceFactory`? It looks like we already have `hardhat.addresses`.
Ditto with the ABIs

pkg/deployments/tasks/20220517-long-wsteth-adapter/index.js
```diff
@@ -0,0 +1,87 @@
...
+      if (!SENSE_MULTISIG.has(chainId)) throw Error("No balancer vault found");
+      const senseAdminMultisigAddress = SENSE_MULTISIG.get(chainId);
+      await hre.network.provider.request({
+        method: "hardhat_impersonateAccount",
+        params: [senseAdminMultisigAddress],
+      });
+      const signer = await hre.ethers.getSigner(senseAdminMultisigAddress);
+
+      divider = new ethers.Contract(mainnet.divider, dividerAbi, signer);
+      periphery = new ethers.Contract(mainnet.periphery, peripheryAbi, signer);
```
abylaw: can also use: https://docs.ethers.io/v5/api/contract/contract/#Contract-connect

pkg/deployments/tasks/20220517-long-wsteth-adapter/index.js
```diff
@@ -0,0 +1,87 @@
...
+const peripheryAbi = require("./abi/Periphery.json");
+
+task(
+  "20220517-long-wsteth-adapter",
+  "Deploys long term wstETH adapter",
+).setAction(async ({}, { ethers }) => {
+  const { deploy } = deployments;
+  const { deployer } = await getNamedAccounts();
+  const chainId = await getChainId();
+  const signer = await ethers.getSigner(deployer);
```
abylaw: nit: `signer` -> `deployerSigner`

pkg/deployments/tasks/20220517-long-wsteth-adapter/index.js
```diff
@@ -0,0 +1,87 @@
...
+      console.log("\n-------------------------------------------------------");
+      console.log("Checking multisig txs by impersonating the address");
+
+      if (!SENSE_MULTISIG.has(chainId)) throw Error("No balancer vault found");
+      const senseAdminMultisigAddress = SENSE_MULTISIG.get(chainId);
+      await hre.network.provider.request({
+        method: "hardhat_impersonateAccount",
+        params: [senseAdminMultisigAddress],
+      });
+      const signer = await hre.ethers.getSigner(senseAdminMultisigAddress);
```
abylaw: nit: `signer` -> `multisigSigner`

(easier to see which signer is which w/o needing the read the rest of the code)

pkg/deployments/tasks/20220517-long-wsteth-adapter/output/localhost/solcInputs/c685b0e941c886293d0434d81554143d.json
```diff
@@ -0,0 +1,194 @@
+{
```
abylaw: Just to remind myself: these files are how hardhat keeps track of deployments on various chains? If so, should we be commiting `localhost` deployments?

pkg/deployments/tasks/20220517-long-wsteth-adapter/index.js
```diff
@@ -0,0 +1,87 @@
...
+      console.log(`Onboard ${contractName} adapter`);
+      await periphery.onboardAdapter(adapterAddress, true).then(tx => tx.wait());
+
+      await hre.network.provider.request({
+        method: "hardhat_stopImpersonatingAccount",
+        params: [senseAdminMultisigAddress],
+      });
+    }
+    divider = new ethers.Contract(mainnet.divider, dividerAbi, signer);
+    periphery = new ethers.Contract(mainnet.periphery, peripheryAbi, signer);
```
abylaw: it doesn't look like we call the divider of periphery in the code below, so maybe it would be better to keep the `periphery`/`divider` contracts within the `for` loop?

abylaw:
Left a few small stylistic/code organization comments. Otherwise, looks straightforward and GTM!

pkg/deployments/tasks/20220517-long-wsteth-adapter/index.js
fedealconada: but, if I use connect, wouldn't it be like an additional step? like:

```
      divider = new ethers.Contract(mainnet.divider, dividerAbi);
      divider.connect(signer);
```

pkg/deployments/tasks/20220517-long-wsteth-adapter/output/localhost/solcInputs/c685b0e941c886293d0434d81554143d.json
fedealconada: You're right and maybe I should have written it somewhere xD, I only committed the localhost because I have not yet deployed this to mainnet (we are still pending some checks on what the `stakeSize` will be). So, in the meantime, kind of to "prove" that I ran deployment on a hardhat fork on my local env, I decided to commit that. My plan was to remove those and replace them with the mainnet ones :)

pkg/deployments/tasks/20220517-long-wsteth-adapter/index.js
fedealconada: yess, that's my bad for copypasting from an older task haha, thanks!

pkg/deployments/tasks/20220517-long-wsteth-adapter/input.js
fedealconada: We could actually grab them from `hardhat.addresses.js` though that's not always up to date.  I think it's better to keep it as it is: each task with its own `input.js`, defining the addresses for 2 reasons:
- to kind of know which were the addresses when we executed this task
- a bit as a sanity check, forcing the developer that is creating the task to go grab the corresponding addresses
WDYT @abylaw ?

pkg/deployments/tasks/20220517-long-wsteth-adapter/index.js
abylaw: Ah, only in the case that you'd like to reuse a contract, as opposed to rewriting a variable with a new contract each time :) like in this case, since you've already initialized divider/periphery above, it would be `divider.connect(signer)`. More of an FYI in case you weren't already familiar with `connect`!

pkg/deployments/tasks/20220517-long-wsteth-adapter/output/localhost/solcInputs/c685b0e941c886293d0434d81554143d.json
abylaw: Ahhhhh ok, makes sense makes sense!

pkg/deployments/tasks/20220517-long-wsteth-adapter/index.js
fedealconada: ohhh, yesss, got it! check my last commit :)

pkg/deployments/tasks/20220517-long-wsteth-adapter/input.js
abylaw: Got it got it.

I've been thinking about maintenance since I've been writing docs :) If we upgrade the `periphery`/`poolManager`/`spaceFactory`, we might have to troll through each of the tasks to find out whether each task has a reference to an older address. 

If we instead centralized the addresses somewhere in the repo, then updating the address means that all tasks will automatically work with the newest version of the protocol. @fedealconada 

Don't think we have to do it now, more of a directional aspiration. I've been thinking it might even be helpful to centralize our addresses/ABIs somewhere so we'd don't have to go through individual repos (as much) either...

pkg/deployments/tasks/20220517-long-wsteth-adapter/index.js
```diff
@@ -0,0 +1,86 @@
...
+    const { contractName, deploymentParams, target } = adapter;
+    const { underlying, ifee, adapterParams } = deploymentParams;
+    
+    divider = divider.connect(deployerSigner);
+    periphery = periphery.connect(deployerSigner);
+
+    console.log(`\nDeploy ${contractName}`);
+    const { address: adapterAddress } = await deploy(contractName, {
+      from: deployer,
+      args: [divider.address, target.address, underlying, ifee, [...Object.values(adapterParams)]],
```
abylaw, line 45: nit: `mainnet.divider`?
Then we don't need to initialize or connect the contracts above :)

abylaw (approved):

PR #249, opened 2022-05-18
GEN-942: crop adapter reconcile

fedealconada:
## Motivation

Through our Bug Bounty programme, we've been made aware of a bug affecting our Crop/sAdapter contracts. Basically, YT holders could disrupt the balance for rewards kept by `CropAdapter` and `CropsAdapter` if they were holding their YTs past maturity leading to a potential dilution of next Series YT holders rewards.

## Solution

Implemented a `reconcile` function which can be called by anyone and, ideally, should be called right after maturity which reconciles the `tBalance` (target balance of each user) for a given maturity.  

**To reviewers: better to start reviewing `CropAdater.sol` first and its tests (`CropAdapter.t.sol`) since `CropsAdapter` is almost the same.**

## Implementer Checklist


pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -87,4 +125,8 @@ abstract contract CropAdapter is BaseAdapter {
         if (divider != msg.sender) revert Errors.OnlyDivider();
         _;
     }
+
+    /* ========== LOGS ========== */
+
+    event Reconciled(address indexed usr);
```
jparklev:
```suggestion
    event Reconciled(address indexed usr, uint256 tBal);
```

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -41,21 +43,57 @@ abstract contract CropAdapter is BaseAdapter {
...
+    /// @param _maturities Maturities of the series that we want to reconcile users on.
+    function reconcile(address[] calldata _usrs, uint256[] calldata _maturities) public {
+        for (uint256 i = 0; i < _usrs.length; i++) {
+            address usr = _usrs[i];
+            for (uint256 j = 0; j < _maturities.length; j++) {
+                (, , address yt, , , , , uint256 mscale, ) = Divider(divider).series(address(this), _maturities[j]);
+
+                // We don't want to reconcile users if maturity has not been reached or if they have already been reconciled
+                if (
+                    _maturities[j] < block.timestamp && ERC20(yt).balanceOf(usr) > 0 && !reconciled[usr][_maturities[j]]
```
jparklev: Consider storing `ERC20(yt).balanceOf(usr)` as a local variable above since it's potentially used twice

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -41,21 +43,57 @@ abstract contract CropAdapter is BaseAdapter {
...
 
+    /// @notice Reconciles users target balances to avoid delution of next Series YT holders.
+    /// This function should be called right after a Series matures.
+    /// @param _usrs Users to reconcile
+    /// @param _maturities Maturities of the series that we want to reconcile users on.
+    function reconcile(address[] calldata _usrs, uint256[] calldata _maturities) public {
+        for (uint256 i = 0; i < _usrs.length; i++) {
+            address usr = _usrs[i];
+            for (uint256 j = 0; j < _maturities.length; j++) {
+                (, , address yt, , , , , uint256 mscale, ) = Divider(divider).series(address(this), _maturities[j]);
```
jparklev: Could use the `mscale` and `yt` functions on the Divider and I think save some gas

pkg/core/src/tests/adapters/CropAdapter.t.sol
```diff
@@ -210,10 +258,495 @@ contract CropAdapters is TestHelper {
...
         // 30 should go to alice and 70 to jim
-        reward.mint(address(adapter), 100 * 1e18);
+        reward.mint(address(cropAdapter), 100 * 1e18);
         jim.doCollect(yt);
         assertClose(reward.balanceOf(address(jim)), 120 * 1e18);
         alice.doCollect(yt);
         assertClose(reward.balanceOf(address(alice)), 96 * 1e18);
     }
+
+    function testFuzzCollectRewardSettleSeriesAndCheckTBalanceIsZero(uint256 tBal) public {
```
jparklev, line 250: Nice addition

pkg/core/src/tests/adapters/CropAdapter.t.sol
```diff
@@ -210,10 +258,495 @@ contract CropAdapters is TestHelper {
...
+
+        // reconcile Bob's position
+        assertEq(cropAdapter.tBalance(address(bob)), (40 * tBal) / 100);
+        assertEq(cropAdapter.reconciledAmt(address(bob)), 0);
+        uint256[] memory maturities = new uint256[](1);
+        maturities[0] = maturity;
+        address[] memory users = new address[](1);
+        users[0] = address(bob);
+        cropAdapter.reconcile(users, maturities);
+        cropAdapter.reconcile(users, maturities);
```
jparklev: Duplicate?

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -41,21 +43,57 @@ abstract contract CropAdapter is BaseAdapter {
         uint256 amt,
         bool join
     ) public override onlyDivider {
+        if (reconciledAmt[_usr] > 0) {
+            if (!join) {
+                if (amt < reconciledAmt[_usr]) {
+                    reconciledAmt[_usr] -= amt;
+                    amt = 0;
+                } else {
+                    amt -= reconciledAmt[_usr];
```
jparklev: We could consider moving all of this down to the conditional below like 
```
        if (amt > 0) {
            if (join) {
            ...
            } else {
                if (amt < reconciledAmt[_usr]) {
                    reconciledAmt[_usr] -= amt;
                    amt = 0;
                } else {
            ...
...
```

jparklev: Although I guess we'd have to add another `if (amt > 0) {`, which is useful for saving gas

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -41,21 +43,57 @@ abstract contract CropAdapter is BaseAdapter {
         uint256 amt,
         bool join
     ) public override onlyDivider {
+        if (reconciledAmt[_usr] > 0) {
+            if (!join) {
+                if (amt < reconciledAmt[_usr]) {
+                    reconciledAmt[_usr] -= amt;
```
jparklev: this can be `unchecked` since we know it won't overflow

pkg/core/src/tests/adapters/CropAdapter.t.sol
```diff
@@ -210,10 +258,495 @@ contract CropAdapters is TestHelper {
...
+        assertEq(cropAdapter.reconciledAmt(address(alice)), 0);
+        assertEq(cropAdapter.reconciledAmt(address(bob)), (40 * tBal) / 100);
+
+        // Both Alice and Bob should get their rewards
+        assertClose(reward.balanceOf(address(alice)), 50 * 1e18);
+        assertClose(reward.balanceOf(address(bob)), 20 * 1e18);
+
+        // Alice should still be able to collect her rewards from
+        // the December series (30 reward tokens)
+        alice.doIssue(address(cropAdapter), maturity, 0 * 1e18);
```
jparklev: `0 * 1e18`?

pkg/core/src/tests/adapters/CropAdapter.t.sol
```diff
@@ -210,10 +258,495 @@ contract CropAdapters is TestHelper {
...
+    /*              Jan                                                          Dec (maturity)
+     *  Dec 2021 Series:  |-------------------------------|-------------------------------|
+     *
+     *                                              July (maturity)
+     *  July 2022 Series: |-------------------------------|--------->
+     *
+     *                                                  ^ July series
+     *                                                    reconcilation
+     */
+    function testFuzzReconcileWhenTwoSeriesOverlap(uint256 tBal) public {
```
jparklev:
```suggestion
    function testFuzzReconcileSingleSeriesWhenTwoSeriesOverlap(uint256 tBal) public {
```

pkg/core/src/tests/adapters/CropAdapter.t.sol
```diff
@@ -210,10 +258,495 @@ contract CropAdapters is TestHelper {
...
+
+        // Bob closes his YT position and rewards
+        bob.doCollect(yt);
+        assertEq(cropAdapter.tBalance(address(bob)), 0);
+        assertEq(ERC20(yt).balanceOf(address(bob)), 0);
+
+        // at this point, no rewards are left to distribute and
+        // - Alice has 30 reward tokens
+        // - Bob has 20 * 1e18 reward tokens
+        assertClose(ERC20(reward).balanceOf(address(cropAdapter)), 0 * 1e18);
```
jparklev:
```suggestion
        assertClose(ERC20(reward).balanceOf(address(cropAdapter)), 0);
```

pkg/core/src/tests/adapters/CropAdapter.t.sol
```diff
@@ -210,10 +258,495 @@ contract CropAdapters is TestHelper {
...
+        // late reconcile Alice's position
+        assertEq(cropAdapter.reconciledAmt(address(alice)), 0);
+        assertEq(cropAdapter.tBalance(address(alice)), (60 * tBal) / 100);
+        uint256[] memory maturities = new uint256[](1);
+        maturities[0] = maturity;
+        address[] memory users = new address[](1);
+        users[0] = address(alice);
+        cropAdapter.reconcile(users, maturities);
+        assertEq(cropAdapter.tBalance(address(alice)), 0);
+        assertEq(cropAdapter.reconciledAmt(address(alice)), (60 * tBal) / 100);
```
jparklev: Should we check that Alice's `reconciledAmt` goes to zero if she collects? I think we should also have a test that calls `reconcile` after a changing scale value and a collect or two pre-maturity, then checks `reconciledAmt`  before or after collecting again after maturity (should zero out exactly, I think)

jparklev (commented):
Nice one! This feels actually like a nicer solution to me now than it did last week – think the tests improve my confidence. 

I mainly reviewed `CropAdapter` as you recommended. A few nits and potential re-orderings, and then a new test suggestion for changing scale

pkg/core/src/tests/adapters/CropAdapter.t.sol
fedealconada: We actually had it on `Divider.t.sol` but moved it here :)

pkg/core/src/tests/adapters/CropAdapter.t.sol
fedealconada: Nope, this test is called `testFuzzReconcileMoreThanOnce` and it's checking that calling reconcile more than once for a user for the same maturity should not affect... It wouldn't revert but it should not reconcile reconcile it again.

pkg/core/src/tests/adapters/CropAdapter.t.sol
fedealconada: lol, i knew one of you was gonna say something about it xD, i put them that way because on these tests we are always using the following format:

- 60 * 1e18
- 40 * 1e18

so it was more so as to keep that format

but i  can just write 0 if you prefer!

pkg/core/src/tests/adapters/CropAdapter.t.sol
fedealconada: pushed!

pkg/core/src/tests/adapters/CropAdapter.t.sol
fedealconada: pushed!

pkg/core/src/adapters/CropAdapter.sol
fedealconada: pushed!

pkg/core/src/tests/adapters/CropAdapter.t.sol
jparklev: 🏎️

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -47,15 +49,50 @@ abstract contract CropAdapter is BaseAdapter {
...
-                totalTarget -= amt;
-                tBalance[_usr] -= amt;
+                if (reconciledAmt[_usr] > 0) {
+                    if (amt < reconciledAmt[_usr]) {
+                        unchecked {
+                            reconciledAmt[_usr] -= amt;
+                        }
+                        amt = 0;
+                    } else {
+                        amt -= reconciledAmt[_usr];
```
jparklev: can be `unchecked` as well, i believe

fedealconada: yes sirrr

fedealconada: pushed!

pkg/core/src/tests/test-helpers/TestHelper.sol
```diff
@@ -277,10 +295,10 @@ contract TestHelper is DSTest {
         BaseFactory.FactoryParams memory factoryParams = BaseFactory.FactoryParams({
             stake: address(stake),
             oracle: ORACLE,
-            ifee: ISSUANCE_FEE,
+            ifee: 0,
             stakeSize: STAKE_SIZE,
             minm: MIN_MATURITY,
-            maxm: MAX_MATURITY,
+            maxm: 48 weeks, // 1 year
```
abylaw: 52 weeks?? (・・;)ゞ

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -47,15 +49,52 @@ abstract contract CropAdapter is BaseAdapter {
...
+                    totalTarget -= amt;
+                    tBalance[_usr] -= amt;
+                }
             }
         }
-
         rewarded[_usr] = tBalance[_usr].fmulUp(share, FixedMath.RAY);
     }
 
+    /// @notice Reconciles users target balances to avoid delution of next Series YT holders.
```
abylaw: Attempted to make it a little more explicit as to what we mean by reconciliation:
```suggestion
    /// @notice Reconciles users target balances to zero by distributing rewards on their holdings, to avoid dilution of next Series' YT holders.
```

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -47,15 +49,52 @@ abstract contract CropAdapter is BaseAdapter {
...
+    /// This function should be called right after a Series matures.
+    /// @param _usrs Users to reconcile
+    /// @param _maturities Maturities of the series that we want to reconcile users on.
+    function reconcile(address[] calldata _usrs, uint256[] calldata _maturities) public {
+        for (uint256 i = 0; i < _usrs.length; i++) {
+            address usr = _usrs[i];
+            for (uint256 j = 0; j < _maturities.length; j++) {
+                address yt = Divider(divider).yt(address(this), _maturities[j]);
```
abylaw: for my own education: what happens when we try to access a mapping with an invalid key? will the evm revert the transaction?

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -47,15 +49,52 @@ abstract contract CropAdapter is BaseAdapter {
...
+                uint256 ytBal = ERC20(yt).balanceOf(usr);
+                // We don't want to reconcile users if maturity has not been reached or if they have already been reconciled
+                if (_maturities[j] < block.timestamp && ytBal > 0 && !reconciled[usr][_maturities[j]]) {
+                    _distribute(usr);
+                    uint256 tBal = ytBal.fdiv(Divider(divider).mscale(address(this), _maturities[j]));
```
abylaw: could we save gas by inverting our for-loops & calling `mscale` only once per maturity?

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -47,15 +49,52 @@ abstract contract CropAdapter is BaseAdapter {
...
+                    totalTarget -= tBal;
+                    tBalance[usr] -= tBal;
+                    reconciledAmt[usr] += tBal;
+                    reconciled[usr][_maturities[j]] = true;
+                    emit Reconciled(usr, tBal);
```
abylaw: maybe it would make sense to include the series' maturity in the emitted event? since reconciliation is indexed by series/user.

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -47,15 +49,52 @@ abstract contract CropAdapter is BaseAdapter {
...
+            address usr = _usrs[i];
+            for (uint256 j = 0; j < _maturities.length; j++) {
+                address yt = Divider(divider).yt(address(this), _maturities[j]);
+                uint256 ytBal = ERC20(yt).balanceOf(usr);
+                // We don't want to reconcile users if maturity has not been reached or if they have already been reconciled
+                if (_maturities[j] < block.timestamp && ytBal > 0 && !reconciled[usr][_maturities[j]]) {
+                    _distribute(usr);
+                    uint256 tBal = ytBal.fdiv(Divider(divider).mscale(address(this), _maturities[j]));
+                    totalTarget -= tBal;
+                    tBalance[usr] -= tBal;
```
abylaw: ~~wondering if it's alright that `totalTarget`/`tBalance[usr]` are not quite zero after reconciliation?  (My understanding is that since `mscale` will always be larger than the scale at issuance when notify is first called, so `tBal_now < tBal_issuance`)~~

~~Could it lead to very gradual dilution of rewards over time as the adapter ages?~~

edit: to clarify my line of thinking above, in a scenario where a user either:
1. never calls `collect`
2. only calls `collect` after reconciliation

=> our accounting for `tBalance`/`totalTarget` will be off 
=> reward dilution?

e.g.
-> issues 10 YTs @ scale 1, `tBalance[usr] = 10`, `totalTarget =  originalTotalTarget + 10`
-> scale grows to 2
-> series matures
-> reconcile is called: `tBalance[usr] = 10 - 10 / 2 = 5`, `reconciledAmt[usr] = 5`, `totalTarget = originalTotalTarget + 5`
-> even if `collect` is called afterwards (like in scenario 2), `tBalance[usr]` remains `5`, while `reconciledAmt[usr] = 5 - 5 = 0`

even in less extreme scenarios where `collect` is called before reconciliation, there will always be a small gap between the `maxscale` (which reconciliation uses) & the `scale` value which `collect` uses. this implies that even after reconciliation `tBalance[usr]` will never quite be zero, and by extension, `totalTarget` will be slightly too large, and that means  rewards get diluted over time.

(fwiw, i also think this was the case before the reconciliation fix, so maybe this was always a feature of crop adapter?)

Or I might be overlooking something totally obvious. Let me know...

abylaw (commented):
First pass of comments, will look at the tests afterwards!

pkg/core/src/tests/adapters/CropAdapter.t.sol
```diff
@@ -210,10 +258,589 @@ contract CropAdapters is TestHelper {
...
+        assumeBounds(tBal);
+        // current date is 1st Jan 2021 00:00 UTC
+        hevm.warp(1609459200);
+
+        // sponsor a 12 month maturity (December 1st 2021)
+        uint256 maturity = getValidMaturity(2021, 12);
+        hevm.startPrank(address(alice));
+        (, address yt) = periphery.sponsorSeries(address(cropAdapter), maturity, true);
+        cropAdapter.setScale(1e18);
+        hevm.stopPrank();
```
abylaw: `alice.doSponsorSeries`? since we seem to use these 4 lines quite frequently :)

abylaw (commented):
LGTM! I think I only have one outstanding concern, which is that it seems to me that CropAdapter is still vulnerable to reward dilution...but as mentioned before, I might be misunderstanding something...

in any case, it may be nice to add a few tests where `scale` values change, and `collect` is called but not necessary immediately before reconciliation (or it is called after reconciliation, or not at all), and making sure `tBalance[usr]` & `totalTarget` go to `0`.

pkg/core/src/adapters/CropAdapter.sol
fedealconada: it does not, mappings are initialised with zero values

pkg/core/src/adapters/CropAdapter.sol
fedealconada: we could! :)

pkg/core/src/tests/test-helpers/TestHelper.sol
fedealconada: hahahahahha

pkg/core/src/tests/adapters/CropAdapter.t.sol
fedealconada: That's a good point. You will see in the code a lot of `doSOMETHING()` functions to simulate the impersonation, that was because there was no impersonation feature yet in Foundry. With the last release, they incorporated the `prank()` and `prankStart()` so, for new code, we are using those ones.  For this particular case, I could actually be using only `hevm.prank()` instead `startPrank()` and `stopPrank()`. But, anyways, yes, it adds some more lines than just using the `doSomething()` workaround.
We have a quality ticket to then polish these things :)

pkg/core/src/adapters/CropAdapter.sol
fedealconada: This was a great catch @abylaw ! I've changed the `reconcile()` function to use the `lscale` (last scale value for a given user on a given maturity) value of a user instead of using the `mscale` (scale at maturity).  This should be fixing the issue about `tBalance` / `totalTarget` being off.
I've included a test case (`testFuzzCantDiluteRewardsInProportionalDistributionWithScaleChangesAndCollectAfterReconcile`) that covers the scenario of users issuing, never collecting and then reconciling.  

Let me know what you think!

abylaw: > This was a great catch @abylaw !

😄😄 Thanks!

> I've changed the reconcile() function to use the lscale (last scale value for a given user on a given maturity) value of a user instead of using the mscale (scale at maturity). 

Nice! I think that makes sense 👍 👍 

~Although I think there may still a scenario where `reconciledAmt` is off (need to think about implications):~

~...same as above...
-> user calls `collect`
-> their YTs will be redeemed, then notify is called with `ytBal/maxscale`: https://github.com/sense-finance/sense-v1/blob/dev/pkg/core/src/Divider.sol#L486~

~if the user has never collected, then `reconciledAmt = ytBal/initscale`, which is larger than `ytBal/maxScale`. So  `reconciledAmt` ends up > 0. Possibly when we subtract it from `tBalance` we use `lscale` but when we add it to reconciledAmt we should use `maxscale`?~

edit: discussed with fede on discord - no problems here! i failed to take into account the second notify after the redeem.

abylaw (approved):

pkg/core/src/adapters/CropAdapter.sol
jparklev: Great shout on this one, btw @abylaw

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -47,15 +49,52 @@ abstract contract CropAdapter is BaseAdapter {
                 totalTarget += amt;
                 tBalance[_usr] += amt;
             } else {
-                // else `exit`
-                totalTarget -= amt;
-                tBalance[_usr] -= amt;
+                if (reconciledAmt[_usr] > 0) {
+                    if (amt < reconciledAmt[_usr]) {
+                        unchecked {
```
jparklev: One practice that we might want to start from brock's suggestions: everytime we use `unchecked`, we have a comment explicitly mentioning why overflow is not a concern

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -47,15 +49,52 @@ abstract contract CropAdapter is BaseAdapter {
                 totalTarget += amt;
                 tBalance[_usr] += amt;
             } else {
-                // else `exit`
-                totalTarget -= amt;
-                tBalance[_usr] -= amt;
+                if (reconciledAmt[_usr] > 0) {
```
jparklev: I think a little comment on what `reconciled` means might be helpful for anybody reading out code!

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -71,25 +71,25 @@ abstract contract CropAdapter is BaseAdapter {
...
     function reconcile(address[] calldata _usrs, uint256[] calldata _maturities) public {
-        for (uint256 i = 0; i < _usrs.length; i++) {
-            address usr = _usrs[i];
-            for (uint256 j = 0; j < _maturities.length; j++) {
-                address yt = Divider(divider).yt(address(this), _maturities[j]);
-                uint256 ytBal = ERC20(yt).balanceOf(usr);
+        for (uint256 j = 0; j < _maturities.length; j++) {
+            for (uint256 i = 0; i < _usrs.length; i++) {
+                address usr = _usrs[i];
+                uint256 ytBal = ERC20(Divider(divider).yt(address(this), _maturities[j])).balanceOf(usr);
```
jparklev: This could be pulled above the `for (uint256 i = 0; i < _usrs.length; i++)` line, right?

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -71,25 +71,25 @@ abstract contract CropAdapter is BaseAdapter {
...
                 // We don't want to reconcile users if maturity has not been reached or if they have already been reconciled
                 if (_maturities[j] < block.timestamp && ytBal > 0 && !reconciled[usr][_maturities[j]]) {
                     _distribute(usr);
-                    uint256 tBal = ytBal.fdiv(Divider(divider).mscale(address(this), _maturities[j]));
+                    uint256 tBal = ytBal.fdiv(Divider(divider).lscales(address(this), _maturities[j], address(usr)));
```
jparklev: I think `usr` is already an address!

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -47,15 +49,52 @@ abstract contract CropAdapter is BaseAdapter {
...
+    /// This function should be called right after a Series matures.
+    /// @param _usrs Users to reconcile
+    /// @param _maturities Maturities of the series that we want to reconcile users on.
+    function reconcile(address[] calldata _usrs, uint256[] calldata _maturities) public {
+        for (uint256 j = 0; j < _maturities.length; j++) {
+            for (uint256 i = 0; i < _usrs.length; i++) {
+                address usr = _usrs[i];
+                uint256 ytBal = ERC20(Divider(divider).yt(address(this), _maturities[j])).balanceOf(usr);
+                // We don't want to reconcile users if maturity has not been reached or if they have already been reconciled
+                if (_maturities[j] < block.timestamp && ytBal > 0 && !reconciled[usr][_maturities[j]]) {
```
jparklev:
```suggestion
                if (_maturities[j] <= block.timestamp && ytBal > 0 && !reconciled[usr][_maturities[j]]) {
```

jparklev: Potentially hacky way to get rid of the `reconciled` mapping: don't let `reconciledAmt` got to `0` after it has been set, only bring it down to `1` at a minimum after it has been set, then check to see if `reconciledAmt` is exectly `0` here

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -47,15 +49,52 @@ abstract contract CropAdapter is BaseAdapter {
                 totalTarget += amt;
                 tBalance[_usr] += amt;
             } else {
-                // else `exit`
-                totalTarget -= amt;
-                tBalance[_usr] -= amt;
+                if (reconciledAmt[_usr] > 0) {
+                    if (amt < reconciledAmt[_usr]) {
```
jparklev: Does it save any gas if we store `reconciledAmt[_usr]` as a variable?

pkg/core/src/tests/adapters/CropAdapter.t.sol
```diff
@@ -210,10 +258,671 @@ contract CropAdapters is TestHelper {
...
+        users[0] = address(bob);
+        cropAdapter.reconcile(users, maturities);
+        assertEq(cropAdapter.tBalance(address(bob)), (40 * tBal) / 100);
+
+        bob.doCollect(yt);
+        assertClose(reward.balanceOf(address(bob)), 24 * 1e18);
+
+        // should not emit a Reconciled() event
+        hevm.expectEmit(true, false, false, false);
+        emit Reconciled(address(bob));
```
jparklev: Should there be a call below this that fails to emits an event?

pkg/core/src/tests/adapters/CropAdapter.t.sol
```diff
@@ -210,10 +258,671 @@ contract CropAdapters is TestHelper {
...
+        assertClose(ERC20(reward).balanceOf(address(bob)), 40 * 1e18);
+
+        // sponsor new Series
+        uint256 newMaturity = getValidMaturity(2021, 11);
+        hevm.startPrank(address(alice));
+        (, address newYt) = periphery.sponsorSeries(address(cropAdapter), newMaturity, true);
+        hevm.stopPrank();
+
+        alice.doIssue(address(cropAdapter), newMaturity, (60 * tBal) / 100);
+        bob.doIssue(address(cropAdapter), newMaturity, (40 * tBal) / 100);
```
jparklev: Should we confirm that `reconcileAmt` decreases if we issue, then combine the full amount? One bit of weirdness there is that then the collect of the user's previously uncollected `yt`s will decrement the `tBalance` rather than `reconcileAmt` – is that ok?

jparklev (commented):
A few final notes, but otherwise LGTM

pkg/core/src/adapters/CropAdapter.sol
fedealconada: Nope, because line 83 depends on `_usrs[i]`

pkg/core/src/adapters/CropAdapter.sol
fedealconada: Hmm... I think we can't do that because `reconcileAmt` does not discriminate between maturities and we need to know if a user has been reconciled or not for a given maturity.

pkg/core/src/tests/adapters/CropAdapter.t.sol
fedealconada: true! but given that I made [this](https://github.com/sense-finance/sense-v1/pull/249#discussion_r887179673) change you've suggested this test would no longer be needed

pkg/core/src/adapters/CropAdapter.sol
fedealconada: pushed this, let me know what you think

pkg/core/src/tests/adapters/CropAdapter.t.sol
fedealconada: > Should we confirm that reconcileAmt decreases if we issue

hmm.. `reconcileAmt` does not decrease when we issue, it only decreases when `notify()` is called with `false` 

> then combine the full amount?

this test is [combining the full amount at the bottom](https://github.com/sense-finance/sense-v1/blob/140e8b4e36418b36407cfd85054da7bcaf9941d8/pkg/core/src/tests/adapters/CropAdapter.t.sol#L573), or you refer to something else?

> One bit of weirdness there is that then the collect of the user's previously uncollected yts will decrement the tBalance rather than reconcileAmt

when user collects (independently if it's the previous uncollected YTs or not), that triggers the `notify()` with `false` which [decreases the `reconciledAmt` (if any)](https://github.com/sense-finance/sense-v1/blob/140e8b4e36418b36407cfd85054da7bcaf9941d8/pkg/core/src/adapters/CropAdapter.sol#L70), [the `tBalance`](https://github.com/sense-finance/sense-v1/blob/140e8b4e36418b36407cfd85054da7bcaf9941d8/pkg/core/src/adapters/CropAdapter.sol#L74)  and [the `totalTarget`](https://github.com/sense-finance/sense-v1/blob/140e8b4e36418b36407cfd85054da7bcaf9941d8/pkg/core/src/adapters/CropAdapter.sol#L73) if `amt` > 0.
(unless I'm missing something from your comment)

jparklev: > hmm.. reconcileAmt does not decrease when we issue, it only decreases when notify() is called with false

I was thinking the sequence of issuing -> then combining. That combine would decrease `reconcileAmt`, right?

> when user collects (independently if it's the previous uncollected YTs or not), that triggers the notify() with false which [decreases the reconciledAmt (if any)](https://github.com/sense-finance/sense-v1/blob/140e8b4e36418b36407cfd85054da7bcaf9941d8/pkg/core/src/adapters/CropAdapter.sol#L70), [the tBalance](https://github.com/sense-finance/sense-v1/blob/140e8b4e36418b36407cfd85054da7bcaf9941d8/pkg/core/src/adapters/CropAdapter.sol#L74) and [the totalTarget](https://github.com/sense-finance/sense-v1/blob/140e8b4e36418b36407cfd85054da7bcaf9941d8/pkg/core/src/adapters/CropAdapter.sol#L73) if amt > 0.
(unless I'm missing something from your comment)

I was just noticing that it's a bit odd that `reconcileAmt` can decrease from future Series `combine` or `collect` if the original Series yts are never collected. Then when we collect on those original yts, it'll decrement from `tBalance`. Think it's fine in the end, just didn't realize it initially. 

Does that make sense?

fedealconada: > I was thinking the sequence of issuing -> then combining. That combine would decrease reconcileAmt, right?

added! FYI other tests already had this check :)

> I was just noticing that it's a bit odd that reconcileAmt can decrease from future Series `combine` or `collect` if the original Series yts are never collected. Then when we collect on those original yts, it'll decrement from tBalance. Think it's fine in the end, just didn't realize it initially.
Does that make sense?

yes, just note that "if the original Series yts are never collected" -> this should not be a problem since even though the user has not collected, we should have had reconciled the user at maturity

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -45,31 +45,32 @@ abstract contract CropAdapter is BaseAdapter {
...
-                    if (amt < uReconciledAmt) {
-                        unchecked {
-                            uReconciledAmt -= amt;
-                        }
-                        amt = 0;
-                    } else {
+            uint256 uReconciledAmt = reconciledAmt[_usr];
+            // if reconciled amount > 0, we need to adjust the `amt` received
+            // both when join is true and false
+            if (uReconciledAmt > 0) {
```
jparklev: Let me know if my understanding is correct. The change here makes it so, if `join` is true, we decrement `reconciledAmt`, but don't touch `amt` (so that it still adds to `tBalance`).

If that's the right understanding, is the following sequence possible?
1. User gets YTs somehow then don't redeem them after maturity (let's say `tBalance` is now 5 for the user)
2. We reconcile their position (`reconciledAmt` is now 5 & `tBalance` is 0 for the user)
3. User issues for a new Series (let's say `amt` is 4, so then `reconciledAmt` becomes 1 and `tBalance` becomes 4)
4. They combine the newly issued tokens (`amt` is 4 but `join` is false, so reconciledAmt` becomes 0 and `tBalance` becomes 1)

Are these steps correct?

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -38,48 +38,122 @@ abstract contract CropAdapter is BaseAdapter {
...
     }
 
+    // What is reconciled amount? It is a user's YT balance in target terms of a matured Series
+    // that it should have been decremented from the `tBalance` but it was not because the user
+    // did not redeemed their YTs (yet).
+    // Note that, when reconciliing, the user receives the corresponding rewards until maturity.
+    // The `reconciledAmt` allows us to prevent diluting other user's future Series rewards because
+    // we can keep track of those amounts and subtract them from the `tBalance` when needed.
+    // Reconciled amounts can be thought as a debt on the `tBalance` that the user will repay
+    // when issuing (join = true)
```
jparklev: "when issuing" or when receiving YTs from transfers!

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -38,48 +38,122 @@ abstract contract CropAdapter is BaseAdapter {
...
+                        // if debt is smaller than `amt` being redeemed/combined, we can just settle it
+                        // because... ELABORATE
+                        // TODO: (what happens if scale changes?? because here we
+                        // are always assuming scale is 1 but `amt` is in YT and reconciled is in target)
+                        uReconciledAmt = 0;
+                    } else {
+                        uReconciledAmt -= amt;
+                    }
+                }
                 if (tBalance[_usr] >= amt) {
```
jparklev: When would `amt` be <= `tBalance`?

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -38,48 +38,122 @@ abstract contract CropAdapter is BaseAdapter {
...
+                tBalance[_usr] += amt;
             } else {
+                // when a user redeems/combines we decrement the user's targetBalance by the `amt` being
+                // redeemed/combined but
+                if (uReconciledAmt > 0) {
+                    if (uReconciledAmt <= amt) {
+                        // if debt is smaller than `amt` being redeemed/combined, we can just settle it
+                        // because... ELABORATE
+                        // TODO: (what happens if scale changes?? because here we
+                        // are always assuming scale is 1 but `amt` is in YT and reconciled is in target)
```
jparklev: isn't the YT amt also in "Target amount backing the YT"?

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -38,48 +38,122 @@ abstract contract CropAdapter is BaseAdapter {
...
-                reconciledAmt[_usr] = uReconciledAmt;
-            }
+            emit Log(uReconciledAmt);
             if (join) {
-                if (amt > 0) {
-                    totalTarget += amt;
-                    tBalance[_usr] += amt;
+                // if user has debt (`reconciledAmt` > 0) we use the `amt` to pay it off (partially
+                // or completely, dependening on the case)
+                // but, note that the `amt` var is not modified because the user needs to have his `tBalance`
```
jparklev: My only question is – if joins reduce `reconciledAmt`, then what happens if a user has old YTs, issues for a new series (decrements `reconciledAmt` and adds `tBalance`), then collects on their old YTs (would it decrement `tBalance` or am i missing something)

jparklev (commented):
A few quick thoughts in case they're useful before i sign off

pkg/core/src/adapters/CropAdapter.sol
fedealconada: yes yes, got confused for a moment!

pkg/core/src/adapters/CropAdapter.sol
fedealconada: yes! just for simplicity i was writing just "issue" but this would be for all `join == true`

pkg/core/src/adapters/CropAdapter.sol
fedealconada: yeah, so I've been thinking a bit more last night and this comment is connected to what you said during the jam call about which is the problem of not being able to differentiate the maturities.
so, these new changes do not solve the issue basically because a user can be decreasing his `tBalance` when combining/redeeming from an older Series, and that's wrong. 

But, 2 comments:
(+) This solution seems like it prevents other users from being diluted
(-) If the user does not collect "in order" (he collects first his old YTs) this would affect his `tBalance` making himself to get diluted and benefiting the rest of the YT holders.

PR #250, opened 2022-05-25
GEN-961: EIP 4626

fedealconada:
## Motivation

Sense wants to support [EIP-4626](https://eips.ethereum.org/EIPS/eip-4626) to benefit from the advantages of its standardisation. This way, integrations will be faster and safer since there won't be a need of implementing customised adapters (unless its preferred).  Later on, we could have a 4626 factory which allows anyone to deploy adapters whose target is an ERC4626.

## Solution

We've created the adapter `ERC4626Adapter.sol` whose target is an ERC4626. Moreover, we've adjusted our tests to also run over these kind of adapters.

## Implementer Checklist


fedealconada:
need to fix some tests but it's ready for review @abylaw @jparklev

pkg/core/src/tests/adapters/CropAdapter.t.sol
```diff
@@ -12,14 +12,23 @@ import { Divider } from "../../Divider.sol";
...
-import { TestHelper } from "../test-helpers/TestHelper.sol";
+import { TestHelper, MockTargetLike } from "../test-helpers/TestHelper.sol";
 
 contract CropAdapters is TestHelper {
     using FixedMath for uint256;
 
+    function setUp() public virtual override {
+        super.setUp();
+
+        // freeze scale to 1e18 (only for no 4626 targets)
```
jparklev: So I can better understand, could you expand on the reasoning for this?

pkg/core/src/adapters/ERC4626CropsAdapter.sol
```diff
@@ -0,0 +1,58 @@
...
+        uint128 _ifee,
+        AdapterParams memory _adapterParams,
+        address[] memory _rewardTokens
+    ) CropsAdapter(_divider, _target, address(ERC4626(_target).asset()), _ifee, _adapterParams, _rewardTokens) {
+        BASE_UINT = 10**ERC4626(target).decimals();
+        ERC20(underlying).approve(target, type(uint256).max);
+    }
+
+    function scale() external override returns (uint256) {
+        return ERC4626(target).convertToAssets(BASE_UINT);
```
jparklev: I've learned something about the expected usage of these functions since working on this last! Namely that some Targets are expected to have variable exchange rates depending on the size of the redemption. ie, the more you redeem, the worse price you get. And the `convertTo` functions are supposed to take that into account and return a conversion from one asset to another for the  “average user”

In the case of a variable ex rate, a single scale value doesn't work perfectly, but I think we can still get away with our pattern as long as we err on the conservative side. I'm thinking we do `convertToAssets` with the entire Target balance held in the adapter, then calculate the ex rate for one Target from there. Then, it's kind of a worst case exchange rate if we needed to unwrap all of the Target at once. Does this seem like the right idea to you?

Then also, `convertTo` converts to an asset balance, which will be in the decimals of the underlying in this case, and that could be something other than `18`, right?

pkg/core/src/tests/GYTManager.t.sol
```diff
@@ -299,4 +301,23 @@ contract GYTsManager is TestHelper {
     //     assertEq(yieldBalanceAfter, aliceYieldBalance);
     //     assertTrue(tBalanceAfter > tBalanceBefore);
     // }
+
+    function calculateExcess(
```
jparklev: tbh we can probably just remove any files related to the gyt contract at this point. wdyt?

jparklev (commented):
Haven't fully reviewed yet, just wanted to drop in a question and share something that i'd learned.

pkg/core/src/tests/adapters/CropAdapter.t.sol
fedealconada: This was basically just a refactoring. All tests were having a `setScale(1e18)` so instead f having it on each test I moved it to the `setUp()` call.

pkg/core/src/adapters/ERC4626CropsAdapter.sol
fedealconada: Great point, didn't think about this. So:
> Does this seem like the right idea to you?
yes, sounds good, so we should be doing something like:
```
uint256 assets = t.convertToAssets(t.totalSupply());
return assets.fdiv(t.totalSupply(), BASE_UINT)
```

> Then also, `convertTo` converts to an asset balance, which will be in the decimals of the underlying in this case, and that could be something other than 18, right?

Yes, it can. But this is already handled by the `ERC4626.sol` contract, right? Not sure if you are asking to double check or because you had a concern. The `ERC4626` implementation we're using (Rari's one) uses the underlying's decimals to set the target's decimals.

jparklev: > yes, sounds good, so we should be doing something like:

So I was thinking
```
uint256 targetBal = t.balanceOf(address(this);
uint256 assets = t.convertToAssets(targetBal));
return assets.fdiv(targetBal, BASE_UINT)
```

But I just realized there's an issue where a flashloan can change the scale value during it's execution 🙃 (since balanceOf would decrease during the loan). I think what you have makes sense, though I haven't thought about it too deeply yet and we might need circle back

> The ERC4626 implementation we're using (Rari's one) uses the underlying's decimals to set the target's decimals.

Yep, but we expect scale to always be in 18 decimals!

pkg/core/src/tests/adapters/CropAdapter.t.sol
jparklev: Right, I wonder if we should just always use the mock 4626 adapter by default to test other contracts (divider, periphery, etc), then we can test different ones for adapter-specific tests?

fedealconada: Yeah, I thought about that but then just decided to have them both (ERC4626 and non-4626) running the whole test suite. My thought was more in the line of, the more tests we have the better. So we run all the test suite with one and with the other. Maybe, a little disadvantage of this approach is that I had to add some conditionals in several places to behave accordingly (you will in some places `if (is4626) ...`).

pkg/core/src/adapters/ERC4626CropAdapter.sol
```diff
@@ -0,0 +1,61 @@
...
+    constructor(
+        address _divider,
+        address _target,
+        uint128 _ifee,
+        AdapterParams memory _adapterParams,
+        address _reward
+    ) CropAdapter(_divider, _target, address(ERC4626(_target).asset()), _ifee, _adapterParams, _reward) {
+        uint256 tDecimals = ERC4626(target).decimals();
+        BASE_UINT = 10**tDecimals;
+        SCALE_FACTOR = 10**(18 - tDecimals); // we assume targets decimals <= 18
```
abylaw: should we add a explicit check for `tDecimals > 18` w/ a nice error message?

pkg/core/src/adapters/ERC4626Adapter.sol
```diff
@@ -0,0 +1,61 @@
...
+
+    constructor(
+        address _divider,
+        address _target,
+        uint128 _ifee,
+        AdapterParams memory _adapterParams
+    ) BaseAdapter(_divider, _target, address(ERC4626(_target).asset()), _ifee, _adapterParams) {
+        uint256 tDecimals = ERC4626(target).decimals();
+        BASE_UINT = 10**tDecimals;
+        SCALE_FACTOR = 10**(18 - tDecimals); // we assume targets decimals <= 18
```
abylaw: same comment about explicitly checking that `tDecimals <= 18`

pkg/deployments/deploy/simulated/04-adapters.js
```diff
@@ -209,12 +230,12 @@ module.exports = async function () {
...
   }
 
-  async function setScale(adapterAddress) {
+  async function setScale(adapterAddress, is4626) {
     const { abi: adapterAbi } = await deployments.getArtifact("MockAdapter");
     const adapter = new ethers.Contract(adapterAddress, adapterAbi, signer);
     const scale = await adapter.callStatic.scale();
     log(`-> scale: ${scale.toString()}`);
-    if (!scale.eq(ethers.utils.parseEther("1.1"))) {
+    if (!is4626 && !scale.eq(ethers.utils.parseEther("1.1"))) {
```
abylaw: nit: since this function is a noop when using a 4626 adapter, perhaps we want to rename it to `trySetScale`  instead? we could optionally log as well.

also, instead of passing in `is4626`, maybe we may want to check for `"setScale" in adapter` instead, which might be a little less error prone & is a bit simpler.

pkg/deployments/deploy/simulated/04-adapters.js
```diff
@@ -151,17 +170,19 @@ module.exports = async function () {
     return await ethers.getContract(underlyingName, signer);
   }
 
-  async function deployTarget(targetName, tDecimals, underlyingAddress) {
+  async function deployTarget(targetName, tDecimals, underlyingAddress, is4626) {
```
abylaw: nit: I slightly prefer having two functions here:
`deployTarget(targetName, tDecimals, underlyingAddress)` & `deploy4626Target(targetName, underlyingAddress)` since they don't share that much code.

pkg/deployments/deploy/simulated/04-adapters.js
```diff
@@ -76,24 +76,41 @@ module.exports = async function () {
   }
 
   // Helpers
-  async function deployAdapter(t, factory) {  
+  async function deployAdapter(t, factory, is4626) {
```
abylaw: nit: same comment as below, it may be easier to read to split into two functions, `deployAdapter` & `deploy4626Adapter`.

pkg/core/src/tests/test-helpers/mocks/MockAdapter.sol
```diff
@@ -5,18 +5,21 @@ import { ERC20 } from "@rari-capital/solmate/src/tokens/ERC20.sol";
...
+import { ERC4626CropAdapter } from "../../../adapters/ERC4626CropAdapter.sol";
+import { ERC4626CropsAdapter } from "../../../adapters/ERC4626CropsAdapter.sol";
 import { FixedMath } from "../../../external/FixedMath.sol";
 import { Divider } from "../../../Divider.sol";
 import { YT } from "../../../tokens/YT.sol";
 import { MockTarget } from "./MockTarget.sol";
 import { MockToken } from "./MockToken.sol";
+import { ERC4626 } from "@rari-capital/solmate/src/mixins/ERC4626.sol";
 
+// Mock crop adapter
```
abylaw, line 16: nit: delete comment

pkg/core/src/tests/test-helpers/TestHelper.sol
```diff
@@ -96,26 +128,33 @@ contract TestHelper is DSTest {
         inputs[1] = "_forge_mock_target_decimals";
         mockTargetDecimals = uint8(abi.decode(hevm.ffi(inputs), (uint256)));
 
+        inputs[1] = "_forge_mock_4626_target";
+        is4626 = uint8(abi.decode(hevm.ffi(inputs), (uint256))) == 0 ? false : true;
```
abylaw, line 132:
```suggestion
        is4626 = uint8(abi.decode(hevm.ffi(inputs), (uint256))) != 0;
```

abylaw: also for my own education: what in the evm is happening here? 🤪

pkg/core/src/tests/test-helpers/TestHelper.sol
```diff
@@ -96,26 +128,33 @@ contract TestHelper is DSTest {
...
-        }
-        if (mockTargetDecimals != 18) {
-            emit log_named_uint(
-                "Running tests with the mock Target token configured with the following number of decimals",
-                uint256(mockTargetDecimals)
-            );
+        // Log target setup
+        if (is4626) {
+            target = MockTargetLike(deployMockTarget(address(underlying), "Compound Dai", "cDAI", mockTargetDecimals));
+            emit log("Running tests with a 4626 mock target");
```
abylaw, line 142:
```suggestion
            emit log(`Running tests with a 4626 mock target with ${mockTargetDecimals} decimals and mock underlying with ${mockUnderlyingDecimals} decimals`);
```
?

then we can delete the `emit`s after :)

pkg/core/src/tests/test-helpers/TestHelper.sol
```diff
@@ -231,29 +255,55 @@ contract TestHelper is DSTest {
...
+        return _addLiquidityToBalancerVault(address(adapter), maturity, tBal);
+    }
+
+    function _addLiquidityToBalancerVault(
+        address adapter,
+        uint256 maturity,
+        uint256 tBal
+    ) internal {
+        MockTargetLike target = MockTargetLike(MockAdapter(adapter).target());
+        MockToken underlying = MockToken(!is4626 ? target.underlying() : target.asset());
```
abylaw, line 288: nit: move this into if statement since it is only used within the `else`?

pkg/core/src/tests/test-helpers/TestHelper.sol
```diff
@@ -264,16 +314,17 @@ contract TestHelper is DSTest {
             mode: MODE,
             tilt: 0
         });
-        someFactory = new MockFactory(address(divider), factoryParams, _reward); // deploy adapter factory
-        someFactory.addTarget(_target, true);
-        divider.setIsTrusted(address(someFactory), true);
-        periphery.setFactory(address(someFactory), true);
+        if (is4626) {
+            someFactory = address(new Mock4626CropFactory(address(divider), factoryParams, _reward)); // deploy 4626 crop adapter factory
```
abylaw: nit: remove comment

pkg/core/src/tests/test-helpers/TestHelper.sol
```diff
@@ -264,16 +314,17 @@ contract TestHelper is DSTest {
...
+        } else {
+            someFactory = address(new MockFactory(address(divider), factoryParams, _reward)); // deploy crop adapter factory
```
abylaw: nit: remove comment (i think code is self-explanatory)

pkg/core/src/tests/test-helpers/TestHelper.sol
```diff
@@ -284,43 +335,68 @@ contract TestHelper is DSTest {
             mode: MODE,
             tilt: 0
         });
-        someFactory = new MockCropsFactory(address(divider), factoryParams, _rewardTokens); // deploy adapter factory
-        someFactory.addTarget(_target, true);
-        divider.setIsTrusted(address(someFactory), true);
-        periphery.setFactory(address(someFactory), true);
+        if (is4626) {
+            someFactory = address(new Mock4626CropsFactory(address(divider), factoryParams, _rewardTokens)); // deploy 4626 crops adapter factory
```
abylaw: nit: remove comment

pkg/core/src/tests/test-helpers/TestHelper.sol
```diff
@@ -284,43 +335,68 @@ contract TestHelper is DSTest {
...
+        } else {
+            someFactory = address(new MockCropsFactory(address(divider), factoryParams, _rewardTokens)); // deploy crops adapter factory
```
abylaw: nit: remove comment

pkg/core/src/tests/test-helpers/mocks/MockAdapter.sol
```diff
@@ -255,6 +312,57 @@ contract MockCropsAdapter is CropsAdapter {
...
+    function doCombine(uint256 maturity, uint256 uBal) external returns (uint256 tBal) {
+        tBal = Divider(divider).combine(address(this), maturity, uBal);
+    }
+
+    function doRedeemPrincipal(uint256 maturity, uint256 uBal) external {
+        Divider(divider).redeem(address(this), maturity, uBal);
+    }
+}
+
+// Mock base adapter
```
abylaw, line 396: nit: remove comment

pkg/core/src/tests/test-helpers/mocks/MockAdapter.sol
```diff
@@ -255,6 +312,57 @@ contract MockCropsAdapter is CropsAdapter {
     }
 }
 
+// Mock ERC4626 crops adapter
```
abylaw, line 338: nit: remove comment

pkg/core/src/tests/test-helpers/mocks/MockAdapter.sol
```diff
@@ -132,12 +135,66 @@ contract MockAdapter is CropAdapter {
...
+    function doCombine(uint256 maturity, uint256 uBal) external returns (uint256 tBal) {
+        tBal = Divider(divider).combine(address(this), maturity, uBal);
+    }
+
+    function doRedeemPrincipal(uint256 maturity, uint256 uBal) external {
+        Divider(divider).redeem(address(this), maturity, uBal);
+    }
+}
+
+// Mock crops adapter
```
abylaw, line 207: nit: remove comment

pkg/core/src/tests/test-helpers/mocks/MockAdapter.sol
```diff
@@ -132,12 +135,66 @@ contract MockAdapter is CropAdapter {
     }
 }
 
+// Mock ERC4626 crop adapter
```
abylaw, line 145: nit: remove comment

pkg/core/src/tests/Periphery.t.sol
```diff
@@ -141,17 +142,23 @@ contract PeripheryTest is TestHelper {
             mode: MODE,
             tilt: 0
         });
-        MockFactory cropFactory = new MockFactory(address(divider), factoryParams, address(reward)); // deploy adapter factory
-        divider.setIsTrusted(address(cropFactory), true);
-        periphery.setFactory(address(cropFactory), true);
+
+        address cropFactory;
+        if (is4626) {
+            cropFactory = address(new Mock4626CropFactory(address(divider), factoryParams, address(reward))); // deploy 4626 crop adapter factory
```
abylaw: nit: comments

pkg/core/src/tests/Periphery.t.sol
```diff
@@ -1137,12 +1079,14 @@ contract PeripheryTest is TestHelper {
         assertEq(ptBal, 1e18);
     }
 
+    event Haha(uint256);
```
abylaw: nit: 🤣

pkg/core/src/tests/adapters/ERC4626Adapter.t.sol
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    function testCantWrapMoreThanBalance() public {
+        uint256 wrapAmt = INITIAL_BALANCE + 1;
+
+        // Approvals
+        target.approve(address(erc4626Adapter), type(uint256).max);
+        underlying.approve(address(erc4626Adapter), type(uint256).max);
+
+        hevm.expectRevert("TRANSFER_FROM_FAILED");
```
abylaw, line 127: is it possible to expect a generic revert? since we may not care too much about the specific error that caused the wrap/unwrap to fail?

pkg/core/src/tests/adapters/ERC4626Adapter.t.sol
```diff
@@ -0,0 +1,193 @@
...
+        wrapAmt = bound(wrapAmt, 1, INITIAL_BALANCE);
+
+        // Approvals
+        target.approve(address(erc4626Adapter), type(uint256).max);
+        underlying.approve(address(erc4626Adapter), type(uint256).max);
+
+        // 1. Run a full wrap -> unwrap cycle
+        uint256 prebal = underlying.balanceOf(address(this));
+        uint256 targetFromWrap = erc4626Adapter.wrapUnderlying(wrapAmt);
+        assertEq(targetFromWrap, target.balanceOf(address(this)));
```
abylaw, line 72: we may want to add a few sanity checks here (`targetFromWrap` > 0), otherwise `wrapUnderlying` & `unwrapTarget` could be noops & still pass the `assertEq(prebal, postbal)` below :)

ditto for part 4 of this test

pkg/core/src/tests/adapters/ERC4626Adapter.t.sol
```diff
@@ -0,0 +1,193 @@
...
+            minm: MIN_MATURITY,
+            maxm: MAX_MATURITY,
+            mode: MODE,
+            tilt: 0,
+            level: Constants.DEFAULT_LEVEL
+        });
+
+        erc4626Adapter = new ERC4626Adapter(address(divider), address(target), ISSUANCE_FEE, adapterParams);
+    }
+
```
abylaw, line 61: do we want to add a test for edge-case of unwrapping/wrapping 0 target/underlying? what is the expected behaviour there?

pkg/core/src/adapters/ERC4626CropAdapter.sol
fedealconada: in case someone wants to create an adapter in which the target passed is > 18 decimals, this subtraction will revert

pkg/core/src/tests/adapters/CropAdapter.t.sol
jparklev: A question is: what are we trying to test? Are we running the tests with two adapter types to test the components themselves (divider, periphery, etc), or is it to test the adapters. For the latter, I think I prefer to test have comprehensive adapter tests that we feel cover our bases, for the former I guess i'd just want us to be intentional about what we're testing

pkg/core/src/tests/Periphery.t.sol
fedealconada: lol

pkg/core/src/tests/Periphery.t.sol
fedealconada: what do you mean? :)

pkg/core/src/tests/adapters/ERC4626Adapter.t.sol
fedealconada: i don;t think there's a generic one but we kind of care about the error message (at least on this test). I would expect there that it reverts because of that specific error and not some other different one :)

pkg/core/src/tests/Periphery.t.sol
abylaw: ah, i think the comments can be removed since the code is fairly self-explanatory :)

fedealconada: gotcha!

pkg/deployments/deploy/simulated/04-adapters.js
fedealconada: i've changed this :) @abylaw

pkg/core/src/tests/adapters/ERC4626Adapter.t.sol
fedealconada: added!

pkg/core/src/tests/adapters/ERC4626Adapter.t.sol
fedealconada: added!

pkg/core/src/tests/adapters/CropAdapter.t.sol
fedealconada: we've decided we're gonna keep the `is4626` at least for now and the idea running all the test suite with both a customised (non-erc4626) adapter and an erc4626 adapter.

pkg/core/src/tests/adapters/ERC4626Adapter.t.sol
abylaw: Ahh, ok :) Mind clarifying a bit for me? I think my confusion stems from the fact that it's not immediately obvious to me why trying to wrap more than balance would lead to a transfer failure.

abylaw (approved):
LGTM! Thanks for the changes!!

pkg/core/src/tests/adapters/ERC4626Adapter.t.sol
fedealconada: >  why trying to wrap more than balance would lead to a transfer failure.

this is because the adapter is first trying to make a `transferFrom` of the user's underlying balance into the adapter ([here](https://github.com/sense-finance/sense-v1/pull/250/files/ecd0348d36737488e87263f41546c8fb1c62e275#diff-4dc0b2e7ea69b908bba676964700f1033425d4a69a28d6b4ec6104e8ebeb1947R54)). As the user does not have that balance that the adapter is trying to pull from his "account", it triggers a `TRANSFER_FROM_FAILED`

abylaw: Ahhhh, ok, that makes sense to me now :)

fedealconada:
approved @jparklev ?

jparklev (approved):
approved

PR #253, opened 2022-05-30
Crop adapter refactor

fedealconada:
## Motivation

- Avoid code duplication (seen in `ERC4626Adapter.sol`, `ERC4626CropAdapter.sol` and `ERC4626CropsAdapter.sol`.
 - Have a better separation of concerns.

## Solution

Decouple Crop and Crops behaviour from respective `CropAdapter.sol` and `CropsAdapter.sol` into separate contracts: `Crop.sol` and `Crops.sol`. Remove existing `CropsAdapter.sol`, `CropAdapter.sol`, `ERC4626CropsAdapter.sol` and `ERC4626CropAdapter.sol` and modify existing contract which are inherting from those mentioned contracts to inherit from `Crop.sol` and `Crops.sol`

## Implementer Checklist


pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -1,90 +1,25 @@
...
-        return;
-    }
-
-    /* ========== MODIFIERS ========== */
-
-    modifier onlyDivider() {
-        if (divider != msg.sender) revert Errors.OnlyDivider();
-        _;
+    ) public override(BaseAdapter, Crop) {
+        Crop.notify(_usr, amt, join);
```
jparklev:
```suggestion
        super.notify(_usr, amt, join);
```

(same comment in a few places)

pkg/core/src/adapters/ERC4626CropAdapter.sol
```diff
@@ -2,60 +2,25 @@
 pragma solidity 0.8.11;
```
jparklev: Probably we can put this and the other 4626 adapter in a `4626` folder?

pkg/core/src/adapters/ERC4626CropAdapter.sol
```diff
@@ -2,60 +2,25 @@
...
+import { Crop } from "./crops/Crop.sol";
 
 /// @notice Adapter contract for ERC4626 Vaults
-contract ERC4626CropAdapter is CropAdapter {
-    using SafeTransferLib for ERC20;
-
-    uint256 public immutable BASE_UINT;
-    uint256 public immutable SCALE_FACTOR;
-
+contract ERC4626CropAdapter is ERC4626Adapter, Crop {
```
jparklev: Might be good to put a little note on the top like "4626 adapter with support for single token airdrops"

pkg/core/src/adapters/CropAdapter.sol
```diff
@@ -1,90 +1,25 @@
...
-    uint256 public share; // accumulated reward token per collected target
-    uint256 public rewardBal; // last recorded balance of reward token
-    uint256 public totalTarget; // total target accumulated by all users
-    mapping(address => uint256) public tBalance; // target balance per user
-    mapping(address => uint256) public rewarded; // reward token per user
-
-    event Distributed(address indexed usr, address indexed token, uint256 amount);
+import { Crop } from "./crops/Crop.sol";
 
+abstract contract CropAdapter is BaseAdapter, Crop {
```
jparklev: With `Crop.sol` now, I wonder if we need the `CropAdapter` – for specific like the `CAdapter`, we could inherit from `Crop` directly. 

Also maybe we could make an `extensions` folder and put `Crop` in there, so it's like theres a base adapter and then a list of extensions you could add to your adapter for more functionality.

Your thoughts?

jparklev (commented):
Really like this directionally, a quick comment about a general concept of `extensions`

pkg/core/src/adapters/CropAdapter.sol
fedealconada: this is just the same, right? i simply decided to use `Crop.` to have it more "clear" to the user that it is the `notify()` from `Crop.sol` the one that is being used and not the one in `BaseAdapter.sol`

pkg/core/src/adapters/CropAdapter.sol
fedealconada: definitely! we cabn remove `Crop(s)Adapter.sol`, the only "downside" (though I think I'm still in favour) is that each crop adapter contract (e.g `CAdapter.sol`) would have to override the `notify()` and will _always_ be the same code:
```
function notify(
        address _usr,
        uint256 amt,
        bool join
    ) public override(BaseAdapter, Crop) {
        super.notify(_usr, amt, join);
    }
    
```
    
re. `extensions` folder, i'm totally down, had kind of the same thought but didn't re-organised the folders because wanted to check if you were down with this refactor (that's also why i put the PR as draft), gonna push these changes for you to review :)

pkg/core/src/adapters/abstract/extensions/Crops.sol
```diff
@@ -32,23 +31,16 @@ abstract contract CropsAdapter is Trust, BaseAdapter {
...
 
-    constructor(
-        address _divider,
-        address _target,
-        address _underlying,
-        uint128 _ifee,
-        BaseAdapter.AdapterParams memory _adapterParams,
-        address[] memory _rewardTokens
-    ) Trust(msg.sender) BaseAdapter(_divider, _target, _underlying, _ifee, _adapterParams) {
+    constructor(address _divider, address[] memory _rewardTokens) Trust(msg.sender) {
```
abylaw, line 34: `Trust(divider)`?

pkg/core/src/adapters/abstract/extensions/Crop.sol
```diff
@@ -6,12 +6,12 @@ import { ERC20 } from "@rari-capital/solmate/src/tokens/ERC20.sol";
...
-import { Divider } from "../Divider.sol";
-import { BaseAdapter } from "./BaseAdapter.sol";
-import { FixedMath } from "../external/FixedMath.sol";
+import { Divider } from "../../../Divider.sol";
+import { FixedMath } from "../../../external/FixedMath.sol";
+import { Trust } from "@sense-finance/v1-utils/src/Trust.sol";
 import { Errors } from "@sense-finance/v1-utils/src/libs/Errors.sol";
 
-abstract contract CropAdapter is BaseAdapter {
+abstract contract Crop is Trust {
```
abylaw, line 14: Nice~ big fan of separating out Crops from Adapter

abylaw (approved):
Also like this reorganization :) +1

pkg/core/src/adapters/abstract/extensions/Crops.sol
fedealconada: this is a good question :)
here, we do the Trust(msg.sender) because as this is Crop**s**, we have a method at the bottom that is the `setRewardTokens` which the owner of this contract should be able to call to update the rewards tokens.
As we also need the divider to be trusted, on the line below we are doing `setIsTrusted(_divider, true);` so trust the divider.

We could also put them the other way round `Trust(_divider)` and then `setIsTrusted(msg.sender, true);`, it's the same

fedealconada:
approved @jparklev ?

jparklev (approved):

PR #255, opened 2022-06-13
GEN-862: use expectRevert

fedealconada:
## Motivation

Have tests up to date with foundry's features

## Solution

Remove all `try/catch` statements and replace them for `expectRevert()`


jparklev (approved):
Looks great. Should we do the same for `CFactory.tm` and `PoolManager.tm` as well?

fedealconada:
> Looks great. Should we do the same for `CFactory.tm` and `PoolManager.tm` as well?

good catch! i missed those, just added them :)

abylaw (approved):
Great, thank you!

PR #256, opened 2022-06-13
chore: replace User.sol for use of prank()

fedealconada:
## Motivation

Have tests up to date with foundry's features

## Solution

Remove `User.sol` (which was a contract used to impersonate accounts) and replace it for the use of `prank()` or `startPrank()`/`stopPrank()`.

## Implementer Checklist


pkg/core/src/tests/YT.t.sol
```diff
@@ -66,20 +72,22 @@ contract Yield is TestHelper {
...
-        uint256 aytBalanceAfter = ERC20(yt).balanceOf(address(alice));
-        uint256 bytBalanceAfter = ERC20(yt).balanceOf(address(bob));
-        uint256 tBalanceAfter = target.balanceOf(address(bob));
+        uint256 lscale = divider.lscales(address(adapter), maturity, bob);
+        uint256 aytBalanceBefore = ERC20(yt).balanceOf(alice);
+        uint256 bytBalanceBefore = ERC20(yt).balanceOf(bob);
+        uint256 tBalanceBefore = target.balanceOf(bob);
+        hevm.prank(bob);
+        ERC20(yt).approve(alice, type(uint256).max);
+        Token(yt).transferFrom(bob, alice, bytBalanceBefore);
```
jparklev, line 87: It looks like we approve the `alice` address but initiate the `transferFrom` from the test contract address?

pkg/core/src/tests/test-helpers/TestHelper.sol
```diff
@@ -226,43 +227,42 @@ contract TestHelper is DSTest {
...
-        updateUser(user, target, stake, amt);
+        alice = address(this); // alice is the default user
+        initUser(alice, target, MAX_TARGET);
+        initUser(bob, target, MAX_TARGET);
+        initUser(jim, target, MAX_TARGET);
     }
 
-    function updateUser(
-        User user,
+    function initUser(
```
jparklev, line 239: Could [`label` the addresses](https://github.com/foundry-rs/forge-std/blob/78adc23f035d9ed4f6fedda5cccf80b1d4a9b9a6/src/Vm.sol#L91) as well so it's easier to tell which address is which in the trace

pkg/core/src/tests/test-helpers/mocks/MockAdapter.sol
```diff
@@ -137,9 +137,9 @@ contract MockAdapter is BaseAdapter, Crop {
         tBal = Divider(divider).combine(address(this), maturity, uBal);
     }
 
-    function doRedeemPrincipal(uint256 maturity, uint256 uBal) external {
-        Divider(divider).redeem(address(this), maturity, uBal);
-    }
+    // function doRedeemPrincipal(uint256 maturity, uint256 uBal) external {
```
jparklev: Good to delete these?

jparklev (approved):
✨ quality work ✨

LGTM. How does it feel to you to do it this way? Also are you still running into the issue with `prank` not always working that you'd mentioned to me before?

pkg/core/src/tests/YT.t.sol
fedealconada: yes, this is because in the `setUp()` i'm setting alice as the default address `alice = address(this)`

fedealconada:
> LGTM. How does it feel to you to do it this way? 

it feels good but i have to say I kind of like what we had to be honest, like calling things like `alice.doIssue` it was very clear who was performing the action but it might be that I just got used to that. And I think we are better following foundry's features rather than having the User.sol workaround.

> Also are you still running into the issue with prank not always working that you'd mentioned to me before?

No! It was actually that I was not seeing that I was doing another call haha. As it was like "inline" i was not seeing it! Something like:
```
hevm.prank(alice);
divider.issue(address(adapter), maturity, ERC20(target).balanceOf(alice));
```
(note that `balanceOf` call)

pkg/core/src/tests/test-helpers/TestHelper.sol
fedealconada: added!

jparklev:
> And I think we are better following foundry's features rather than having the User.sol workaround.

Yep, feel similarly and ultimately agree

> No! It was actually that I was not seeing that I was doing another call haha

Ahh classic, well glad you caught it. This PR LGTM now :)

pkg/core/src/tests/adapters/CropsAdapter.t.sol
```diff
@@ -39,9 +40,10 @@ contract CropsAdapters is TestHelper {
         cropsAdapter = MockCropsAdapter(f);
         divider.setGuard(address(cropsAdapter), 10 * 2**128);
 
-        updateUser(alice, aTarget, stake, MAX_TARGET);
-        updateUser(bob, aTarget, stake, MAX_TARGET);
-        updateUser(jim, aTarget, stake, MAX_TARGET);
+        alice = alice;
```
abylaw: nit: what's this for? :)

pkg/core/src/tests/test-helpers/mocks/MockAdapter.sol
```diff
@@ -423,7 +419,7 @@ contract MockBaseAdapter is BaseAdapter {
         return 1e18;
     }
 
-    function doSetAdapter(Divider d, address _adapter) public {
-        d.setAdapter(_adapter, true);
-    }
+    // function doSetAdapter(Divider d, address _adapter) public {
+    //     d.setAdapter(_adapter, true);
+    // }
```
abylaw: delete these too?

abylaw (approved):
thanks for doing this :)

> And I think we are better following foundry's features rather than having the User.sol workaround.

+1

pkg/core/src/tests/adapters/CropsAdapter.t.sol
fedealconada: removing! :)

pkg/core/src/tests/test-helpers/mocks/MockAdapter.sol
fedealconada: yes! ty!

PR #257, opened 2022-06-18
GEN-942: reconcile

fedealconada:
## Motivation

Through our Bug Bounty programme, we've been made aware of a bug affecting our Crop/sAdapter contracts. Basically, YT holders could disrupt the balance for rewards kept by `CropAdapter` and `CropsAdapter` if they were holding their YTs past maturity leading to a potential dilution of next Series YT holders rewards.

## Solution

Implemented a `reconcile` function which can be called by anyone and, ideally, should be called right after maturity which reconciles the `tBalance` (target balance of each user) for a given maturity.  

**To reviewers: better to start reviewing `CropAdater.sol` first and its tests (`CropAdapter.t.sol`) since `CropsAdapter` is almost the same.**

## Implementer Checklist

- [x]  Get the PR reviewed by at least two people


pkg/core/src/adapters/abstract/extensions/Crop.sol
```diff
@@ -40,15 +45,58 @@ abstract contract Crop is Trust {
...
+                        amt = 0;
+                    } else {
+                        unchecked {
+                            amt -= uReconciledAmt;
+                        }
+                        uReconciledAmt = 0;
+                    }
+                    reconciledAmt[_usr] = uReconciledAmt;
+                }
+                if (amt > 0) {
```
jparklev, line 63: We could move the decrements below into the `else` above since `amt` will always be greater than zero in that case, right?

jparklev (approved):
👍

pkg/core/src/adapters/abstract/extensions/Crop.sol
fedealconada: hmm... but if we move them to the else above they will only be executed when `uReconciledAmt > 0` and we want the decrements also when  `uReconciledAmt == 0` @jparklev

pkg/core/src/adapters/abstract/extensions/Crop.sol
```diff
@@ -22,13 +23,17 @@ abstract contract Crop is Trust {
...
+    mapping(address => uint256) public reconciledAmt; // reconciled target amount per user
+    mapping(address => mapping(uint256 => bool)) public reconciled; // whether a user has been reconciled for a given maturity
 
     event Distributed(address indexed usr, address indexed token, uint256 amount);
 
     constructor(address _divider, address _reward) Trust(_divider) {
         reward = _reward;
     }
 
+    /// @notice Distribute the rewards tokens to the user according to they share
```
abylaw:
```suggestion
    /// @notice Distribute the rewards tokens to the user according to their share
```

abylaw (approved):
Yay!! 🎉 🎉

PR #259, opened 2022-07-13
GEN-1148: add 4626 factory

fedealconada:
## Motivation

To be able to permissionlessly deploy ERC-4626 adapters

## Solution

`ERC4626Factory.sol` and `ERC4626CropsFactory.sol` have  been created. This factories allows anyone to call `deployAdapter()` which will deploy either an `ERC4626Adapter.sol` (for non crop adapters) or a `ERC4626CropsAdapter.sol` (for crop(s) adapters).

### Claimer
Crop(s) contracts need to override the `_claimRewards` function in case that target's protocol does not airdrop rewards tokens (airdrop = they transfer the rewards tokens to the user) but the holder address has to manually claim them by calling some function defined on their contracts (for example this is what happens on the [CAdapter.sol](https://github.com/sense-finance/sense-v1/blob/dev/pkg/core/src/adapters/implementations/compound/CAdapter.sol#L127)).

On one hand, Crop(s) contracts now have a `address claimer` which implements the `IClaimer.sol` interface which exposes a `claim()` function that knows how to claim rewards. In cases of manual claiming, a trusted address can call `setClaimer` passing the address of a claimer contract which would be a contract that knows how to claim rewards on the protocol.  On the other hand, the `_claimRewards`would do the following:
- Transfer all the adapter's target balance to the Claimer contract
- Call `claimer.claim()` which will transfer reward tokens to the Claimer contract
- Get target back from Claimer to the adapter
- Transfer reward tokens from the Claimer to the adapter
        
### Oracles
As opposed to other existing factories where we pass the oracle address and, then, all deployed adapters use that factory we can not do this for this factory.
Instead, we are doing the following:

- Created a `MasterPriceOracle.sol` which, inspired by [Rari's Master Oracle](https://github.com/Rari-Capital/fuse-contracts/blob/master/contracts/oracles/MasterPriceOracle.sol), it's a contract that Sense will manage and it basically returns prices from different oracles. 
  - If there's an oracle address specified for the given underlying, it calls that oracle to return the price. If not, it tries to retrieve the price from Sense Chainlink's Oracle. Ultimately, if there's no price available it **reverts**. 
   - Trusted addresses can add new oracle addresses for underlyings by using the `add()` function.
- Created a `ChainlinkPriceFeed.sol` which, inspired by [Rari's Chainlink Price Oracle](https://github.com/Rari-Capital/fuse-v1/blob/development/src/oracles/ChainlinkPriceOracleV3.sol), it's a contract that Sense will manage and it basically returns prices `UNDERLYING-ETH` from available Chainlink oracles.
- We've decided not to use Rari's master oracle since we don't want to rely 100% on a third party and, having our own, gives us more flexibility.
- In the case of `mUSD` and `mBTC`, we will be using [Rari's mStable Price Feed](https://github.com/Rari-Capital/fuse-v1/blob/development/src/oracles/MStablePriceOracle.sol) which will be added to the Sense Master Oracle when deploying it.

If we want to add support for an underlying price that is not available in Chainlink, we can call Sense `MasterPriceOracle.sol` `add([underlying], [oracle])` function with the address of the underlying and the address of an oracle which must implement the `IPriceFeed.sol` interface.    

### Mainnet tests
- There are mainnet tests using `imUSD` (mStable's ERC4626 vault whose underlying is `mUSD`).

## Implementer Checklist


pkg/core/src/adapters/abstract/erc4626/ERC4626Adapter.sol
```diff
@@ -0,0 +1,161 @@
...
+            int256 answer,
+            uint256 startedAt,
+            uint256 updatedAt,
+            uint80 answeredInRound
+        );
+}
+
+/// @notice Adapter contract for ERC4626 Vaults
+contract ERC4626Adapter is
+    BaseAdapter // TODO:, Trust {
```
fedealconada: If I add `Trust` here `ERC4626CropsAdapter` complains because:

`contract ERC4626CropsAdapter is ERC4626Adapter, Crops {`

both `ERC4626Adapter` and `Crops` inherit`Trust`.

pkg/core/src/adapters/abstract/erc4626/ERC4626Adapter.sol
```diff
@@ -0,0 +1,161 @@
...
+// Internal references
+import { IPriceFeed } from "../../implementations/oracles/IPriceFeed.sol";
+import { FixedMath } from "../../../external/FixedMath.sol";
+import { BaseAdapter } from "../BaseAdapter.sol";
+
+interface ChainlinkOracleLike {
+    /// @notice get data about the latest round. Consumers are encouraged to check
+    /// that they're receiving fresh data by inspecting the updatedAt and
+    /// answeredInRound return values.
+    /// Note that different underlying implementations of AggregatorV3Interface
```
jparklev: Hmm is this a problem for us? Ie some oracles return data in a slightly different format?

pkg/core/src/adapters/abstract/erc4626/ERC4626Adapter.sol
```diff
@@ -0,0 +1,161 @@
...
+                // Sanity checks
+                if (block.timestamp - updatedAt > 1 hours) revert Errors.InvalidPrice();
+                if (block.timestamp - ethUpdatedAt > 1 hours) revert Errors.InvalidPrice();
+
+                // Calculate Underlying-ETH price
+                price = uint256(answer).fdiv(uint256(ethPrice));
+            } catch {
+                // If not found, we check if there's a custom oracle or revert otherwise
+                if (adapterParams.oracle == address(0)) revert Errors.PriceNotFound();
+                (price, ) = IPriceFeed(adapterParams.oracle).price();
```
jparklev: Should this be checked first? (if `adapterParams.oracle` exists, use that, otherwise use chainlink?)

pkg/core/src/adapters/implementations/oracles/IPriceFeed.sol
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
jparklev: Hmm feels a little odd to me to have this in the `/implementations` folder. What do you think about putting it in `/abstract`?

jparklev (commented):
Oop, started reviewing then saw it was still in draft. Lemme know when you want me to take a look

pkg/core/src/adapters/abstract/erc4626/ERC4626Adapter.sol
fedealconada: I left this comment there actually because of that, I'm checking that out and will let you know!

pkg/core/src/adapters/abstract/erc4626/ERC4626Adapter.sol
fedealconada: Well, I actually thought about ALWAYS defaulting to Chainlink. But I guess it makes more sense to do it that way.

pkg/core/src/adapters/implementations/oracles/IPriceFeed.sol
fedealconada: Completely!

pkg/core/src/adapters/implementations/oracles/ChainlinkPriceOracle.sol
```diff
@@ -0,0 +1,130 @@
...
+            uint80 answeredInRound
+        );
+
+    function decimals(address base, address quote) external view returns (uint8);
+}
+
+/// @title ChainlinkPriceOracle
+/// @notice Returns prices from Chainlink.
+/// @dev Implements `IPricefeed` and `Trust`.
+/// @author Inspired on: https://github.com/Rari-Capital/fuse-contracts/blob/master/contracts/oracles/.sol
```
jparklev: Is this a direct copy or did we modify things?

pkg/core/foundry.toml
```diff
@@ -1,16 +1,16 @@
-[default]
+[profile.default]
 root = "."
 libs = ['../../node_modules/', '../../']
 tests = "./src/tests"
 ffi = true
 verbosity = 3
-# Fork block mined on July 15 2022 at 02:09:53 PM UTC
-fork_block_number = 15147715
+# Fork block mined on July 12 2022 at 04:20:08 PM UTC
```
jparklev: What's the reason to change from `July 15`?

pkg/core/src/adapters/implementations/oracles/ChainlinkPriceOracle.sol
fedealconada: It's not an exact copy. Their contract is [here](https://github.com/Rari-Capital/fuse-contracts/blob/master/contracts/oracles/ChainlinkPriceOracleV3.sol) for you to compare.
I've added:
- `maxSecondsBeforePriceIsStale` which I took the idea from an older version of their oracle contract (seems like they have removed it on the last version)
- I'm using the `updatedAt` param returned from Chainlink's calls while they are using the `answeredInRound` and `roundId` but I was told this on Chainlink's discord:

image

pkg/core/foundry.toml
fedealconada: I think I was testing something, I've just removed this change

pkg/core/src/adapters/abstract/erc4626/ERC4626Adapter.sol
```diff
@@ -5,24 +5,21 @@ pragma solidity 0.8.11;
...
+import { MasterPriceOracle } from "../../implementations/oracles/MasterPriceOracle.sol";
+import { FixedMath } from "../../../external/FixedMath.sol";
+import { BaseAdapter } from "../BaseAdapter.sol";
 
 /// @notice Adapter contract for ERC4626 Vaults
 contract ERC4626Adapter is BaseAdapter {
     using SafeTransferLib for ERC20;
     using FixedMath for uint256;
 
+    address public constant RARI_MASTER_ORACLE = 0x1887118E49e0F4A78Bd71B792a49dE03504A764D; // Rari's Master Oracle
```
abylaw: super nit (variable is self-document):
```suggestion
    address public constant RARI_MASTER_ORACLE = 0x1887118E49e0F4A78Bd71B792a49dE03504A764D;
```

pkg/core/src/adapters/implementations/oracles/ChainlinkPriceOracle.sol
```diff
@@ -0,0 +1,139 @@
...
+
+/// @title ChainlinkPriceOracle
+/// @notice Returns prices from Chainlink.
+/// @dev Implements `IPricefeed` and `Trust`.
+/// @author Inspired on: https://github.com/Rari-Capital/fuse-contracts/blob/master/contracts/oracles/MStablePriceOracle.sol
+contract ChainlinkPriceOracle is IPriceFeed, Trust {
+    using FixedMath for uint256;
+
+    // Chainlink's denominations
+    address public constant ETH = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
```
abylaw, line 32: lolll

pkg/utils/src/libs/Errors.sol
```diff
@@ -19,6 +19,9 @@ library Errors {
     error SenderNotEligible();
     error TargetMismatch();
     error TargetNotSupported();
+    error InvalidAdapterType();
+    error PriceOracleNotFound();
+    error ValidOracleExists();
```
abylaw: Not sure I see this error being used :) 

The naming is also a bit confusing to me because it's not immediately clear why having a valid oracle would be an error haha

pkg/core/src/tests/test-helpers/Hevm.sol
```diff
@@ -2,64 +2,149 @@
 pragma solidity 0.8.11;
 
 abstract contract Hevm {
-    // Sets the block timestamp to x
-    function warp(uint256 x) public virtual;
+    // This allows us to getRecordedLogs()
```
abylaw, line 5: for my own education: do we get this file from `foundry`? :)

pkg/core/src/tests/oracles/MasterPriceOracle.t.sol
```diff
@@ -0,0 +1,115 @@
...
+        hevm.expectRevert("UNTRUSTED");
+        address[] memory data;
+        oracle.add(data, data);
+    }
+
+    function testAddOracle() public {
+        assertEq(oracle.oracles(address(underlying)), address(0));
+
+        // add oracle for underlying
+        address[] memory underlyings = new address[](1);
```
abylaw, line 95: is it possible to init arrays in a single statement in solidity?
some syntax i googled:
```
address[] memory underlyings[1] = [address(underlying)];
```

pkg/core/src/tests/oracles/MasterPriceOracle.t.sol
```diff
@@ -0,0 +1,115 @@
...
+
+contract MasterPriceOracleTest is MasterPriceOracleTestHelper {
+    function testMainnetDeployOracle() public {
+        address[] memory underlyings = new address[](1);
+        underlyings[0] = address(underlying);
+        address[] memory oracles = new address[](1);
+        oracles[0] = address(mockOracle);
+
+        MasterPriceOracle otherOracle = new MasterPriceOracle(address(chainlinkOracle), underlyings, oracles);
+        assertTrue(address(otherOracle) != address(0));
```
abylaw, line 51: for my own education: is this check to make sure we've deployed the oracle contract successfully?

pkg/core/src/tests/oracles/MasterPriceOracle.t.sol
```diff
@@ -0,0 +1,115 @@
...
+        // Mock call to Chainlink's oracle
+        uint256 price = 123e18;
+        bytes memory data = abi.encode(price); // return data
+        hevm.mockCall(
+            address(chainlinkOracle),
+            abi.encodeWithSelector(chainlinkOracle.price.selector, address(underlying)),
+            data
+        );
+        assertEq(oracle.price(address(underlying)), price);
+    }
```
abylaw, line 82: as mentioned in another comment, should we add a test for mutliple/custom oracles?

pkg/core/src/tests/oracles/ChainlinkPiceOracle.tm.sol
```diff
@@ -0,0 +1,76 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
abylaw, line 1: Filename: `ChainlinkPiceOracle` -> `ChainlinkPriceOracle` 🤪 
(also applies to `ChainlinkPiceOracle.t.sol`)

abylaw: also for my own education, what's the diff between the `.tm.sol` vs `.t.sol`? :)

pkg/core/src/tests/oracles/ChainlinkPiceOracle.t.sol
```diff
@@ -0,0 +1,95 @@
...
+        oracle.price(address(underlying));
+    }
+
+    function testCantSetMaxSecondsBeforePriceIsStaleIfNotTrusted() public {
+        hevm.prank(address(123));
+        hevm.expectRevert("UNTRUSTED");
+        oracle.setMaxSecondsBeforePriceIsStale(12345);
+    }
+
+    function testSetMaxSecondsBeforePriceIsStaleIfNotTrusted() public {
```
abylaw, line 198:
```suggestion
    function testSetMaxSecondsBeforePriceIsStaleIfTrusted() public {
```

pkg/core/src/tests/factories/ERC4626Factory.tm.sol
```diff
@@ -0,0 +1,186 @@
...
+        address[] memory rewardTokens;
+        bytes memory data = abi.encode(0, rewardTokens);
+
+        factory.deployAdapter(AddressBook.IMUSD, data);
+    }
+
+    function testMainnetCanSupportTarget() public {
+        assertTrue(!factory.supportedTargets(AddressBook.cDAI));
+        factory.supportTarget(AddressBook.cDAI, true);
+        assertTrue(factory.supportedTargets(AddressBook.cDAI));
```
abylaw, line 201: also would be nice to test `factory.supportTarget(AddressBook.cDAI, false)`

pkg/core/src/tests/factories/ERC4626Factory.t.sol
```diff
@@ -0,0 +1,224 @@
...
+
+        // Deploy ERC4626 Crops factory
+        ERC4626Factory someFactory = ERC4626Factory(deployERC4626Factory(address(someTarget)));
+
+        // Prepare data for crops adapter
+        address[] memory rewardTokens;
+        bytes memory data = abi.encode(1, rewardTokens);
+
+        // Deploy crops adapter
+        ERC4626CropsAdapter adapter = ERC4626CropsAdapter(someFactory.deployAdapter(address(someTarget), data));
```
abylaw, line 177: maybe we want to deploy a second adapter so we can test that the reward tokens have also been set on the other adapter?

pkg/core/src/tests/factories/ERC4626Factory.t.sol
```diff
@@ -0,0 +1,224 @@
...
+        MockToken someUnderlying = new MockToken("Some Underlying", "SU", 18);
+        MockTarget someTarget = new MockTarget(address(someUnderlying), "Some Target", "ST", 18);
+        factory.supportTarget(address(someTarget), true);
+        hevm.expectRevert(abi.encodeWithSelector(Errors.OnlyPeriphery.selector));
+        factory.deployAdapter(address(someTarget), "");
+    }
+
+    function testFailDeployAdapterIfAlreadyExists() public {
+        divider.setPeriphery(alice);
+        factory.deployAdapter(address(target), "");
```
abylaw, line 163: maybe an extra check to see that the adapter has not changed? (i.e. the address has stayed the same before we attempted to deploy the adapter again`?)

pkg/core/src/tests/factories/ERC4626Factory.t.sol
```diff
@@ -0,0 +1,224 @@
...
+    function testCantDeployAdapterIfNotPeriphery() public {
+        MockToken someUnderlying = new MockToken("Some Underlying", "SU", 18);
+        MockTarget someTarget = new MockTarget(address(someUnderlying), "Some Target", "ST", 18);
+        factory.supportTarget(address(someTarget), true);
+        hevm.expectRevert(abi.encodeWithSelector(Errors.OnlyPeriphery.selector));
+        factory.deployAdapter(address(someTarget), "");
+    }
+
+    function testFailDeployAdapterIfAlreadyExists() public {
+        divider.setPeriphery(alice);
```
abylaw, line 162: even though i know why we do this, it's still surprising when i read it :p
could we add a comment along the lines of: `// setting ourselves as periphery, since only the periphery can deploy adapters`

pkg/core/src/tests/factories/ERC4626Factory.t.sol
```diff
@@ -0,0 +1,224 @@
...
+        assertTrue(adapter != address(0));
+
+        uint256 scale = MockAdapter(adapter).scale();
+        assertEq(scale, 1e18);
+
+        hevm.warp(block.timestamp + 1 days);
+        uint256 maturity = DateTimeFull.timestampFromDateTime(2021, 10, 1, 0, 0, 0);
+
+        // Sponsor series
+        (address principal, address yield) = periphery.sponsorSeries(adapter, maturity, true);
```
abylaw, line 148: super nit:
```suggestion
        (address pt, address yt) = periphery.sponsorSeries(adapter, maturity, true);
```
for consistency?

pkg/core/src/tests/adapters/ERC4626Adapter.tm.sol
```diff
@@ -0,0 +1,158 @@
...
+            oracle: address(masterOracle),
+            stake: AddressBook.WETH,
+            stakeSize: STAKE_SIZE,
+            minm: MIN_MATURITY,
+            maxm: MAX_MATURITY,
+            mode: 0,
+            tilt: 0,
+            level: DEFAULT_LEVEL
+        });
+        erc4626Adapter = new ERC4626Adapter(address(divider), AddressBook.IMUSD, ISSUANCE_FEE, adapterParams); // imUSD ERC-4626 adapter
```
abylaw, line 87: nit: don't think comment is needed

pkg/core/src/tests/adapters/ERC4626Adapter.t.sol
```diff
@@ -205,7 +223,60 @@ contract ERC4626AdapterTest is DSTestPlus {
...
+
+        // Add oracle to Sense master price oracle
+        masterOracle.add(underlyings, oracles);
+
+        // Assert that the price returned is the one from the custom oracle and not the one Rari
+        assertEq(erc4626Adapter.getUnderlyingPrice(), customOracle.price(address(underlying)));
+    }
+
+    // function testCantGetUnderlyingPriceIfNoFeed() public {
+    // }
```
abylaw, line 296: nit: delete?

pkg/core/src/tests/adapters/ERC4626Adapter.t.sol
```diff
@@ -205,7 +223,60 @@ contract ERC4626AdapterTest is DSTestPlus {
         assertEq(erc4626Adapter.scaleStored(), scale);
     }
 
-    function testGetUnderlyingPrice() public {
-        assertEq(erc4626Adapter.getUnderlyingPrice(), MockOracle(mockOracle).price(address(underlying)));
+    function testGetUnderlyingPriceUsingSenseChainlinkOracle() public {
+        // Mock call to Rari's master oracle
+        uint256 price = 123e18;
+        bytes memory data = abi.encode(price); // return data
```
abylaw, line 230: nit:
```suggestion
        bytes memory returnData = abi.encode(price);
```

pkg/core/src/tests/adapters/ERC4626Adapter.t.sol
```diff
@@ -205,7 +223,60 @@ contract ERC4626AdapterTest is DSTestPlus {
...
+        hevm.mockCall(
+            address(chainlinkOracle),
+            abi.encodeWithSelector(chainlinkOracle.price.selector, address(underlying)),
+            data
+        );
+
+        assertEq(erc4626Adapter.getUnderlyingPrice(), price);
```
abylaw, line 237: I feel like this might not be the right place for testing `MasterPriceOracle`'s fallback behaviour, since the adapter should be agnostic to how the oracle is implemented.

Maybe we should move this test (and the two below) to `MasterPriceOracle.t.sol`, since that is where it is implemented?

pkg/core/src/tests/oracles/MasterPriceOracle.t.sol
```diff
@@ -0,0 +1,115 @@
...
+    function testCantSetNewChainlinkOracle() public {
+        hevm.prank(address(123));
+        hevm.expectRevert("UNTRUSTED");
+        oracle.setSenseChainlinkPriceOracle(address(111));
+    }
+
+    function testSetNewChainlinkOracle() public {
+        assertEq(oracle.senseChainlinkPriceOracle(address(chainlinkOracle)));
+        oracle.setSenseChainlinkPriceOracle(address(123));
+        assertEq(oracle.senseChainlinkPriceOracle(), address(123));
```
abylaw, line 121: quick check that the event is being emitted?

pkg/core/src/tests/oracles/MasterPriceOracle.t.sol
```diff
@@ -0,0 +1,115 @@
...
+        assertEq(oracle.oracles(address(underlying)), address(0));
+
+        // add oracle for underlying
+        address[] memory underlyings = new address[](1);
+        underlyings[0] = address(underlying);
+        address[] memory oracles = new address[](1);
+        oracles[0] = address(oracle);
+
+        oracle.add(underlyings, oracles);
+        assertEq(oracle.oracles(address(underlying)), address(oracle));
```
abylaw, line 101: maybe add a test that existing oracles are untouched if we only set one oracle?

pkg/core/src/adapters/implementations/oracles/ChainlinkPriceOracle.sol
```diff
@@ -0,0 +1,139 @@
...
+    FeedRegistryLike public feedRegistry = FeedRegistryLike(0x47Fb2585D2C56Fe188D0E6ec628a38b74fCeeeDf); // Chainlink feed registry contract
+
+    constructor(uint256 _maxSecondsBeforePriceIsStale) public Trust(msg.sender) {
+        maxSecondsBeforePriceIsStale = _maxSecondsBeforePriceIsStale;
+    }
+
+    /// @dev Internal function returning the price in ETH of `underlying`.
+    function _price(address underlying) internal view returns (uint256) {
+        // Return 1e18 for WETH
+        if (underlying == 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2) return 1e18;
```
abylaw, line 48: pull address into variable? or is that more expensive in gas? :)

pkg/core/src/adapters/implementations/oracles/ChainlinkPriceOracle.sol
```diff
@@ -0,0 +1,139 @@
...
+
+        // Try token/ETH to get token/ETH
+        try feedRegistry.latestRoundData(underlying, ETH) returns (
+            uint80,
+            int256 tokenEthPrice,
+            uint256,
+            uint256 updatedAt,
+            uint80
+        ) {
+            if (tokenEthPrice <= 0) return 0;
```
abylaw, line 58: for my own education, when can this happen?

pkg/core/src/tests/oracles/ChainlinkPiceOracle.t.sol
```diff
@@ -0,0 +1,95 @@
...
+        // Mock call to Chainlink's oracle
+        uint256 price = 123e18;
+        bytes memory data = abi.encode(1, int256(price), block.timestamp, block.timestamp, 1); // return data
+        hevm.mockCall(
+            address(feedRegistry),
+            abi.encodeWithSelector(feedRegistry.latestRoundData.selector, address(underlying), oracle.ETH()),
+            data
+        );
+        assertEq(oracle.price(address(underlying)), price);
+    }
```
abylaw, line 64: tests if token/eth pair doesn't exist?

token/usd & token/btc?

pkg/core/src/adapters/implementations/oracles/ChainlinkPriceOracle.sol
```diff
@@ -0,0 +1,139 @@
...
+            if (tokenUsdPrice <= 0) return 0;
+            _validatePrice(updatedAt);
+
+            int256 ethUsdPrice;
+            (, ethUsdPrice, , updatedAt, ) = feedRegistry.latestRoundData(ETH, USD);
+            if (ethUsdPrice <= 0) return 0;
+            _validatePrice(updatedAt);
+
+            return
+                uint256(tokenUsdPrice).fmul(1e26).fdiv(10**uint256(feedRegistry.decimals(underlying, USD))).fdiv(
```
abylaw, line 82: maybe a quick comment here about `1e26` or the number of dp for `ethUsdPrice`?

abylaw (commented):
niiiiice!! left some comments mainly around test completeness & nits :) 
thanks as usual for all your hard work!

pkg/core/src/adapters/implementations/oracles/ChainlinkPriceOracle.sol
jparklev: Update: @fedealconada & I talked about favoring a global stale `maxSecondsBeforePriceIsStale` for the chainlink oracle which will primarily be to confirm that an oracle hasn't been deprecated. Regardless, this comment is resolved and I think "inspired by" makes sense

pkg/utils/src/libs/Errors.sol
fedealconada: removing it!

pkg/core/src/tests/test-helpers/Hevm.sol
fedealconada: From [here](https://book.getfoundry.sh/cheatcodes/) :)

pkg/core/src/tests/oracles/MasterPriceOracle.t.sol
fedealconada: You can initialise fixed-sized arrays as you say though the syntax would be:
`address[1] memory underlyings = [address(123)];`

The problem here is that we are passing that array to the `add` function (`masterOracle.add(underlyings, oracles);`) which expects a dynamic array :(

pkg/core/src/adapters/implementations/oracles/ChainlinkPriceOracle.sol
fedealconada: I feel like it would change the gas cost, just didn't add it because that's thee only place it's being used. But I can add it.

pkg/core/src/adapters/implementations/oracles/ChainlinkPriceOracle.sol
fedealconada: Should never happen but it's a safety check, maybe if Chainlink's decides to deprecate an oracle and, for example, they decide to set the price to 0

pkg/core/src/adapters/implementations/oracles/ChainlinkPriceOracle.sol
abylaw: Ah no need, I think the comment works just fine!

pkg/core/src/adapters/implementations/oracles/ChainlinkPriceOracle.sol
abylaw: Ahhh, got it! Thanks!

pkg/core/src/tests/oracles/MasterPriceOracle.t.sol
abylaw: ahh! makes sense
didn't realize the two types of arrays were distinct in solidity, good to know :)

fedealconada:
@jparklev @abylaw I pushed a new commit which adds the `Claimer` idea. Check the **Claimer** section of the PR description. Also, note that `ERC4626Factory.sol` has now been split and we have `ERC4626Factory.sol` for non-crop 4626 adapters and `ERC4626CropsFactory.sol` for crop(s) 4626 adapters.

pkg/core/src/tests/oracles/ChainlinkPiceOracle.t.sol
```diff
@@ -58,6 +63,109 @@ contract ChainlinkPriceOracleTest is ChainPriceOracleTestHelper {
...
+
+    function testPriceAttemptFailed() public {
+        // Deploy mocked Feed Registry
+        MockFeedRegistry feedRegistry = new MockFeedRegistry();
+
+        // Deploy mocked Chainlink price oracle
+        MockChainlinkPriceOracle oracle = new MockChainlinkPriceOracle(feedRegistry);
+
+        // Mock calls to decimals()
+        hevm.mockCall(address(feedRegistry), abi.encodeWithSelector(feedRegistry.decimals.selector), abi.encode(18));
```
jparklev, line 160: Nice cheatcode!

pkg/core/src/adapters/implementations/oracles/ChainlinkPriceOracle.sol
jparklev: Also interesting they're using `int` instead of `uint` – do yo know why that is?

jparklev:
Screen Shot 2022-07-25 at 1 49 39 PM
Using the new coverage feature? 👀 Also nice job using the new cheatcodes ~

Re the claimer: the idea is to have it ready for when we need it, but not deploy it yet, right? I'm still a tiny bit hesitant to commit to a model before we see what the market is doing for out-of-standard airdrops (though I understand the approach here is quite flexible)

pkg/core/src/adapters/abstract/extensions/Crops.sol
```diff
@@ -128,9 +130,31 @@ abstract contract Crops is Trust {
...
+                // Get the rewards
+                for (uint256 i = 0; i < rewardTokens.length; i++) {
+                    uint256 rBal = ERC20(rewardTokens[i]).balanceOf(claimer);
+                    ERC20(rewardTokens[i]).transferFrom(claimer, address(this), rBal);
+                }
+            }
+        }
     }
 
+    event Hola(uint256);
```
jparklev: como estas?

pkg/core/src/adapters/implementations/oracles/ChainlinkPriceOracle.sol
jparklev: yea, interesting they're using int instead of uint. Should we ask their team if a price is ever expected to go negative @fedealconada?

pkg/core/src/adapters/implementations/oracles/ChainlinkPriceOracle.sol
```diff
@@ -0,0 +1,138 @@
...
+
+        // Try token/USD to get token/ETH
+        try feedRegistry.latestRoundData(underlying, USD) returns (
+            uint80,
+            int256 tokenUsdPrice,
+            uint256,
+            uint256 updatedAt,
+            uint80
+        ) {
+            if (tokenUsdPrice <= 0) return 0;
```
jparklev, line 74: Is this what Rari does? I wonder if there's any risk of setting it to 0 rather than reverting? I know fuse just reverts on 0 (ie you can't borrow or lend the asset in that case), but not sure if other protocols do the same

pkg/core/src/adapters/abstract/extensions/Crop.sol
```diff
@@ -119,10 +121,36 @@ abstract contract Crop is Trust {
...
+
+            if (tBal > 0) {
+                // We send all the target balance to the claimer contract to it can claim rewards
+                ERC20(target).transfer(claimer, tBal);
+
+                // Make claimer to claim rewards
+                IClaimer(claimer).claim();
+
+                // Get the target back
+                ERC20(target).transferFrom(claimer, address(this), tBal);
```
jparklev: Another option, rather than assume the claimer has given approvals to the adapter, is to assume the claimer will send the Target and reward tokens back to the adapter itself. WDYT @fedealconada?

pkg/core/src/tests/test-helpers/mocks/MockClaimer.sol
```diff
@@ -0,0 +1,30 @@
...
+import { ERC20 } from "@rari-capital/solmate/src/tokens/ERC20.sol";
+
+// Internal references
+import { ERC4626Adapter } from "../../../adapters/abstract/erc4626/ERC4626Adapter.sol";
+import { MockToken } from "./MockToken.sol";
+import { MockTarget } from "./MockTarget.sol";
+import { IClaimer } from "../../../adapters/abstract/IClaimer.sol";
+import { Trust } from "@sense-finance/v1-utils/src/Trust.sol";
+
+contract MockClaimer is IClaimer {
```
jparklev, line 14: Looks like we could deploy a new claimer for each adapter – I suppose we could also add like an authd `addAdapter` function so that claimers could be re-used, but I don't feel strongly about it and I support keeping it as simple as it can be overall

jparklev (approved):
A few comments on the claimer mechanisms, and looks like CI needs a fix, but overall LGTM

pkg/core/src/adapters/implementations/oracles/ChainlinkPriceOracle.sol
fedealconada: just asked, let's see what they answer (they are a bit slow to asnwer)

fedealconada:
> Screen Shot 2022-07-25 at 1 49 39 PM
> 
> Using the new coverage feature? 👀 Also nice job using the new cheatcodes ~

I just meant that I've added a couple more tests, but not measuring it. ~Let me try adding it though :)~ `forge coverage` is giving me probelsm when running it (seems like people are having the same [issue](https://github.com/foundry-rs/foundry/issues/2162)).

> Re the claimer: the idea is to have it ready for when we need it, but not deploy it yet, right? I'm still a tiny bit hesitant to commit to a model before we see what the market is doing for out-of-standard airdrops (though I understand the approach here is quite flexible)

correct! that's what we discussed last time in Paris. I will submit the PR for the deployment scripts today which will be: deploying both oracles (master and chainlink) + 4626 NON crop factory. There will be no support for crop(s) (unless you we want to to deploy that factory but only to support airdrop ones and we just don't make use of the claimer).

pkg/core/src/adapters/abstract/extensions/Crops.sol
fedealconada: todo bienn!

pkg/core/src/adapters/implementations/oracles/ChainlinkPriceOracle.sol
fedealconada: This is what the Rari oracles do (they can return 0). Fuse would revert if they get the price from the oracle an it is 0. So, we should think what's the best approach for us: just reverting at an oracle level or let the oracle return 0 and revert at the adapter level (on the `getUnderlyingPrice()` function) WDYT?

pkg/core/src/adapters/abstract/extensions/Crop.sol
fedealconada: Both options could work, not feeling strong on any of them. In the one you mention we don't have to do the approval when deploying the Claimer (for the adapter to pull the target) but instead we have to add a check on this function to guarantee the adapter is getting the target back.

pkg/core/src/tests/test-helpers/mocks/MockClaimer.sol
fedealconada: Yeah, totally. I think that there are multiple ways in which we could implement claimers so they are more flexible and also re-usable across adapters. But, maybe we can leave this up to when we actually decide to start using 4626 crops.

jparklev: yes, sgtm

pkg/core/src/adapters/implementations/oracles/ChainlinkPriceOracle.sol
jparklev: hmm I think it'd be ok to do it on the level of `getUnderlyingPrice`, since it's a little unexpected for an oracle itself to revert (my feeling atm at least)?

fedealconada: i completely agree

pkg/core/src/adapters/abstract/extensions/Crop.sol
jparklev: Yep, I think I have a slight preference for that (the check to guarantee it got the target back) if it makes sense to you!

abylaw: one argument for doing a check each time might mean it's easier for 3rd parties to integrate with us, since they won't need to remember to do approvals if they deploy an adapter themselves. 

makes it easier for us too, since we don't have to remember to grant approvals either.

pkg/core/src/tests/oracles/ChainlinkPiceOracle.t.sol
```diff
@@ -0,0 +1,203 @@
...
+        // Mock call to BTC/USD Chainlink's oracle
+        uint256 btcEthPrice = 789e18;
+        data = abi.encode(1, int256(btcEthPrice), block.timestamp, block.timestamp, 1); // return data
+        hevm.mockCall(
+            address(feedRegistry),
+            abi.encodeWithSelector(feedRegistry.latestRoundData.selector, oracle.BTC(), oracle.ETH()),
+            data
+        );
+        assertEq(oracle.price(address(underlying)), underBtcPrice.fmul(btcEthPrice).fdiv(1e18));
+    }
```
abylaw, line 131: niiice 👍

pkg/core/src/tests/test-helpers/mocks/MockClaimer.sol
abylaw: for my own education: are contract deploys expensive, gas-wise? (by what factor x does it cost more than calling a function)?

abylaw (approved):

pkg/core/src/adapters/abstract/extensions/Crop.sol
fedealconada: sgtm! we can do the same fo the reward tokens, instead of doing a `transferFrom` from the adapter, we could just make the Claimer to transfer them to the adapter.

pkg/core/src/tests/test-helpers/mocks/MockClaimer.sol
fedealconada: yes, deploying is usually far more expensive than calling functions. For example, [this](https://etherscan.io/tx/0x345bfd26441a0a9a8d9b0002231d280b32467a948bbe1e6fac9adabc7ea21232) is the tx that deployed our Divider and had a cost of `0.107` while calling `issue` had a cost of `0.009` ([here](https://etherscan.io/tx/0xfbb97d48f90f4fb9150375c1e81f4a53b38efb7c4472f854238aa4737edcc15d)) and calling `redeem` had a cost of `0.001` ([here](https://etherscan.io/tx/0x1efc172fefbb0beac729799e9a35ae77ec558fbb99f4954875617eb4e98accc9))

pkg/core/src/adapters/abstract/extensions/Crop.sol
fedealconada: added this

pkg/core/src/adapters/implementations/oracles/ChainlinkPriceOracle.sol
fedealconada: pushed

pkg/utils/src/libs/Errors.sol
```diff
@@ -21,6 +21,7 @@ library Errors {
     error TargetNotSupported();
     error InvalidAdapterType();
     error PriceOracleNotFound();
+    error UnexpectedTargetAmount();
```
jparklev: `BadContractInteration` or `BadClaimerInteration`

As an option – something I saw used in locke protocol

jparklev (approved):
Reviewed the latest commit and ✅ from me (pending word from kenton on the `setGuard` stuff we talked about)

pkg/core/src/tests/test-helpers/mocks/MockClaimer.sol
abylaw: got it, thank you!

pkg/utils/src/libs/Errors.sol
fedealconada: Gone with first one, potentially more reusable for other contracts :)

fedealconada:
> Reviewed the latest commit and ✅ from me (pending word from kenton on the `setGuard` stuff we talked about)

I've added this [commit](e5586c8108a479a5fa941dbf554bebee71811468) which will test that the guard is initialised. This test will now failed until we merge the [PR](https://github.com/sense-finance/sense-v1/pull/264) that has the setGuard which can already review hihi (cc @jparklev @abylaw)

jparklev:
> > Reviewed the latest commit and ✅ from me (pending word from kenton on the `setGuard` stuff we talked about)
> 
> I've added this [commit](e5586c8108a479a5fa941dbf554bebee71811468) which will test that the guard is initialised. This test will now failed until we merge the [PR](https://github.com/sense-finance/sense-v1/pull/264) that has the setGuard which can already review hihi (cc @jparklev @abylaw)

Perfect (read the change and thought imUSD had 36 decimals for a sec hehe)

PR #260, opened 2022-07-20
GEN-1233: wsteth with steth as underlying

fedealconada:
## Motivation

We want to remove the stETH peg exposure out from our adapter.

## Solution

Remove ETH as the underlying of this adapter and use stETH instead.

## Implementer Checklist


jparklev (approved):
Carried over from the private repo – this & the new task LGTM

abylaw (approved):

PR #262, opened 2022-07-26
GEN-1233: add deployment output

fedealconada:
## Motivation

-

## Solution

Adding deployment output + removing `underlying` argument from wsteth adapter.

## Implementer Checklist


pkg/deployments/tasks/20220720-wsteth-adapter/index.js
```diff
@@ -80,7 +81,7 @@ task(
     // verify adapter contract on etherscan
     if (chainId !== "111") {
       console.log("\n-------------------------------------------------------");
-      await verifyOnEtherscan(adapterAddress, [divider, target.address, await target.underlying(), ifee, adapterParams, reward]);
+      await verifyOnEtherscan("0x29872d0a2672bA73D540c53330E1A073c3cE2673", [divider.address, target.address, ifee, adapterParams]);
```
jparklev: What's the reason to hardcode this?

jparklev (approved):
LGTM!

pkg/deployments/tasks/20220720-wsteth-adapter/index.js
fedealconada: good catch, i was just testing and forgot to put it back :)

PR #264, opened 2022-07-28
feat(Factories): set guard on deployAdapter()

fedealconada:
## Motivation

When deploying adapters via factories, we need to automatically set the guard. Otherwise, these adapters won't be able to issue until `guarded` is set to `false` or until admins set a guard.

## Solution

On `deployAdapter()`, make a call to set the guard. 

## Implementer Checklist

### DIFF
--- a/pkg/core/src/adapters/abstract/factories/BaseFactory.sol
+++ b/pkg/core/src/adapters/abstract/factories/BaseFactory.sol
@@ -78,16 +78,13 @@ abstract contract BaseFactory {
 
         // Get Underlying-ETH price
         try BaseAdapter(adapter).getUnderlyingPrice() returns (uint256 underlyingPriceInEth) {
-            // Get ETH-USD price from Chainlink
+            // Get ETH-USD price from Chainlink (in 8 decimals base)
             (, int256 ethPrice, , uint256 ethUpdatedAt, ) = ChainlinkOracleLike(ETH_USD_PRICEFEED).latestRoundData();
 
-            if (block.timestamp - ethUpdatedAt > 1 hours) revert Errors.InvalidPrice();
+            if (block.timestamp - ethUpdatedAt > 1.5 hours) revert Errors.InvalidPrice();
 
             // Calculate Underlying-USD price (normalised to 18 deicmals)
-            ChainlinkOracleLike(ETH_USD_PRICEFEED).decimals();
-            uint256 price = underlyingPriceInEth.fmul(
-                uint256(ethPrice) * 10**(18 - ChainlinkOracleLike(ETH_USD_PRICEFEED).decimals())
-            );
+            uint256 price = underlyingPriceInEth.fmul(uint256(ethPrice) * 1e10);
 
             // Calculate Target-USD price
             price = BaseAdapter(adapter).scale().fdiv(price);
--- a/pkg/core/src/adapters/abstract/factories/BaseFactory.sol
+++ b/pkg/core/src/adapters/abstract/factories/BaseFactory.sol
@@ -75,8 +75,6 @@ abstract contract BaseFactory {
     function _setGuard(address adapter) internal {
         // We only want to execute this if divider is guarded
         if (Divider(divider).guarded()) {
-            uint256 guard = factoryParams.guard;
-
             // Get Underlying-ETH price
             try BaseAdapter(adapter).getUnderlyingPrice() returns (uint256 underlyingPriceInEth) {
                 // Get ETH-USD price from Chainlink (in 8 decimals base)
@@ -90,10 +88,9 @@ abstract contract BaseFactory {
 
                 // Calculate Target-USD price
                 price = BaseAdapter(adapter).scale().fdiv(price);
-                guard = guard.fdiv(price);
-            } catch {}
 
-            Divider(divider).setGuard(adapter, guard);
+                Divider(divider).setGuard(adapter, factoryParams.guard.fdiv(price));
+            } catch {}
         }
     }
 
--- a/pkg/core/src/adapters/abstract/factories/BaseFactory.sol
+++ b/pkg/core/src/adapters/abstract/factories/BaseFactory.sol
@@ -81,7 +81,7 @@ abstract contract BaseFactory {
             // Get ETH-USD price from Chainlink (in 8 decimals base)
             (, int256 ethPrice, , uint256 ethUpdatedAt, ) = ChainlinkOracleLike(ETH_USD_PRICEFEED).latestRoundData();
 
-            if (block.timestamp - ethUpdatedAt > 1.5 hours) revert Errors.InvalidPrice();
+            if (block.timestamp - ethUpdatedAt > 2 hours) revert Errors.InvalidPrice();
 
             // Calculate Underlying-USD price (normalised to 18 deicmals)
             uint256 price = underlyingPriceInEth.fmul(uint256(ethPrice) * 1e10);
--- a/pkg/core/src/adapters/abstract/factories/BaseFactory.sol
+++ b/pkg/core/src/adapters/abstract/factories/BaseFactory.sol
@@ -6,7 +6,6 @@ import { Errors } from "@sense-finance/v1-utils/src/libs/Errors.sol";
 import { BaseAdapter } from "../BaseAdapter.sol";
 import { Divider } from "../../../Divider.sol";
 import { FixedMath } from "../../../external/FixedMath.sol";
-import { Divider } from "../../../Divider.sol";
 
 interface ChainlinkOracleLike {
     function latestRoundData()
@@ -74,24 +73,28 @@ abstract contract BaseFactory {
     /// Set adapter's guard to $100`000 in target
     /// @notice if Underlying-ETH price feed returns 0, we set the guard to 100000 target.
     function _setGuard(address adapter) internal {
-        uint256 guard = factoryParams.guard;
+        // We only want to execute this if divider is guarded
+        if (Divider(divider).guarded()) {
+            uint256 guard = factoryParams.guard;
 
-        // Get Underlying-ETH price
-        try BaseAdapter(adapter).getUnderlyingPrice() returns (uint256 underlyingPriceInEth) {
-            // Get ETH-USD price from Chainlink (in 8 decimals base)
-            (, int256 ethPrice, , uint256 ethUpdatedAt, ) = ChainlinkOracleLike(ETH_USD_PRICEFEED).latestRoundData();
+            // Get Underlying-ETH price
+            try BaseAdapter(adapter).getUnderlyingPrice() returns (uint256 underlyingPriceInEth) {
+                // Get ETH-USD price from Chainlink (in 8 decimals base)
+                (, int256 ethPrice, , uint256 ethUpdatedAt, ) = ChainlinkOracleLike(ETH_USD_PRICEFEED)
+                    .latestRoundData();
 
-            if (block.timestamp - ethUpdatedAt > 2 hours) revert Errors.InvalidPrice();
+                if (block.timestamp - ethUpdatedAt > 2 hours) revert Errors.InvalidPrice();
 
-            // Calculate Underlying-USD price (normalised to 18 deicmals)
-            uint256 price = underlyingPriceInEth.fmul(uint256(ethPrice) * 1e10);
+                // Calculate Underlying-USD price (normalised to 18 deicmals)
+                uint256 price = underlyingPriceInEth.fmul(uint256(ethPrice) * 1e10);
 
-            // Calculate Target-USD price
-            price = BaseAdapter(adapter).scale().fdiv(price);
-            guard = guard.fdiv(price);
-        } catch {}
+                // Calculate Target-USD price
+                price = BaseAdapter(adapter).scale().fdiv(price);
+                guard = guard.fdiv(price);
+            } catch {}
 
-        Divider(divider).setGuard(adapter, guard);
+            Divider(divider).setGuard(adapter, guard);
+        }
     }
 
     /* ========== LOGS ========== */
--- a/pkg/core/src/adapters/abstract/factories/BaseFactory.sol
+++ b/pkg/core/src/adapters/abstract/factories/BaseFactory.sol
@@ -56,6 +56,7 @@ abstract contract BaseFactory {
         uint128 ifee; // issuance fee
         uint16 mode; // 0 for monthly, 1 for weekly
         uint64 tilt; // tilt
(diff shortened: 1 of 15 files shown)


pkg/core/src/adapters/abstract/factories/BaseFactory.sol
```diff
@@ -49,6 +70,31 @@ abstract contract BaseFactory {
     /// @param _data Additional data needed to deploy the adapter
     function deployAdapter(address _target, bytes memory _data) external virtual returns (address adapter) {}
 
+    /// Set adapter's guard to $100`000 in target
+    /// @notice if Underlying-ETH price feed returns 0, we set the guard to 100000 target.
+    function _setGuard(address adapter) internal {
+        uint256 guard = uint256(100000 * 1e18);
```
jparklev: Should this be in a constant or a constructor arg?

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
```diff
@@ -49,6 +70,31 @@ abstract contract BaseFactory {
...
+
+        // Get Underlying-ETH price
+        uint256 underlyingEthPrice = BaseAdapter(adapter).getUnderlyingPrice();
+
+        if (underlyingEthPrice > 0) {
+            // Get ETH-USD price from Chainlink
+            (, int256 ethPrice, , uint256 ethUpdatedAt, ) = ChainlinkOracleLike(ETH_USD_PRICEFEED).latestRoundData();
+            if (block.timestamp - ethUpdatedAt > 1 hours) revert Errors.InvalidPrice();
```
jparklev: I'm guessing the eth price feed is updated pretty frequently?

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
```diff
@@ -49,6 +70,31 @@ abstract contract BaseFactory {
     /// @param _data Additional data needed to deploy the adapter
     function deployAdapter(address _target, bytes memory _data) external virtual returns (address adapter) {}
 
+    /// Set adapter's guard to $100`000 in target
+    /// @notice if Underlying-ETH price feed returns 0, we set the guard to 100000 target.
+    function _setGuard(address adapter) internal {
+        uint256 guard = uint256(100000 * 1e18);
+
+        // Get Underlying-ETH price
+        uint256 underlyingEthPrice = BaseAdapter(adapter).getUnderlyingPrice();
```
jparklev:
```suggestion
        uint256 underlyingPriceInEth = BaseAdapter(adapter).getUnderlyingPrice();
```

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
```diff
@@ -49,6 +70,31 @@ abstract contract BaseFactory {
...
+
+        if (underlyingEthPrice > 0) {
```
jparklev: When is this not true? If `getUnderlyingPrice` is not implemented for some adapter and returns `0`?

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
```diff
@@ -49,6 +70,31 @@ abstract contract BaseFactory {
...
+            // Get ETH-USD price from Chainlink
+            (, int256 ethPrice, , uint256 ethUpdatedAt, ) = ChainlinkOracleLike(ETH_USD_PRICEFEED).latestRoundData();
+            if (block.timestamp - ethUpdatedAt > 1 hours) revert Errors.InvalidPrice();
+
+            // Calculate Underlying-USD price (normalised to 18 deicmals)
+            uint256 price = underlyingEthPrice.fmul(uint256(ethPrice)) *
+                10**(18 - ChainlinkOracleLike(ETH_USD_PRICEFEED).decimals());
+
+            // Calculate Target-USD price
+            price = (BaseAdapter(adapter).scale()).fdiv(price);
```
jparklev:
```suggestion
            price = BaseAdapter(adapter).scale().fdiv(price);
```

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
```diff
@@ -49,6 +70,31 @@ abstract contract BaseFactory {
...
+        uint256 underlyingEthPrice = BaseAdapter(adapter).getUnderlyingPrice();
+
+        if (underlyingEthPrice > 0) {
+            // Get ETH-USD price from Chainlink
+            (, int256 ethPrice, , uint256 ethUpdatedAt, ) = ChainlinkOracleLike(ETH_USD_PRICEFEED).latestRoundData();
+            if (block.timestamp - ethUpdatedAt > 1 hours) revert Errors.InvalidPrice();
+
+            // Calculate Underlying-USD price (normalised to 18 deicmals)
+            uint256 price = underlyingEthPrice.fmul(uint256(ethPrice)) *
+                10**(18 - ChainlinkOracleLike(ETH_USD_PRICEFEED).decimals());
```
jparklev: Will the decimals for `ETH_USD_PRICEFEED` change over time? It looks like all the `/ ETH` feeds use 18 decimals, and the `/ USD` ones are mostly `8` decimals (tho I see one 18 decimal feed 🙃 )
Screen Shot 2022-07-28 at 4 46 29 PM
Screen Shot 2022-07-28 at 4 47 50 PM

jparklev (commented):
Nits, a question about the conditional, and a question on the decimals calc. Otherwise looks good

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
fedealconada: I've just read this 
image

fedealconada: (btw, this `1 hours` is the same we have on the wsteth adapter)

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
fedealconada: yeah, actually, now that you ask, we agreed on the 4626 PR that `getUnderlyingPrice()` should revet if price is 0 ([here](https://github.com/sense-finance/sense-v1/pull/259#discussion_r930189053)) and [here](https://github.com/sense-finance/sense-v1/pull/259/files#diff-94b155056ce95a99c4444e5da787acc98576862a622b2da075f29c1df352b3feR48) the change 

So, I should actually do a try catch on line 79 instead of checking the condition...

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
fedealconada: > Will the decimals for ETH_USD_PRICEFEED change over time?
I don't think so but, theoretically, it can change... It's safer to make the call right?

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
jparklev: Ahh ok, should we add a little buffer (like 30 mins or something in case they can't get a tx through), or you feel ok about this guard?

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
jparklev: Hmm I suppose in theory, but I think it'd be quite disruptive for them to change it and it seems pretty unlikely? I'm not opposed to keeping it, just wondering if we can save some gas there

fedealconada: yeah, i guess its highly unlikely, i can remove it

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
fedealconada: so, 1 hour 30 min? sgtm

jparklev: Yea, might even do 2 hours or something to account for extreme conditions (though that's just an guess, maybe i'll check with kenton on this as well to his feeling)

fedealconada: sure, check with him and let me know! we should apply this same logic to the wsteth adapter then (it has 1 hour now). And, if so, we should re-deploy it.

fedealconada: we've decided for the 2 hours

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
```diff
@@ -49,6 +71,29 @@ abstract contract BaseFactory {
...
+
+            // Calculate Underlying-USD price (normalised to 18 deicmals)
+            uint256 price = underlyingPriceInEth.fmul(uint256(ethPrice) * 1e10);
+
+            // Calculate Target-USD price
+            price = BaseAdapter(adapter).scale().fdiv(price);
+            guard = guard.fdiv(price);
+        } catch {}
+
+        Divider(divider).setGuard(adapter, guard);
```
abylaw: It worries me a little bit that now all Factories will be trusted on the `Divider`.

Do we expect users to write their own Factory contracts and extend `BaseFactory`? I might be misunderstanding, but if so, wouldn't we have to call `setTrusted` on each factory, and we would no longer be "permissionless"? It would also be more of a security risk.

fedealconada: This is a good point!
> Do we expect users to write their own Factory contracts and extend BaseFactory? 

Permissionless mode would allow deploying adapters permissionlessly but not factories. It could be the case that a third-paty wants to integrate with us and implements a factory but we will be the ones having to call `setFactory()` to actually set the factory into the Periphery and do the `setIsTrusted` for the Divider. In that case, we just could decide not to mark that factory as trusted fo the divider if we want to be on the safe side. Otherwise, we could "audit" the code and decide that it is safe.
cc @jparklev

jparklev: Hmm right, so the options here are: 
a) don't trust the factory and set the guards manually
b) trust the factories and automatically set guards for people adding adapters

My worry with (b) is that it defeats the purpose of the factories if we have to set the guards ourselves (ie my understanding of the vision of the factories is that we want ppl to be able to deploy and use specific adapters themselves, without our intervention). 

However, it's definitely worth thinking about the security concerns with that option – in theory we'd be giving these factories total authority over the divider and if they're malicious or poorly coded there's a big downside.

The mitigating factor is, as fede mentioned, the fact that we have to be the ones to add factories (even once permisionless mode is enabled), so we'll always review them first. Then once permisionless mode is enabled, anybody can add any adapter they want without the need for factories (in their current form at least).

How do those tradeoffs feel to you @abylaw?

abylaw: Got it - thanks both. Totally understand the need for setting guards automatically, and the mitigating factors make sense to me @fedealconada & @jparklev. 

I understood Factories as a concept that would persist into permissionless mode as a more convenient way to deploy new adapters. I can see it being useful, since protocols could integrate with us and expose a dashboard that allows their users to add assets to Sense, for example (similar to what we're planning to do with 4626...or even more exciting, we could automatically generate a factory page for them??).

Just spitballing now, but perhaps it would be worth considering for V2 having the Divider automatically add guards when an adapter first registers itself? (assuming our architecture remains similar :) )

P.s 
Also, am I right in saying that:
1) once permissionless-mode is enabled,
2) all new adapters deployed will *not* have guards?

fedealconada: > once permissionless-mode is enabled, all new adapters deployed will not have guards?

Technically, we can be on permissionless mode and still have guards, this is controlled by the `guarded` variable on the Divider.
So, once we set `guarded` to `false` we will remove the guards from all adapters.
I'm actually wondering if I should add  a check like:

```
function _setGuard(address adapter) internal {
        if (Divider(divider).guarded()) {
            ....
        }
 }       
```
So once `guarded` is set to `false`, new adapte deployments on existing will skip the the `setGuard()` call. For future factories we deploy (once guarded is false), we will actually remove `_setGuard` at all.

Thoughts @abylaw @jparklev

abylaw: > I'm actually wondering if I should add a check like:

Hmm yea. That's what I initially thought when I first read the code!

Although I can see having guarded adapters being useful in select circumstances, e.g. integrating with a new protocol => our Divider *always* remains in guarded mode, even when most of our other adapters do *not* have a guard.

Wdygt @fedealconada @jparklev

fedealconada: Yeah, it's a possible scenario as well!

fedealconada: I've just pushed this last change, check last commit please cc @jparklev @abylaw :)

abylaw (approved):
lgtm~

just to put my mind to rest: is there any chance we'd turn *on* guarded after turning it off? if so, setting a default guard would be useful even if guarded mode is off

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
```diff
@@ -74,24 +73,28 @@ abstract contract BaseFactory {
...
-            price = BaseAdapter(adapter).scale().fdiv(price);
-            guard = guard.fdiv(price);
-        } catch {}
+                // Calculate Target-USD price
+                price = BaseAdapter(adapter).scale().fdiv(price);
+                guard = guard.fdiv(price);
+            } catch {}
 
-        Divider(divider).setGuard(adapter, guard);
+            Divider(divider).setGuard(adapter, guard);
```
jparklev: Hmm might it make sense to fall back to a guard of `0`, rather than `guard` if we don't know the price since the value could be anything?

jparklev (approved):
Curious about your thoughts on the default guard, but otherwise last commits make sense to me

jparklev:
> lgtm~
> 
> just to put my mind to rest: is there any chance we'd turn _on_ guarded after turning it off? if so, setting a default guard would be useful even if guarded mode is off

Right, off hand this feels pretty unlikely (rather I think we'd `pause` the entire Divider if we needed to limit TVL after setting `guarded` off), but I can talk with kenton on this a bit mon/tues to align.

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
fedealconada: sure, i can set it to 0!

PR #265, opened 2022-07-28
refactor(yt-buying-lib): enable initial space conditions and factor scale in correctly

jparklev:
## Motivation

We got an external request for this library, so I wanted to make sure it was flexible and correct

## Solution

Add the new features from the version of this library in the frontend while maintaining the previous ability to bootstrap the fluxer without a live space pool for data. In addition, add the optimize function into the fluxer directly (rather than assuming the user will bring their own optimization lib)

### DIFF
--- a/yt-buying-lib/index.js
+++ b/yt-buying-lib/index.js
@@ -2,25 +2,31 @@ import Decimal from "decimal.js";
 import optimjs from "optimization-js";
 import SETTINGS from "./settings.js";
 
-class SpaceFluxer {
-  constructor(now, ttl, TS, G2, initScale, ifee) {
-    this.now = now;
-    this.ttl = ttl;
-    this.TS = TS;
-    this.G2 = G2;
-    this.initScale = initScale;
-    this.scale = initScale;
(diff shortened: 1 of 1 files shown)


yt-buying-lib/index.js
```diff
@@ -40,22 +46,28 @@ class SpaceFluxer {
     return targetOut;
   }
 
-  buyYTs(initialTarget, desiredTargetBack = new Decimal("0")) {
+  // Create a function to a YT buy and return the amount of Target remaining at the end
```
alexfertel:
```suggestion
  // Create a function to make/do a YT buy and return the amount of Target remaining at the end
```

alexfertel (approved):
LGTM! Now I remembered that I also refactored a bit on the FE side, so thank you for doing this 🙏🏻

yt-buying-lib/index.js
alexfertel: Didn't know if the correct word was `make` or `do` 😅

fedealconada (approved):
looks good!

PR #266, opened 2022-08-08
GEN-1151: Ribbon adapter

fedealconada:
## Motivation

Have fixed income on [Ribbon Vaults](https://docs.ribbon.finance/).

## Solution

NOTE: THIS PROJECT IS TABLED UNTIL WE ASSIGN IT A HIGHER PRIORITY.
There has been slow progress on this. Below some relevant notes:

Ribbon vaults earn yield on its deposits by running a weekly automated covered call strategy/puts selling strategy. The vault reinvests the yield earned back into the strategy, effectively compounding the yields for depositors over time.
When user deposits some target, this deposit is deployed in the vault’s weekly strategy every Friday at 11am UTC. This means that if user makes a deposit on a Wednesday, those assets will remain "pending" until the cycle starts.

Withdrawals
Once user funds have been used in the vault’s weekly strategy they cannot be withdrawn until the vault closes its position the following Friday at 12pm UTC.
But, users can withdraw their funds instantly during the weekly timelock period where the vault closes its previous position and opens its new one.

Because of Ribbon's dynamics, `wrapUnderlying` and `unwrapTarget` would have to behave differently:

**wrapUnderlying**

For existing adapters at Sense, when wrapping underlying you immediately get target back. In this case, if I make a deposit of underlying on Ribbon, I won't get anything back (immediately) as the next cycle has not yet started. I can only claim the target back once the cycle started. So, we've decided to use a target wrapper so, when a deposit happens, we mint this wrapped target token and we can immediately return it back to the caller. 

**unwrapTarget**

For existing adapters at Sense, when unwrapping target you immediately get your underlying back. In this case, we can't withdraw from Ribbon at any time. This is a 2-step process:
- user initiates a withdrawal by calling [`initiateUnwrapTarget`](https://github.com/sense-finance/sense-v1/compare/dev...GEN-1151-ribbon-adapter#diff-aed4c8fb8fb7f504e715b7f7af7fc1e1e06ca332a7d2fd348310cb8e802ae4f0R172) and the adapter will call `initiateWithdraw` on Ribbon so as to inform them that we don't want to rollover our target to the next cycle. 
- once the cycle ends, we can complete withdrawal by calling `unwrapTarget` (which calls Ribbon's `completeWithdrawal`).

As this dynamic affects the behaviour `unwrapTarget` it will also affect any contract making use of these functions (e.g our Periphery).

Relevant to `initiateUnwrapTarget` there's this particular scenario:

- User A calls wrapUnderlying -> we mint wTarget at current pps (pricePerShare)
- ROUND CLOSES
- User B calls `wrapUnderlying` -> we mint wTarget at current pps
- User B calls `initiateUnwrapTarget` (without waiting for the round to be closed)
  - This call will make the adapter to initiate a withdraw and will make Ribbon not to rollover some target to the next cycle. This is not correct because User's B target is still pending (never got into an active cycle) so we should not be initiating a withdraw on the active deposits.
  - This is why we first need to [check](https://github.com/sense-finance/sense-v1/compare/dev...GEN-1151-ribbon-adapter#diff-aed4c8fb8fb7f504e715b7f7af7fc1e1e06ca332a7d2fd348310cb8e802ae4f0R197) if there's any pending target in Ribbon and we will make the adapter to call `withdrawInstant` which will return underlying and we could send this underlying back to the user (considering that the amount the underlying returned is enough to cover what the user is trying to withdraw. Otherwise, we would just initiate a normal withdraw **(*)**

**(*)** There's also un unresolved issue related to this explained a bit [here](https://github.com/sense-finance/sense-v1/compare/dev...GEN-1151-ribbon-adapter#diff-aed4c8fb8fb7f504e715b7f7af7fc1e1e06ca332a7d2fd348310cb8e802ae4f0R182). This is related to how to convert the wrapped target tokens the user is trying to unwrap to underlying.  

Note: there might still be other non-trivial, unexplored scenarios. 

## Implementer Checklist

### DIFF
--- a/pkg/core/src/adapters/implementations/ribbon/RAdapter.sol
+++ b/pkg/core/src/adapters/implementations/ribbon/RAdapter.sol
@@ -1,0 +1,261 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.11;
+
+// External references
+import { ERC20 } from "@rari-capital/solmate/src/tokens/ERC20.sol";
+import { MockToken } from "../../../tests/test-helpers/mocks/MockToken.sol"; // TODO: replace for MintableToken interface
+import { ERC4626 } from "@rari-capital/solmate/src/mixins/ERC4626.sol";
+import { SafeTransferLib } from "@rari-capital/solmate/src/utils/SafeTransferLib.sol";
+
+// Internal references
+import { Errors } from "@sense-finance/v1-utils/src/libs/Errors.sol";
+import { BaseAdapter } from "../../abstract/BaseAdapter.sol";
+import { FixedMath } from "../../../external/FixedMath.sol";
+
+interface WETHLike {
+    function deposit() external payable;
+
+    function withdraw(uint256 wad) external;
+}
+
+interface RTokenLike {
+    function pricePerShare() external view returns (uint256);
+
+    function roundPricePerShare(uint256) external view returns (uint256);
+
+    function decimals() external view returns (uint8);
+
+    function underlying() external view returns (address);
+
+    function vaultParams()
+        external
+        view
+        returns (
+            bool,
+            uint8,
+            address,
+            address,
+            uint56,
+            uint104
+        );
+
+    function vaultState()
+        external
+        view
+        returns (
+            uint16,
+            uint104,
+            uint104,
+            uint128,
+            uint128
+        );
+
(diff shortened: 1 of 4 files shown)


jparklev:
Archiving this since we've tabled the work here. It's linked in linear so it'll be easy to find again in the future

PR #268, opened 2022-08-10
GEN01334: fix _setGuard

fedealconada:
## Motivation

The new new changes in BaseFactory.sol introduced a bug which was causing the newly deployed adapters using a factory to set a really high cap because it was not considering the target's decimals. 

## Solution

Take target's decimals into account.

## Implementer Checklist

### DIFF
--- a/pkg/core/src/adapters/abstract/factories/BaseFactory.sol
+++ b/pkg/core/src/adapters/abstract/factories/BaseFactory.sol
@@ -7,6 +7,10 @@ import { BaseAdapter } from "../BaseAdapter.sol";
 import { Divider } from "../../../Divider.sol";
 import { FixedMath } from "../../../external/FixedMath.sol";
 
+interface ERC20 {
+    function decimals() external view returns (uint256 decimals);
+}
+
 interface ChainlinkOracleLike {
     function latestRoundData()
         external
@@ -55,7 +59,7 @@ abstract contract BaseFactory {
         uint128 ifee; // issuance fee
         uint16 mode; // 0 for monthly, 1 for weekly
         uint64 tilt; // tilt
-        uint256 guard; // adapter guard (in target)
+        uint256 guard; // adapter guard (in usd, 18 decimals)
     }
 
     constructor(address _divider, FactoryParams memory _factoryParams) {
@@ -72,11 +76,13 @@ abstract contract BaseFactory {
 
     /// Set adapter's guard to $100`000 in target
     /// @notice if Underlying-ETH price feed returns 0, we set the guard to 100000 target.
-    function _setGuard(address adapter) internal {
+    function _setGuard(address _adapter) internal {
         // We only want to execute this if divider is guarded
         if (Divider(divider).guarded()) {
+            BaseAdapter adapter = BaseAdapter(_adapter);
+
             // Get Underlying-ETH price
-            try BaseAdapter(adapter).getUnderlyingPrice() returns (uint256 underlyingPriceInEth) {
+            try adapter.getUnderlyingPrice() returns (uint256 underlyingPriceInEth) {
                 // Get ETH-USD price from Chainlink (in 8 decimals base)
                 (, int256 ethPrice, , uint256 ethUpdatedAt, ) = ChainlinkOracleLike(ETH_USD_PRICEFEED)
                     .latestRoundData();
@@ -86,10 +92,12 @@ abstract contract BaseFactory {
                 // Calculate Underlying-USD price (normalised to 18 deicmals)
                 uint256 price = underlyingPriceInEth.fmul(uint256(ethPrice) * 1e10);
 
-                // Calculate Target-USD price
-                price = BaseAdapter(adapter).scale().fdiv(price);
-
-                Divider(divider).setGuard(adapter, factoryParams.guard.fdiv(price));
+                // Calculate Target-USD price (scale and price are in 18 decimals)
+                price = adapter.scale().fdiv(price);
+                Divider(divider).setGuard(
+                    _adapter,
+                    (factoryParams.guard * 10**ERC20(adapter.target()).decimals()) / price
+                );
             } catch {}
         }
     }
--- a/pkg/core/src/adapters/abstract/factories/BaseFactory.sol
+++ b/pkg/core/src/adapters/abstract/factories/BaseFactory.sol
@@ -42,9 +42,6 @@ abstract contract BaseFactory {
     /// @notice Sense core Divider address
     address public immutable divider;
 
-    /// @notice target -> adapter
-    mapping(address => address) public adapters;
-
     /// @notice params for adapters deployed with this factory
     FactoryParams public factoryParams;
 
--- a/pkg/core/src/adapters/abstract/factories/BaseFactory.sol
+++ b/pkg/core/src/adapters/abstract/factories/BaseFactory.sol
@@ -81,22 +81,25 @@ abstract contract BaseFactory {
         if (Divider(divider).guarded()) {
             BaseAdapter adapter = BaseAdapter(_adapter);
 
-            // Get Underlying-ETH price
+            // Get Underlying-ETH price (18 decimals)
             try adapter.getUnderlyingPrice() returns (uint256 underlyingPriceInEth) {
-                // Get ETH-USD price from Chainlink (in 8 decimals base)
+                // Get ETH-USD price from Chainlink (8 decimals)
                 (, int256 ethPrice, , uint256 ethUpdatedAt, ) = ChainlinkOracleLike(ETH_USD_PRICEFEED)
                     .latestRoundData();
 
                 if (block.timestamp - ethUpdatedAt > 2 hours) revert Errors.InvalidPrice();
 
                 // Calculate Underlying-USD price (normalised to 18 deicmals)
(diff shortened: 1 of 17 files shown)


pkg/core/src/adapters/abstract/factories/BaseFactory.sol
```diff
@@ -86,10 +92,12 @@ abstract contract BaseFactory {
...
 
-                // Calculate Target-USD price
-                price = BaseAdapter(adapter).scale().fdiv(price);
-
-                Divider(divider).setGuard(adapter, factoryParams.guard.fdiv(price));
+                // Calculate Target-USD price (scale and price are in 18 decimals)
+                price = adapter.scale().fdiv(price);
+                Divider(divider).setGuard(
+                    _adapter,
+                    (factoryParams.guard * 10**ERC20(adapter.target()).decimals()) / price
```
jparklev: could we also do `factoryParams.guard.fdiv(price, 10**ERC20(adapter.target()).decimals())`?

abylaw (approved):
LGTM

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
```diff
@@ -86,10 +92,12 @@ abstract contract BaseFactory {
                 // Calculate Underlying-USD price (normalised to 18 deicmals)
                 uint256 price = underlyingPriceInEth.fmul(uint256(ethPrice) * 1e10);
 
-                // Calculate Target-USD price
-                price = BaseAdapter(adapter).scale().fdiv(price);
-
-                Divider(divider).setGuard(adapter, factoryParams.guard.fdiv(price));
+                // Calculate Target-USD price (scale and price are in 18 decimals)
+                price = adapter.scale().fdiv(price);
```
jparklev: Hmm, just noticing should this be multiplied rather than divided?

jparklev (commented):
Quick confirmation on the price calculation

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
abylaw: ah, i also noticed that but thought i might have been mistaken, guess i should trust my intuition more. 

is `getUnderlyingPrice` ETH/underlying & `ethPrice` USD/ETH?

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
fedealconada: yes, it's the same, but looks better hehe, just pushed the change

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
fedealconada: this is a nice find that tests were not catching, i've fixed the tests so they revert if i try with `fdiv` and they pass if they are with `fmul` as they should

fedealconada: cc @jparklev @abylaw

jparklev: sweet, will give them a look. ty fed

pkg/core/src/tests/factories/BaseFactory.t.sol
```diff
@@ -110,52 +104,25 @@ contract Factories is TestHelper {
...
-        assertEq(maxm, MAX_MATURITY);
-        assertEq(MockAdapter(adapter).mode(), MODE);
+        // On this test we assert that the guard calculated is close to the one calculated below
+        // underlying-ETH price (18 decimals)
+        uint256 underlyingPriceInEth = adapter.getUnderlyingPrice();
 
-        uint256 scale = MockAdapter(adapter).scale();
-        assertEq(scale, 1e18);
+        // Calculate Underlying-USD price (normalised to 18 deicmals)
+        uint256 price = underlyingPriceInEth.fmul(Constants.DEFAULT_CHAINLINK_ETH_PRICE, 1e8);
```
jparklev, line 203: To make sure I understand, we mock the oracle call so the internal `_setGuard` will use this same value, right? (if that's right, could we get ride of the `assertClose` and assert exactly?)

pkg/core/src/tests/factories/BaseFactory.t.sol
```diff
@@ -110,52 +104,25 @@ contract Factories is TestHelper {
...
 
-        // Since the call to Chainlink reverted when deploying the adapter via factory
-        // guard should be set to 0
-        (, , , , , , , , uint256 factoryGuard) = someFactory.factoryParams();
-        (, , uint256 guard, ) = divider.adapterMeta(adapter);
-        assertEq(guard, factoryGuard);
+        // Convert DEFAULT_GUARD (which is $100'000 in 18 decimals) to target
+        // using target's price (18 decimals)
+        uint256 guardInTarget = DEFAULT_GUARD.fdiv(price, 10**tDecimals);
+        assertClose(guard, guardInTarget, guard.fmul(0.001e18));
```
jparklev: Since this is for the local test with mocks, I wonder if we can also do an assert with a constant value somewhere as well. That way we could have that sanity check on the guard like "scale is 2 and eth price is 1900, then it makes sense that guard would be 26.xxx" – wdyt?

jparklev (approved):
Updates lgtm @fedealconada! One quick thought on testing, then curious about aby's final thoughts on the changes

pkg/core/src/tests/factories/BaseFactory.t.sol
fedealconada: Correct, I could change it!

abylaw (approved):
lgtm, aside from what jpark has already brought up. :)

pkg/core/src/tests/factories/BaseFactory.t.sol
fedealconada: sgtm! just added it

pkg/core/src/tests/factories/BaseFactory.t.sol
```diff
@@ -113,16 +122,33 @@ contract Factories is TestHelper {
...
-        // Convert DEFAULT_GUARD (which is $100'000 in 18 decimals) to target
+        // (3) Convert DEFAULT_GUARD (which is $100'000 in 18 decimals) to target
         // using target's price (18 decimals)
         uint256 guardInTarget = DEFAULT_GUARD.fdiv(price, 10**tDecimals);
-        assertClose(guard, guardInTarget, guard.fmul(0.001e18));
+        assertEq(guard, guardInTarget);
+
+        // Sanity checks with constants values:
+        // as underlying price is 1e18, ETH price is 1900 (in 8 decimals) and scale is 1e18,
+        // (1) Underlying-USD price should be: 1e18 * (1900 * 1e8) -> normalised to 18 decimals = 1900 * 10**tDecimals
```
jparklev: Know it's annoying but might be good to use a scale other than `1` here to catch the differences b/w `div` & `mul`?

pkg/core/src/tests/factories/BaseFactory.t.sol
```diff
@@ -165,7 +191,7 @@ contract Factories is TestHelper {
         rewardTokens[0] = address(someReward);
         MockCropFactory someFactory = MockCropFactory(deployCropsFactory(address(someTarget), rewardTokens, false));
 
-        address f = periphery.deployAdapter(address(someFactory), address(someTarget), "");
+        address f = periphery.deployAdapter(address(someFactory), address(someTarget), abi.encode(rewardTokens));
```
jparklev, line 275: good catch, we were missing these before?

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
```diff
@@ -42,9 +42,6 @@ abstract contract BaseFactory {
     /// @notice Sense core Divider address
     address public immutable divider;
 
-    /// @notice target -> adapter
-    mapping(address => address) public adapters;
```
jparklev, line 42: Fede mentioned to me that this is currently unused! Previously we used it to ensure a factory didn't deploy the same Target adapter twice, but now we use `CREATE2` with the target address as the salt, so it'll fail anyway (for context for aby)

jparklev (approved):

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
fedealconada: exactly! ty for adding context :)

pkg/core/src/tests/factories/BaseFactory.t.sol
fedealconada: if what we are using is a non 4626 factory, we are find sending an empty string. But when the factory is a crops 4626 factory (as it it in the 4626 tests), `deployAdapter` expects to receive an array of addresses

pkg/core/src/tests/factories/BaseFactory.t.sol
fedealconada: added it!

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
abylaw: tyty @jparklev for the context, good to know :) *interesting*

PR #269, opened 2022-08-18
GEN-863 expect emit [QUALITY WORK]

fedealconada:
## Motivation

Improve code quality

## Solution

Improve it by increasing a bit the test coverage but mainly adding `expectEmit` to asset events are being emitted.  

## Implementer Checklist

### DIFF
--- a/pkg/core/src/adapters/abstract/extensions/Crop.sol
+++ b/pkg/core/src/adapters/abstract/extensions/Crop.sol
@@ -30,7 +30,8 @@ abstract contract Crop is Trust {
 
     event Distributed(address indexed usr, address indexed token, uint256 amount);
 
-    constructor(address _divider, address _reward) Trust(_divider) {
+    constructor(address _divider, address _reward) Trust(msg.sender) {
+        setIsTrusted(_divider, true);
         reward = _reward;
     }
 
--- a/pkg/core/src/tests/Divider.t.sol
+++ b/pkg/core/src/tests/Divider.t.sol
@@ -129,6 +129,10 @@ contract Dividers is TestHelper {
         divider.setAdapter(address(adapter), true);
         hevm.warp(1631664000); // 15-09-21 00:00 UTC
         uint256 maturity = DateTimeFull.timestampFromDateTime(2021, 10, 4, 0, 0, 0); // Monday
+
+        hevm.expectEmit(true, true, true, false);
+        emit SeriesInitialized(address(adapter), maturity, address(0), address(0), address(this), adapter.target());
+
         (address pt, address yt) = periphery.sponsorSeries(address(adapter), maturity, true);
 
         assertTrue(pt != address(0));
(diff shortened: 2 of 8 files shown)


pkg/core/src/adapters/abstract/extensions/Crop.sol
```diff
@@ -30,7 +30,8 @@ abstract contract Crop is Trust {
 
     event Distributed(address indexed usr, address indexed token, uint256 amount);
 
-    constructor(address _divider, address _reward) Trust(_divider) {
+    constructor(address _divider, address _reward) Trust(msg.sender) {
+        setIsTrusted(_divider, true);
```
fedealconada, line 34: This was missing! It affects our current cUSDC and cDAI adapters since we won't be able to neither call `setClaimer` nor `setRewardToken` cc @jparklev

jparklev (approved):
👏

pkg/core/src/adapters/abstract/extensions/Crop.sol
jparklev: Following up here for visibility, fede & I discussed that this doesn't necessitate a re-deploy for the CAdapter, since the `Comp` token is very unlikely to change in the future, and if it did we'd have to update the adapter anyway with the new comptroller. But, this will be carried forward to future crop adapters

pkg/core/src/tests/Divider.t.sol
```diff
@@ -1870,11 +1917,101 @@ contract Dividers is TestHelper {
...
+
+    function testCanSetPaused() public {
+        assertTrue(!divider.paused());
+        divider.setPaused(true);
+        assertTrue(divider.paused());
+    }
+
+    /* ========== LOGS ========== */
+
+    event Backfilled(
```
abylaw, line 1990: there's no way to use the events from the contract directly?

pkg/core/src/tests/adapters/CropAdapter.t.sol
```diff
@@ -1082,6 +1091,107 @@ contract CropAdapters is TestHelper {
...
     }
 
+    // update reward token tests
+
+    function testSetRewardsTokens() public {
+        MockToken reward2 = new MockToken("Reward Token 2", "RT2", 18);
+
+        hevm.prank(address(factory)); // only factory can call `setRewardToken` as it was deployed via cropsFactory
+        // 4626 adapters is crops
+        if (is4626) {
```
abylaw: would it perhaps be better to have two tests? `testSetRewardsTokens4646()` & `testSetRewardsTokens()`
The test that is currently not needed can abort early.

It would be easier to pinpoint the failure w/o checking config

that way it would be clearer which "path" fails.

pkg/core/src/tests/Divider.t.sol
```diff
@@ -488,10 +496,16 @@ contract Dividers is TestHelper {
         uint256 tBase = 10**target.decimals();
         uint256 fee = convertToBase(ISSUANCE_FEE, target.decimals()).fmul(tBal, tBase); // 1 target
         uint256 tBalanceBefore = target.balanceOf(alice);
-        divider.issue(address(adapter), maturity, tBal);
 
         // Formula = newBalance.fmul(scale)
         uint256 mintedAmount = (tBal - fee).fmul(adapter.scale());
+
+        // Expect Issued event
```
abylaw: nit: remove unnecessary comment :b

abylaw (approved):
Nice!! Just two small suggestions.

PR #270, opened 2022-08-18
GEN-1339: add eslint to deploments pkg [QUALITY WORK]

fedealconada:
## Motivation

Improve code quality

## Solution

Improve code quality by adding eslint + prettier to the `/deployments` packcage.
It adds 2 new scripts:
`yarn lint` -> only lints code
`yarn fix` -> lints and tries fixing what's possible

## Implementer Checklist


pkg/deployments/.prettierrc
```diff
@@ -0,0 +1,10 @@
+{
+  "semi": true,
```
jparklev, line 2: What're your thoughts on just taking the defaults for all of these? (except for `printWidth` which I agree with)

jparklev (approved):
so pretty

abylaw:
p r e t t y

we could also lint on commit ~ https://typicode.github.io/husky/#/
for example: https://github.com/sense-finance/web3-yield-curves

pkg/deployments/.prettierrc
fedealconada: let's do it!

fedealconada:
> p r e t t y
> 
> we could also lint on commit ~ https://typicode.github.io/husky/#/ for example: https://github.com/sense-finance/web3-yield-curves

added!

abylaw (approved):

PR #272, opened 2022-08-29
GEN-1396: allow non ERC20 tokens

fedealconada:
## Motivation

Add support to non-ERC20 tokens like USDT.

## Solution

Use `safeApprove()` from `SafeTransferLib` target, underlying and stake approvals.

## Implementer Checklist


pkg/core/src/tests/test-helpers/mocks/MockToken.sol
```diff
@@ -38,3 +38,151 @@ contract AuthdMockToken is ERC20, Trust {
         _burn(account, amount);
     }
 }
+
+// Non-ERC20 token
```
jparklev, line 42: Solmate has some "weird token" examples we could use as well if we wanted? (probably could import directly) What do you think?

https://github.com/transmissions11/solmate/tree/v7/src/test/utils/weird-tokens

.github/workflows/ci.yml
```diff
@@ -68,6 +68,15 @@ jobs:
     - name: Run local core tests where the mock Target is an ERC4626 and decimals are 8
       run: yarn workspace @sense-finance/v1-core run test:4626-8-decimal-val
 
+    - name: Run local core tests where the mock Target is non ERC20 compliant
```
jparklev: Thinking out loud I wonder what the foundry folks would say is the idiomatic way to do this kind of thing?

jparklev (approved):
My understanding is that the only place we're still using normal `approve` is in the wstETH adapter, right? (which would be totally fine since we know wstETH and stETH are compliant)

Comment about utilizing solmate, but LGTM ~

pkg/core/src/tests/test-helpers/TestHelper.sol
```diff
@@ -439,8 +443,43 @@ contract TestHelper is DSTest {
     ) internal returns (address _target) {
         if (is4626) {
             _target = address(new MockERC4626(ERC20(_underlying), _name, _symbol));
+            emit log("Running tests with a 4626 mock target");
+        } else {
+            if (nonERC20Target) {
```
abylaw: else if?

pkg/core/src/tests/test-helpers/TestHelper.sol
```diff
@@ -130,20 +137,17 @@ contract TestHelper is DSTest {
         mockUnderlyingDecimals = uint8(hevm.envUint("FORGE_MOCK_UNDERLYING_DECIMALS"));
         mockTargetDecimals = uint8(hevm.envUint("FORGE_MOCK_TARGET_DECIMALS"));
         is4626 = hevm.envBool("FORGE_MOCK_4626_TARGET");
```
abylaw: For consistency, `is4626Target`?

justfile
```diff
@@ -50,6 +57,24 @@ turbo-test-match-contract *exp="": && _timer
...
+turbo-test-local-non-erc20-underlying *cmds="": && _timer
+	cd {{ invocation_directory() }}; \
+		export FORGE_MOCK_NON_ERC20_UNDERLYING="true"; \
+		forge test --no-match-path "*.tm*" {{ cmds }}
+
+# run tests with 18 decimals and non-ERC20 stake
+turbo-test-local-non-erc20-stake *cmds="": && _timer
+	cd {{ invocation_directory() }}; \
+		export FORGE_MOCK_NON_ERC20_STAKE="true"; \
+		forge test --no-match-path "*.tm*" {{ cmds }}
```
abylaw, line 78: do we want to run some tests with combinations like:
- 4626 target, non-ERC20 underlying
- non-ERC20 target, non-ERC20 underlying

etc...

I think it may be a good option to have a "complete" test suite that we can run only occasionally (before a  release, whenever someone is feeling nervous in dev), and our ci can run an "abbreviated" ("smoke") version per PR

jparklev: This is a cool idea (could be as simple as `dev` run everything vs <any_other_branch>` run abbreviated)

pkg/core/src/tests/test-helpers/mocks/MockToken.sol
fedealconada: Nice tokens haha, I've just had a look at them but none seem to be what we need... or am I wrong?

justfile
fedealconada: This is something I actually talked with @jparklev yesterday, about the combinations and I said that maybe it was too much to have on the CI but I totally agree that we could have the setup so we can run everything once in a while like you said @abylaw :)
Just a small comment: I didn't do any combination between ERC4626 and non-ERC20 target since, as per the 4626 standard, a 4626 token IS an ERC20 and its underlying must be an ERC20 as well.

fedealconada: I've just pushed some changes including this one (check the [justfile](https://github.com/sense-finance/sense-v1/pull/272/files#diff-deb9bb56fb122db0b605aa5b63f95a4665c905b18dd670e1fa6c877576a94ff1R129)) cc @jparklev @abylaw

package.json
```diff
@@ -17,6 +17,7 @@
     "turbo-test-local-non-erc20-target": "yarn workspaces @sense-finance/v1-core run turbo-test:4626",
     "turbo-test-local-non-erc20-underlying": "yarn workspaces @sense-finance/v1-core run turbo-test:4626",
     "turbo-test-local-non-erc20-stake": "yarn workspaces @sense-finance/v1-core run turbo-test:4626",
+    "turbo-test-local-all": "yarn workspaces @sense-finance/v1-core run turbo-test:all",
```
jparklev, line 20: Should this be connected up to a CI step?

justfile
```diff
@@ -123,6 +125,40 @@ turbo-test-mainnet-match *exp="": && _timer
 
 turbo-test-mainnet-match-contract *exp="": && _timer
 	@cd {{ invocation_directory() }}; forge test --fork-url {{ MAINNET_RPC }} --match-contract {{ exp }}
+
+# run for all possible combinations between ERC20 and non-ERC20 target, underlying and stake and 6, 8 and 18 decimals
+# also run, for ERC4626 target, all possible combinations between 6, 8 and 18 target (and underlying) decimals
+turbo-test-local-all *cmds="": && _timer
```
jparklev, line 129: Have you run this one yet? Would be curious to know what the runtime is

justfile
```diff
@@ -156,7 +156,7 @@ turbo-test-local-all *cmds="": && _timer
 			export FORGE_MOCK_4626_TARGET="true"; \
 			export FORGE_MOCK_TARGET_DECIMALS=${combination[1]}; \
 			export FORGE_MOCK_UNDERLYING_DECIMALS=${combination[1]}; \
-			forge test --no-match-path "*.tm*" {{ cmds }}; \
+			forge test --match-path "**/*.t.sol" --no-match-path "**/Adapter.t.sol" {{ cmds }}; \
```
jparklev, line 159: What was teh reason to stop running `Adapter.t.sol` for this step?

justfile
```diff
@@ -123,6 +125,40 @@ turbo-test-mainnet-match *exp="": && _timer
 
 turbo-test-mainnet-match-contract *exp="": && _timer
 	@cd {{ invocation_directory() }}; forge test --fork-url {{ MAINNET_RPC }} --match-contract {{ exp }}
+
+# run for all possible combinations between ERC20 and non-ERC20 target, underlying and stake and 6, 8 and 18 decimals
+# also run, for ERC4626 target, all possible combinations between 6, 8 and 18 target (and underlying) decimals
+turbo-test-local-all *cmds="": && _timer
+	for c in --target\ {true,false}\ --underlying\ {true,false}\ --stake\ {true,false}\ --tDecimals\ {6,8,18}\ --uDecimals\ {6,8,18}\ --sDecimals\ {6,8,18} ; do \
```
jparklev: These labels end up in `combination` but go unused because they're signposts, right? I wonder if they can be comments instead (unless i'm missing the use)?

jparklev (approved):
Mostly questions/clarifications on my end – assuming they don't bring anything new up, latest changes LGTM

package.json
fedealconada: I didn't add it cause we mention something like this could be trigger one in a while, it takes a while to run... But if we want to be on the safe side we could. Maybe we could run it only when merging the PR?

justfile
fedealconada: I did, but don't remember exactly how much time it took, in my computer I believe it was more than 1 hour (or more).

justfile
fedealconada: To be honest, I'm doing what we were already doing. But the reason  was that when running the Adapter.t.sol tests some test cases were not compatible with 4626 (maybe the mint function or something like that). We may want (quality work) to adapt them so they work in both cases: erc4626 and erc20.

package.json
jparklev: That's what I was thinking, yea! check if the branch is dev, and run it if so(?)

.github/workflows/ci.yml
fedealconada: 🤷🏼‍♂️

package.json
fedealconada: I've just added this!

justfile
fedealconada: Completely! I was using them just to give them a name and make it more "readable" but a comment would be enough, I agree.

fedealconada: Pushed!

jparklev (approved):
LGTM

abylaw (approved):
LGTM

PR #274, opened 2022-09-08
feat(core/Periphery): intrpret a zero address pool manager as a skip

jparklev:
## Motivation

Our current Pool Manager, points to Fuse, which is disabled and unusable at the moment, so we don't want to keep queuing Targets/PTs/LP shares there. And even if Fuse wasn't disabled, our current Pool Manager generally has a problem in that it reverts if the same Target is added twice 

## Solution

Since we already have a way to set the Pool Manager on the Periphery via admin function, we can encode a `null` state for the Manager (where interactions are skipped) for `address(0)`. 

In addition, i've added the option of a no-op Pool Manager that allows the Periphery to work without any additional changes

Resolves GEN-1399

## Implementer Checklist

### DIFF
--- a/pkg/fuse/src/NoopPoolManager.sol
+++ b/pkg/fuse/src/NoopPoolManager.sol
@@ -1,0 +1,33 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.11;
+
+// Pool manager implementation with no restrictions where every function is a no-op. Refer to "PoolManager.sol" for
+// the normal behavior of a Pool Manager.
+contract NoopPoolManager {
+    function deployPool(
+        string calldata name,
+        uint256 closeFactor,
+        uint256 liqIncentive,
+        address fallbackOracle
+    ) external returns (uint256 _poolIndex, address _comptroller) {
+        emit PoolDeployed(name, _comptroller, _poolIndex, closeFactor, liqIncentive);
+    }
+
+    function addTarget(address target, address adapter) external returns (address cTarget) {
+        emit TargetAdded(target, cTarget);
+    }
+
+    function queueSeries(
+        address adapter,
+        uint256 maturity,
+        address pool
+    ) external {
+        emit SeriesQueued(adapter, maturity, pool);
+    }
+
+    function addSeries(address adapter, uint256 maturity) external returns (address, address) { }
+
+    event PoolDeployed(string name, address comptroller, uint256 poolIndex, uint256 closeFactor, uint256 liqIncentive);
+    event TargetAdded(address indexed target, address indexed cTarget);
+    event SeriesQueued(address indexed adapter, uint256 indexed maturity, address indexed pool);
+}
--- a/pkg/fuse/src/NoopPoolManager.sol
+++ b/pkg/fuse/src/NoopPoolManager.sol
@@ -25,7 +25,7 @@ contract NoopPoolManager {
         emit SeriesQueued(adapter, maturity, pool);
     }
 
-    function addSeries(address adapter, uint256 maturity) external returns (address, address) { }
+    function addSeries(address adapter, uint256 maturity) external returns (address, address) {}
 
     event PoolDeployed(string name, address comptroller, uint256 poolIndex, uint256 closeFactor, uint256 liqIncentive);
     event TargetAdded(address indexed target, address indexed cTarget);
--- a/pkg/fuse/src/NoopPoolManager.sol
+++ b/pkg/fuse/src/NoopPoolManager.sol
@@ -1,33 +1,23 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 pragma solidity 0.8.11;
 
-// Pool manager implementation with no restrictions where every function is a no-op. Refer to "PoolManager.sol" for
-// the normal behavior of a Pool Manager.
+// Pool manager implementation with no restrictions where every function is a no-op. Refer to "PoolManager.sol"
+// for an exemplar of a normal Pool Manager.
 contract NoopPoolManager {
     function deployPool(
         string calldata name,
         uint256 closeFactor,
         uint256 liqIncentive,
         address fallbackOracle
-    ) external returns (uint256 _poolIndex, address _comptroller) {
-        emit PoolDeployed(name, _comptroller, _poolIndex, closeFactor, liqIncentive);
-    }
+    ) external returns (uint256 _poolIndex, address _comptroller) {}
 
-    function addTarget(address target, address adapter) external returns (address cTarget) {
-        emit TargetAdded(target, cTarget);
-    }
+    function addTarget(address target, address adapter) external returns (address cTarget) {}
 
     function queueSeries(
         address adapter,
         uint256 maturity,
         address pool
-    ) external {
-        emit SeriesQueued(adapter, maturity, pool);
-    }
+    ) external {}
 
     function addSeries(address adapter, uint256 maturity) external returns (address, address) {}
-
-    event PoolDeployed(string name, address comptroller, uint256 poolIndex, uint256 closeFactor, uint256 liqIncentive);
-    event TargetAdded(address indexed target, address indexed cTarget);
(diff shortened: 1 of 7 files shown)


jparklev:
To Do:

- [ ] Add new tests for these paths
- [ ] Confirm that we'd be able to backfill Targets for pool managers in case they're missed (might be that we just need to enable this on the Pool Manager-side going forward)

pkg/fuse/src/NoopPoolManager.sol
```diff
@@ -0,0 +1,33 @@
...
+// Pool manager implementation with no restrictions where every function is a no-op. Refer to "PoolManager.sol" for
+// the normal behavior of a Pool Manager.
+contract NoopPoolManager {
+    function deployPool(
+        string calldata name,
+        uint256 closeFactor,
+        uint256 liqIncentive,
+        address fallbackOracle
+    ) external returns (uint256 _poolIndex, address _comptroller) {
+        emit PoolDeployed(name, _comptroller, _poolIndex, closeFactor, liqIncentive);
```
fedealconada: do we want to emit this event it we are actually NOT deploying anything?

pkg/fuse/src/NoopPoolManager.sol
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    function addTarget(address target, address adapter) external returns (address cTarget) {
+        emit TargetAdded(target, cTarget);
```
fedealconada: Same as before

pkg/fuse/src/NoopPoolManager.sol
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    function queueSeries(
+        address adapter,
+        uint256 maturity,
+        address pool
+    ) external {
+        emit SeriesQueued(adapter, maturity, pool);
```
fedealconada: Same as before

pkg/fuse/src/tests/NoopPoolManager.tm.sol
```diff
@@ -0,0 +1,125 @@
...
+import { Errors } from "@sense-finance/v1-utils/src/libs/Errors.sol";
+import { DSTest } from "@sense-finance/v1-core/src/tests/test-helpers/test.sol";
+import { MockAdapter } from "@sense-finance/v1-core/src/tests/test-helpers/mocks/MockAdapter.sol";
+import { MockToken } from "@sense-finance/v1-core/src/tests/test-helpers/mocks/MockToken.sol";
+import { MockTarget } from "@sense-finance/v1-core/src/tests/test-helpers/mocks/MockTarget.sol";
+import { Hevm } from "@sense-finance/v1-core/src/tests/test-helpers/Hevm.sol";
+import { DateTimeFull } from "@sense-finance/v1-core/src/tests/test-helpers/DateTimeFull.sol";
+import { MockBalancerVault, MockSpaceFactory, MockSpacePool } from "@sense-finance/v1-core/src/tests/test-helpers/mocks/MockSpace.sol";
+
+contract NoopPoolManagerTest is DSTest {
```
fedealconada, line 25: I feel like these tests are more like "sanity checks", they are just asserting that calling no-op functions don't fail which is kind of what is expected.

I'd probably add some "integration" tests:

Test 1:
- Deploy a NoopPoolManager
- Call `periphery.setPoolManager` with the address of the new pool manager
- Call `periphery.deployAdapter()`
- Make respective assertions (prev call should have worked fine, events have been emitted (or not)).

Test 2:
- Call `periphery.setPoolManager` with null address
- Call `periphery.deployAdapter()`
- Make respective assertions (prev call should have worked fine, no events have been emitted on the pool manager side).

WDYT @jparklev ?

pkg/fuse/src/NoopPoolManager.sol
jparklev: Yea, I wasn't actually for sure here. I thought maybe indexers would be expecting it to verify certain calls? But now that I think about it more maybe it's a bit confusing

pkg/fuse/src/tests/NoopPoolManager.tm.sol
jparklev: Yea, I was thinking the same – for a quick pass last night I just duplicated the tests we had in `PoolManager.tm.sol` thinking that was a decent sanity check. But indeed a proper integration test showing we can still work with the Periphery like we would expect would be good! I'll add something

jparklev: Added with a few others! Lmk what you think ~

pkg/fuse/src/NoopPoolManager.sol
```diff
@@ -1,33 +1,23 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 pragma solidity 0.8.11;
 
-// Pool manager implementation with no restrictions where every function is a no-op. Refer to "PoolManager.sol" for
-// the normal behavior of a Pool Manager.
+// Pool manager implementation with no restrictions where every function is a no-op. Refer to "PoolManager.sol"
+// for an exemplar of a normal Pool Manager.
 contract NoopPoolManager {
```
fedealconada, line 6: Instead of having this Noop contract you could be using Foundy's `mockCall`, but I'm fine with this

pkg/fuse/src/tests/NoopPoolManager.tm.sol
```diff
@@ -0,0 +1,244 @@
...
+        poolManager.setParams("TARGET_PARAMS", paramsTarget);
+
+        poolManager.deployPool("Sense Pool", 0.051 ether, 1 ether, MASTER_ORACLE);
+
+        hevm.expectEmit(true, false, false, false);
+        emit TargetAdded(address(target), address(0));
+
+        periphery.verifyAdapter(address(mockAdapter), true);
+    }
+
```
fedealconada, line 232: We could probably add one more test for mainnet that would be [this](https://github.com/sense-finance/sense-v1/pull/274/files#diff-436d9c3fc5f5cd905555572e7160247590a368be185e8bb283471e56be636ae3R512) one you have on local tests. The one you call `periphery.setPoolManager(address(0))` (and make sure to use the our deployed periphery and not the new one you are deploying here in the test.

pkg/fuse/src/tests/NoopPoolManager.tm.sol
```diff
@@ -0,0 +1,244 @@
...
+    PoolManager internal poolManager;
+
+    MockBalancerVault internal balancerVault;
+    MockSpaceFactory internal spaceFactory;
+
+    BaseAdapter.AdapterParams internal adapterParams;
+
+    Hevm internal constant hevm = Hevm(HEVM_ADDRESS);
+
+    address public constant POOL_DIR = 0x835482FE0532f169024d5E9410199369aAD5C77E;
```
fedealconada: shall we move all these to the AddressBrook?

fedealconada (approved):
some small comments but LGTM

pkg/fuse/src/NoopPoolManager.sol
jparklev: What do you mean? The no-op is for live production usage (swap it out on the periphery)

pkg/fuse/src/tests/NoopPoolManager.tm.sol
jparklev: We could! Copied these from the normal PoolManager test for teh time being, but we could~

pkg/fuse/src/NoopPoolManager.sol
fedealconada: Ohhhh, right right! My baddd

pkg/fuse/src/NoopPoolManager.sol
jparklev: removed them in the latest commits

pkg/fuse/src/tests/NoopPoolManager.tm.sol
jparklev: Added!

PR #275, opened 2022-09-12
GEN-1454: add 4626 crop factory

fedealconada:
## Motivation

Same as we have the 4626 Factory and the 4626 Crops Factory, we also want to have a 4626 Crop Factory to support factories with only one reward token. 
Though the 46726 Crops Factory already has support for multiple reward tokens (which, of course, includes support for just 1), in the case that we now that the adapter will only be dealing with just 1 reward token, the former factory would be more expensive (gas wise) than the latter. 

## Solution

Create a new `ERC4626CropFactory.sol` and a hardhat task to deploy it.

## Implementer Checklist


jparklev (approved):
Pending passing tests, LGTM!

pkg/deployments/tasks/20220912-4626-crop-factory/index.js
```diff
@@ -0,0 +1,187 @@
...
+        }
+
+        console.log("\n-------------------------------------------------------");
+
+        // Unset deployer and set multisig as trusted address
+        console.log(`Set multisig as trusted address of 4626CropFactory`);
+        await (await factoryContract.setIsTrusted(senseAdminMultisigAddress, true)).wait();
+
+        console.log(`Unset deployer as trusted address of 4626CropFactory`);
+        await (await factoryContract.setIsTrusted(deployer, true)).wait();
```
jparklev: Isn't this true by default? Or should it be `false`?

pkg/deployments/tasks/20220912-4626-crop-factory/index.js
```diff
@@ -0,0 +1,187 @@
...
+          await stakeContract.approve(periphery.address, ethers.constants.MaxUint256).then(tx => tx.wait());
+
+          console.log(`Mint stake tokens...`);
+          await generateStakeTokens(stake, deployer, deployerSigner);
+
+          // deploy Series
+          for (let m of t.series) {
+            console.log(`\nSponsor Series for ${t.name} @ ${t.address} with maturity ${m}`);
+            await periphery.callStatic.sponsorSeries(adapterAddress, m, false);
+            await (await periphery.sponsorSeries(adapterAddress, m, false)).wait();
```
jparklev, line 162: OOC, why is `withPool` `false` here? The adapter will be verified so it'll deploy a pool anyway, so maybe it doesn't matter?

fedealconada: Right, it just does not matter haha. It's false because it has been false in other hardhat tasks... And this even runs only on hardhat fork.

pkg/deployments/tasks/20220912-4626-crop-factory/index.js
fedealconada: Good catch! Should be false!

pkg/core/src/tests/adapters/CropAdapter.t.sol
```diff
@@ -1095,17 +1095,14 @@ contract CropAdapters is TestHelper {
...
         if (!is4626) return;
-        ERC4626CropsAdapter a = ERC4626CropsAdapter(address(adapter));
-
-        address[] memory rewardTokens = new address[](1);
-        rewardTokens[0] = address(0xfede);
+        ERC4626CropAdapter a = ERC4626CropAdapter(address(adapter));
 
         hevm.expectEmit(true, false, false, true);
-        emit RewardTokensChanged(rewardTokens);
+        emit RewardTokenChanged(address(0xfede));
```
abylaw, line 1101: lol~

pkg/deployments/tasks/20220912-4626-crop-factory/index.js
```diff
@@ -0,0 +1,187 @@
...
+      oracle: masterOracleAddress,
+      factories,
+    } = data[chainId] || data[CHAINS.MAINNET];
+    let divider = new ethers.Contract(dividerAddress, dividerAbi, deployerSigner);
+    let periphery = new ethers.Contract(peripheryAddress, peripheryAbi, deployerSigner);
+    let masterOracle = new ethers.Contract(masterOracleAddress, oracleAbi, deployerSigner);
+
+    console.log("\n-------------------------------------------------------");
+    console.log("\nDeploy Factories");
+    for (let factory of factories) {
```
abylaw:
```suggestion
    for (const factory of factories) {
```

pkg/deployments/tasks/20220912-4626-crop-factory/index.js
```diff
@@ -0,0 +1,187 @@
...
+          minm,
+          maxm,
+          mode,
+          oracle: masterOracleAddress,
+          tilt,
+          guard,
+        })}}`,
+      );
+      const factoryParams = [masterOracleAddress, stake, stakeSize, minm, maxm, ifee, mode, tilt, guard];
+      const f = await deploy(factoryContractName, {
```
abylaw:
```suggestion
      const { address: factoryAddress, abi } = await deploy(factoryContractName, {
```

pkg/deployments/tasks/20220912-4626-crop-factory/index.js
```diff
@@ -0,0 +1,187 @@
...
+          console.log(`Mint stake tokens...`);
+          await generateStakeTokens(stake, deployer, deployerSigner);
+
+          // deploy Series
+          for (let m of t.series) {
+            console.log(`\nSponsor Series for ${t.name} @ ${t.address} with maturity ${m}`);
+            await periphery.callStatic.sponsorSeries(adapterAddress, m, false);
+            await (await periphery.sponsorSeries(adapterAddress, m, false)).wait();
+          }
+        }
```
abylaw, line 164: Notice that we have this same process of deploying adapters/series sponsorship repeated in a few of our tasks. can we pull parts out in some way or are they subtlely different?

pkg/deployments/tasks/20220912-4626-crop-factory/index.js
```diff
@@ -0,0 +1,187 @@
...
+          await (await factoryContract.supportTarget(t.address, true)).wait();
+
+          console.log(`\nOnboard target ${t.name} @ ${t.address} via ${factoryContractName}`);
+          const data = ethers.utils.defaultAbiCoder.encode(["address[]"], [[]]);
+          const adapterAddress = await periphery.callStatic.deployAdapter(factoryAddress, t.address, data);
+          await (await periphery.deployAdapter(factoryAddress, t.address, data)).wait();
+          console.log(`${t.name} adapter address: ${adapterAddress}`);
+
+          console.log(`\nCan call scale value`);
+          const ADAPTER_ABI = ["function scale() public view returns (uint256)"];
```
abylaw: nit: move to top of file

pkg/deployments/tasks/20220912-4626-crop-factory/index.js
```diff
@@ -0,0 +1,187 @@
...
+        await (await divider.setIsTrusted(factoryAddress, true)).wait();
+
+        await hre.network.provider.request({
+          method: "hardhat_stopImpersonatingAccount",
+          params: [senseAdminMultisigAddress],
+        });
+
+        console.log("\n-------------------------------------------------------");
+        console.log(`Deploy adapters for: ${factoryContractName}`);
+        for (let t of targets) {
```
abylaw: nit: `const t of`

abylaw (approved):
LGTM!

pkg/deployments/tasks/20220912-4626-crop-factory/index.js
fedealconada: completely, seems like we are repeating a lot of the same stuff for new factories/adapters tasks, i'll create a quality work task so we can tackle that there

fedealconada: Task created [here](https://linear.app/sense-finance/issue/GEN-1455/avoid-code-repetition-on-hardhat-tasks)

PR #276, opened 2022-09-13
feat(core/adapters): add morpho-specific adapter

jparklev:
## Motivation

Morpho distributes rewards periodically in chunks, rather than continuously

## Solution

Allow anyone to push the accrued rewards into an address of our chosing, from we could then use to airdrop the reward tokens retroactively to previous YT holders.

Also we shouldn't feel rushed to deploy this without tests or anything since we're not actually launching tomorrow, just wanted to get something on the table

## Implementer Checklist

### DIFF
--- a/pkg/core/src/adapters/implementations/morpho/MorphoAdapter.sol
+++ b/pkg/core/src/adapters/implementations/morpho/MorphoAdapter.sol
@@ -1,0 +1,69 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.11;
+
+// External references
+import { ERC20 } from "@rari-capital/solmate/src/tokens/ERC20.sol";
+import { SafeTransferLib } from "@rari-capital/solmate/src/utils/SafeTransferLib.sol";
+import { ERC4626 } from "@rari-capital/solmate/src/mixins/ERC4626.sol";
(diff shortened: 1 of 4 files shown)


pkg/core/src/adapters/implementations/morpho/MorphoAdapter.sol
```diff
@@ -0,0 +1,69 @@
...
+        address _divider,
+        address _target,
+        uint128 _ifee,
+        AdapterParams memory _adapterParams,
+        address _rewardRecipient
+    ) BaseAdapter(_divider, _target, address(ERC4626(_target).asset()), _ifee, _adapterParams) {
+        uint256 tDecimals = ERC4626(target).decimals();
+        BASE_UINT = 10**tDecimals;
+        SCALE_FACTOR = 10**(18 - tDecimals); // we assume targets decimals <= 18
+        rewardRecipient = _rewardRecipient;
```
fedealconada, line 38: only downside I see here is that, if we are not creating the reward recipient contract now, we will have to instantiate this with a normal address (like sense's cup address or something like that) and, if in the future we want to change this for a contract that knows how to distribute tokens, we would not be able todo this.

jparklev:
Archiving as we're going with a different solution (adding a general solution to this to the 4626 adapter)

PR #277, opened 2022-09-14
Noop pool manager task

jparklev:
## Motivation

Add legibility to pool manager swap

## Solution

Create a task to deploy, verify, and swap in the Noop Pool Manager on the Periphery


pkg/deployments/tasks/20220914-noop-pool-manager/index.js
```diff
@@ -0,0 +1,65 @@
...
+      const multisigSigner = await hre.ethers.getSigner(senseAdminMultisigAddress);
+      periphery = periphery.connect(multisigSigner);
+
+      console.log(`\Set Noop Pool Manager on the Periphery`);
+      await (await periphery.setPoolManager(poolManagerAddress)).wait();
+
+      await hre.network.provider.request({
+        method: "hardhat_stopImpersonatingAccount",
+        params: [senseAdminMultisigAddress],
+      });
```
fedealconada, line 62: i know we have the tests but, if you want (not mandatory), you could add a sanity check call to `periphery.deployAdapter()` so that will call the no-op contract and we check that's all good.

fedealconada (approved):
Very small comment but LGTM.

Feel free to run this task and then copy output into a new `/output` folder (check the other tasks for reference) before merging this PR (or you can then just commit directly to dev). In the future we should automate this process of copying things over.

pkg/deployments/tasks/20220914-noop-pool-manager/index.js
jparklev: Good shout – added :)

jparklev:
Pushed the output @fedealconada – give it a quick look?

fedealconada:
> Pushed the output @fedealconada – give it a quick look?

looks good! (I was always copying the solcInputs folder as well, but I guess it's not really needed).
good to merge!

PR #283, opened 2022-09-15
GEN-1381 dependabot

fedealconada:
## Motivation

We have been receiving some bug submissions on Immunefi due bugs in some of the dependencies we are using (most of them not having an impact on us).

We need to keep up to date with these kind of bugs so make sure we update them and prevent future bugs.

## Solution

- Add Dependabot to alert us about new dependency bugs
- Allow Dependabot to create PRs with the fixes though DO NOT allow it to merge those changes automatically (we need to review this changes in case they are breaking something).

## Implementer Checklist

### DIFF
--- a/pkg/core/package.json
+++ b/pkg/core/package.json
@@ -38,7 +38,7 @@
         "gas-snapshot": "just gas-snapshot"
     },
     "dependencies": {
-        "@openzeppelin/contracts": "4.3.2",
+        "@openzeppelin/contracts": "^4.7.3",
         "@rari-capital/solmate": "rari-capital/solmate#c8278b3cb948cffda3f1de5a401858035f262060",
         "@sense-finance/v1-fuse": "workspace:pkg/fuse",
         "@sense-finance/v1-utils": "workspace:pkg/utils"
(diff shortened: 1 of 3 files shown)


jparklev (approved):
Don't think (?) we're using those OZ contracts anywhere, so nothing we need to check on, right?

fedealconada:
> Don't think (?) we're using those OZ contracts anywhere, so nothing we need to check on, right?

Glad you asked! I should have written it down. Actually, we are. The only place (apart from tests where we use their ERC20 somewhere) where we are using it is on the `Divider` that we are using the `Pausable.sol` from OZ. But we are good with this update since it does not have an any breaking change. This contract got a very small update they made but, as I said, without breaking changes.

jparklev:
> > Don't think (?) we're using those OZ contracts anywhere, so nothing we need to check on, right?
> 
> Glad you asked! I should have written it down. Actually, we are. The only place (apart from tests where we use their ERC20 somewhere) where we are using it is on the `Divider` that we are using the `Pausable.sol` from OZ. But we are good with this update since it does not have an any breaking change. This contract got a very small update they made but, as I said, without breaking changes.

Got it, ty for confirming

PR #289, opened 2022-09-19
GEN-1416: Euler Adapter with Yield Daddy + extractable rewards

fedealconada:
## Motivation

We want to be able to have Euler adapters and we want to take advantage of our ERC4626 Factory.
Since Euler markets are not compliant, we are using the [Yield Daddy](https://github.com/timeless-fi/yield-daddy) (Timeless Finance) wrappers which make Euler markets compliant with the standard.

But, one caveat is that these wrappers do not handle rewards. Even though at the moment Euler is not distributing rewards on lending, we want to have some mechanism to be able to extract those rewards in case they start distributing them.

## Solution
- [x] Fork yield daddy's ERC4626 base factory (and Euler factory as well)
- [x] Fork yield daddy's ERC4626 base wrapper (and Euler wrapper as well)
- [x] Add extractToken method to the base wrapper to be able to extract tokens
- [ ] Add tests (there are some, but I'm working on them)

## Implementer Checklist


pkg/core/src/adapters/abstract/erc4626/timeless/base/ExtractableReward.sol
```diff
@@ -0,0 +1,57 @@
...
+pragma solidity 0.8.11;
+
+import { ERC20 } from "@rari-capital/solmate/src/tokens/ERC20.sol";
+import { SafeTransferLib } from "@rari-capital/solmate/src/utils/SafeTransferLib.sol";
+import { Trust } from "@sense-finance/v1-utils/src/Trust.sol";
+import { Errors } from "@sense-finance/v1-utils/src/libs/Errors.sol";
+
+/// @title ExtractableReward
+/// @notice Allows to extract rewards from the contract to the `rewardsRecepient`
+abstract contract ExtractableReward is Trust {
```
jparklev: What was the reason to do it this way rather than inheriting from `EulerERC4626` and adding these functions?

jparklev (commented):
This largely looks good to me! I do wish there were a way to import the contracts from timeless directly (ie like `import { EulerETokenMock }  from "@timeless/mocks/...`), you know?

fedealconada:
> This largely looks good to me! I do wish there were a way to import the contracts from timeless directly (ie like `import { EulerETokenMock } from "@timeless/mocks/...`), you know?

you mean to avoid all the '../../../'?

pkg/core/src/adapters/abstract/erc4626/timeless/base/ExtractableReward.sol
fedealconada: I did it that way thinking that, maybe, in a future, we may have more wrappers for different protocols and we may want to use it, WDYT?

jparklev:
> > This largely looks good to me! I do wish there were a way to import the contracts from timeless directly (ie like `import { EulerETokenMock } from "@timeless/mocks/...`), you know?
> 
> 
> 
> you mean to avoid all the '../../../'?

Like having timeless as a dependency rather than copying all of the contracts over

fedealconada:
> > > This largely looks good to me! I do wish there were a way to import the contracts from timeless directly (ie like `import { EulerETokenMock } from "@timeless/mocks/...`), you know?
> > 
> > 
> > you mean to avoid all the '../../../'?
> 
> Like having timeless as a dependency rather than copying all of the contracts over

I did it that way at the beginning, but their contracts have a different compiler version

jparklev:
> > > > This largely looks good to me! I do wish there were a way to import the contracts from timeless directly (ie like `import { EulerETokenMock } from "@timeless/mocks/...`), you know?
> 
> > > 
> 
> > > 
> 
> > > you mean to avoid all the '../../../'?
> 
> > 
> 
> > Like having timeless as a dependency rather than copying all of the contracts over
> 
> 
> 
> I did it that way at the beginning, but their contracts have a different compiler version 

Ah that's right – now I remember you mentioning. Perhaps we should consider bumping our codebase? Let me check the solidity changelog real quick

fedealconada:
> > > > > This largely looks good to me! I do wish there were a way to import the contracts from timeless directly (ie like `import { EulerETokenMock } from "@timeless/mocks/...`), you know?
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > you mean to avoid all the '../../../'?
> > 
> > 
> > > 
> > 
> > 
> > > Like having timeless as a dependency rather than copying all of the contracts over
> > 
> > 
> > I did it that way at the beginning, but their contracts have a different compiler version
> 
> Ah that's right – now I remember you mentioning. Perhaps we should consider bumping our codebase? Let me check the solidity changelog real quick

Oh, yes, definitely that could be an option. Sure, let me know!

jparklev:
> > > > > > This largely looks good to me! I do wish there were a way to import the contracts from timeless directly (ie like `import { EulerETokenMock } from "@timeless/mocks/...`), you know?
> > > 
> > > 
> > > > >
> > > 
> > > 
> > > > >
> > > 
> > > 
> > > > > you mean to avoid all the '../../../'?
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > Like having timeless as a dependency rather than copying all of the contracts over
> > > 
> > > 
> > > I did it that way at the beginning, but their contracts have a different compiler version
> > 
> > 
> > Ah that's right – now I remember you mentioning. Perhaps we should consider bumping our codebase? Let me check the solidity changelog real quick
> 
> Oh, yes, definitely that could be an option. Sure, let me know!

Think we're compatible and safe to bump to `0.8.13`! Main changes are around user-defined types, which we're not using yet, and `abi.encodeCall`, which we're also not using. Besides that, it's largely compiler updates an bug fixes

pkg/core/lib/yield-daddy/.gitmodules
```diff
@@ -0,0 +1,6 @@
+[submodule "lib/solmate"]
```
jparklev: shall we `.gitignore` the yield daddy files?

fedealconada: yeah, i think there was an issue with how the module got added, because modules files are supposed to be ignored by default, i've just fixed that

jparklev (approved):
pending passing CI, think it looks good!

fedealconada:
> pending passing CI, think it looks good!

Foundry bug has been fixed and build is passing :) Shall I merge? cc @jparklev

PR #290, opened 2022-09-19
chore: add extractToken

fedealconada:
## Motivation

Even though we have Crop and Crops solutions to handle airdropped or claimed reward tokens, it is hard to predict all the different mechanism protocols create to distribute their reward tokens.

Because of this, we need to provide a method that would allow a trusted address to extract an arbitrary token from it. 

## Solution

- Create an `extractToken` function that transfers all the adapter's balance for a given token (which can NOT be the stake, target nor a reward token in case of Crop(s)) to the `rewardsRecipient` address
- Provide a method to update the `rewardsRecipient` address

## Implementer Checklist

### DIFF
--- a/pkg/core/src/adapters/abstract/BaseAdapter.sol
+++ b/pkg/core/src/adapters/abstract/BaseAdapter.sol
@@ -215,5 +215,5 @@ abstract contract BaseAdapter is Trust, IERC3156FlashLender {
     /* ========== LOGS ========== */
 
     event RewardsRecipientChanged(address indexed oldRecipient, address indexed newRecipient);
-    event RewardsClaimed(address indexed token, address indexed recipient);
+    event RewardsClaimed(address indexed token, address indexed recipient, uint256 indexed amount);
 }
--- a/pkg/core/src/adapters/abstract/BaseAdapter.sol
+++ b/pkg/core/src/adapters/abstract/BaseAdapter.sol
@@ -64,9 +64,7 @@ abstract contract BaseAdapter is Trust, IERC3156FlashLender {
         /// (e.g. 0101 enables `collect` and `issue`, but not `combine`)
         uint48 level;
         /// @notice 0 for monthly, 1 for weekly
-        uint8 mode;
-        /// @notice Reward type: 0 for non-crop, 1 for crop and 2 for crops
-        uint8 rType;
+        uint16 mode;
     }
 
     /* ========== METADATA STORAGE ========== */
@@ -161,30 +159,7 @@ abstract contract BaseAdapter is Trust, IERC3156FlashLender {
 
     /// @notice Transfers reward tokens from the adapter to Sense's reward container
     /// @dev If adapter is either Crop or Crops, we check token is not any of the reward tokens
-    function extractToken(address token) external {
-        if (adapterParams.rType == 1) {
-            if (token == Crop(address(this)).reward()) revert Errors.TokenNotSupported();
-        } else if (adapterParams.rType == 2) {
-            uint256 i = 0;
-            Crops adapter = Crops(address(this));
-            while (true) {
-                try adapter.rewardTokens(i) returns (address rewardToken) {
-                    if (token == rewardToken) revert Errors.TokenNotSupported();
-                } catch {
-                    break;
-                }
-                unchecked {
-                    ++i;
-                }
-            }
-        }
-
-        // Check that token is neither the target nor the stake
-        if (token == target || token == adapterParams.stake) revert Errors.TokenNotSupported();
-        ERC20 t = ERC20(token);
-        t.safeTransfer(rewardsRecipient, t.balanceOf(address(this)));
-        emit RewardsClaimed(token, rewardsRecipient);
-    }
+    function extractToken(address token) external virtual;
 
     /* ========== OPTIONAL HOOKS ========== */
 
@@ -227,10 +202,6 @@ abstract contract BaseAdapter is Trust, IERC3156FlashLender {
 
     function mode() external view returns (uint256) {
         return adapterParams.mode;
-    }
-
-    function rType() external view returns (uint256) {
-        return adapterParams.rType;
     }
 
     function tilt() external view returns (uint256) {
--- a/pkg/core/src/adapters/abstract/BaseAdapter.sol
+++ b/pkg/core/src/adapters/abstract/BaseAdapter.sol
@@ -165,14 +165,16 @@ abstract contract BaseAdapter is Trust, IERC3156FlashLender {
         if (adapterParams.rType == 1) {
             if (token == Crop(address(this)).reward()) revert Errors.TokenNotSupported();
         } else if (adapterParams.rType == 2) {
-            bool keepGoing = true;
             uint256 i = 0;
             Crops adapter = Crops(address(this));
-            while (keepGoing) {
+            while (true) {
                 try adapter.rewardTokens(i) returns (address rewardToken) {
                     if (token == rewardToken) revert Errors.TokenNotSupported();
                 } catch {
-                    keepGoing = false;
+                    break;
+                }
+                unchecked {
+                    ++i;
                 }
             }
         }
--- a/pkg/core/src/adapters/abstract/BaseAdapter.sol
+++ b/pkg/core/src/adapters/abstract/BaseAdapter.sol
@@ -9,10 +9,13 @@ import { IERC3156FlashBorrower } from "../../external/flashloan/IERC3156FlashBor
 
 // Internal references
 import { Divider } from "../../Divider.sol";
+import { Crop } from "./extensions/Crop.sol";
+import { Crops } from "./extensions/Crops.sol";
 import { Errors } from "@sense-finance/v1-utils/src/libs/Errors.sol";
+import { Trust } from "@sense-finance/v1-utils/src/Trust.sol";
 
 /// @title Assign value to Target tokens
-abstract contract BaseAdapter is IERC3156FlashLender {
+abstract contract BaseAdapter is Trust, IERC3156FlashLender {
     using SafeTransferLib for ERC20;
 
     /* ========== CONSTANTS ========== */
@@ -32,6 +35,9 @@ abstract contract BaseAdapter is IERC3156FlashLender {
 
     /// @notice Issuance fee
     uint128 public immutable ifee;
+
+    /// @notice Rewards recipient
+    address public rewardsRecipient;
 
     /// @notice adapter params
     AdapterParams public adapterParams;
@@ -58,7 +64,9 @@ abstract contract BaseAdapter is IERC3156FlashLender {
         /// (e.g. 0101 enables `collect` and `issue`, but not `combine`)
         uint48 level;
         /// @notice 0 for monthly, 1 for weekly
-        uint16 mode;
+        uint8 mode;
+        /// @notice Reward type: 0 for non-crop, 1 for crop and 2 for crops
+        uint8 rType;
     }
 
     /* ========== METADATA STORAGE ========== */
@@ -71,12 +79,14 @@ abstract contract BaseAdapter is IERC3156FlashLender {
         address _divider,
         address _target,
         address _underlying,
+        address _rewardsRecipient,
         uint128 _ifee,
         AdapterParams memory _adapterParams
-    ) {
+    ) Trust(msg.sender) {
         divider = _divider;
         target = _target;
         underlying = _underlying;
+        rewardsRecipient = _rewardsRecipient;
         ifee = _ifee;
         adapterParams = _adapterParams;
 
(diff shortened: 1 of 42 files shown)


pkg/core/src/adapters/abstract/BaseAdapter.sol
```diff
@@ -144,6 +152,38 @@ abstract contract BaseAdapter is IERC3156FlashLender {
         return ERC20(token).balanceOf(address(this));
     }
 
+    function setRewardsRecipient(address recipient) external {
```
jparklev: Should be trusted, right?

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
```diff
@@ -101,4 +115,19 @@ abstract contract BaseFactory {
             } catch {}
         }
     }
+
+    function setAdmin(address _admin) external requiresTrust {
```
jparklev: What's the `admin` address for?

jparklev (commented):
Know it's a WIP, so just two quick comments ~

pkg/core/src/adapters/abstract/BaseAdapter.sol
fedealconada: indeed!

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
fedealconada: `admin` would be the trusted address we will set when deploying new adapters (`deployAdapter` call. If you feel like we should just keep what we had before, I'm good with that.

pkg/core/src/adapters/abstract/extensions/Crop.sol
```diff
@@ -30,7 +30,8 @@ abstract contract Crop is Trust {
 
     event Distributed(address indexed usr, address indexed token, uint256 amount);
 
-    constructor(address _divider, address _reward) Trust(msg.sender) {
+    constructor(address _divider, address _reward) {
+        setIsTrusted(msg.sender, true);
```
fedealconada: Do we want to add this here? Because it's already set on the base adapter.
The reason I've added it was kind of thinking about this contract in isolation. If one would, somehow, use it for something different and if we dont have this line, then it would be a problem. Thoughts?

pkg/core/src/adapters/abstract/extensions/Crops.sol
```diff
@@ -36,7 +36,8 @@ abstract contract Crops is Trust {
         mapping(address => uint256) rewarded;
     }
 
-    constructor(address _divider, address[] memory _rewardTokens) Trust(msg.sender) {
+    constructor(address _divider, address[] memory _rewardTokens) {
+        setIsTrusted(msg.sender, true);
```
fedealconada: Same as above

pkg/core/src/adapters/abstract/extensions/Crop.sol
jparklev: I feel like it's ok to stipulate that this contract should only be used with the base adapter (maybe we can put that in a comment on the top)!

Also is there duplication currently if we have trust on this contract and the base adapter?

fedealconada: > I feel like it's ok to stipulate that this contract should only be used with the base adapter (maybe we can put that in a comment on the top)!

sgtm! 

> Also is there duplication currently if we have trust on this contract and the base adapter?

AFAIK the compiler would only import it once

jparklev: > AFAIK the compiler would only import it once

Sweet, could we verify this? Either finding it in docs or trying crops with and without "Trust" and comparing contract sizes (or maybe there's another way?)

My only concern is if we'll have two instantiations of "Trust" that don't shares state (totally independent trusted addressees)

fedealconada: Yes! I'm on itt

pkg/core/src/adapters/abstract/BaseAdapter.sol
```diff
@@ -144,6 +154,36 @@ abstract contract BaseAdapter is IERC3156FlashLender {
...
+            if (token == Crop(address(this)).reward()) revert Errors.TokenNotSupported();
+        } else if (adapterParams.rType == 2) {
+            bool keepGoing = true;
+            uint256 i = 0;
+            Crops adapter = Crops(address(this));
+            while (keepGoing) {
+                try adapter.rewardTokens(i) returns (address rewardToken) {
+                    if (token == rewardToken) revert Errors.TokenNotSupported();
+                } catch {
+                    keepGoing = false;
```
jparklev: What do you think about doing a "break" instead of storing a var?

Also think we could ++i in an unchecked block for the ultimate micro gas optimization 😉

jparklev: Ex on line 510 of the roller https://github.com/sense-finance/auto-roller/blob/main/src/AutoRoller.sol

(Would link it exactly but on mobile)

fedealconada: > What do you think about doing a "break" instead of storing a var?

I just forgot about `break` at all xD, I'm adding it now

> Also think we could i++ in an unchecked block for the ultimate micro gas optimization 

This is actually a bug, I didn;'t even add the i++ at all 🤦

pkg/core/src/adapters/abstract/extensions/Crop.sol
fedealconada: So, I've tried these:
A) `forge flatten` -> output shows only one `Trust.sol`
B)
- Removed all usages of `Trust` from `Crop` but kept the inheritance and wrote down the contract size of `CAdapter`, who inherits `BaseAdapter` and `Crop` (and both contracts inherit `Trust`)
- Then, I removed the inheritance of `Trust` completely from `Crop` and checked the size and it matched the previous one (which makes sense).

C) Also, current tests would fail if we were to have different states (anyways, I've added a sanity check on `CropsAdapter.t.sol` -> `testAdapterHasParams`
D) Tried googling but couldn't find anything (still on it)

pkg/core/src/adapters/abstract/BaseAdapter.sol
jparklev: Looks like you added in commit 276, so will resolve ~

pkg/core/src/adapters/abstract/extensions/Crop.sol
jparklev: Beautiful!! With these I feel pretty comfortable. Maybe one last thing along the lines of (B) is that we could do this "Then, I removed the inheritance of Trust completely from Crop and checked the size and it matched the previous one (which makes sense)." and then compare the compiled bytecode in `out` from `forge build`, and that would confirm it

pkg/core/src/tests/test-helpers/Constants.sol
```diff
@@ -4,13 +4,18 @@ pragma solidity 0.8.11;
 /// @notice System constants
 library Constants {
     // adapter config
+    address public constant REWARDS_RECIPIENT = address(0xfed);
```
jparklev, line 7: 😆

pkg/core/src/adapters/implementations/compound/CFactory.sol
```diff
@@ -46,6 +47,7 @@ contract CFactory is CropFactory {
             minm: factoryParams.minm,
             maxm: factoryParams.maxm,
             mode: factoryParams.mode,
+            rType: factoryParams.rType,
```
jparklev: This could be hardcoded in the factories themselves, right? (e.g. since the CFactory will always be a Crop factory)

pkg/core/src/adapters/abstract/BaseAdapter.sol
```diff
@@ -144,6 +154,38 @@ abstract contract BaseAdapter is IERC3156FlashLender {
...
     }
 
+    function setRewardsRecipient(address recipient) external requiresTrust {
+        rewardsRecipient = recipient;
+        emit RewardsRecipientChanged(rewardsRecipient);
+    }
+
+    /// @notice Transfers reward tokens from the adapter to Sense's reward container
+    /// @dev If adapter is either Crop or Crops, we check token is not any of the reward tokens
+    function extractToken(address token) external {
```
jparklev: Out there idea i'd be curious to hear your thoughts on: what if we made this `virtual` and had each individual adapter override it and handle rewards in the correct way depending on if they're Crop(s) or not.

For ex, `ERC4626Adapter`, `ERC4626CropAdapter`, `CAdapter`, etc would implement it themselves. WDYT?

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
```diff
@@ -101,4 +111,23 @@ abstract contract BaseFactory {
...
+    }
+
+    /// Set an adapter's rewards recipient
+    function setAdapterRewardsRecipient(address adapter, address recipient) external requiresTrust {
+        BaseAdapter(adapter).setRewardsRecipient(recipient);
+    }
+
+    /* ========== LOGS ========== */
+
+    event AdminChanged(address indexed adapter);
```
jparklev: Unused?

pkg/core/src/tests/factories/BaseFactory.t.sol
```diff
@@ -296,4 +317,25 @@ contract Factories is TestHelper {
...
+        hevm.expectEmit(true, true, true, true);
+        emit RewardsRecipientChanged(address(0x111));
+
+        factory.setRewardsRecipient(address(0x111));
+        assertEq(factory.rewardsRecipient(), address(0x111));
+    }
+
+    /* ========== LOGS ========== */
+
+    event AdminChanged(address indexed adapter);
```
jparklev: Unused?

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
```diff
@@ -101,4 +111,23 @@ abstract contract BaseFactory {
             } catch {}
         }
     }
+
+    /// Set factory rewards recipient
+    /// @notice all future deployed adapters will have the new rewards recipient
+    /// @dev existing adapters rewards recipients will not be changed and can be
+    /// done through `setAdapterRewardsRecipient`
+    function setRewardsRecipient(address _recipient) external requiresTrust {
```
jparklev, line 118: Might consider emitting the previous recipient as well like we do here: https://github.com/sense-finance/sense-v1/blob/2b4ffebc784bde8c883b13b40bd221b57a1cb2ac/pkg/core/src/Periphery.sol#L425

jparklev (commented):
Got this out fast, nice work 🚢 A few comments last comments/suggestions, but mostly looking good

pkg/core/src/adapters/abstract/extensions/Crop.sol
fedealconada: Hmm... but the bytecodes will be different, won't they?
I've also checked some other stuff on the `out` like "`methodIdentifiers` and `sources` and checked that `Trust` only appears once and it is not duplicated.

pkg/core/src/adapters/implementations/compound/CFactory.sol
fedealconada: Oh, definitely!

pkg/core/src/adapters/abstract/BaseAdapter.sol
fedealconada: That could work as well, the pro: we dont have to add this `rType` to the adapter params (which i don't like it much), the con: we have to implement the same function every time we create a new adapter. Give the fact that we envision more and more 4626, hopefully the amount of custom adapters would decrease. And, also, implementing this function is no big deal (usually would be copy&paste).
Gonna add this change!

pkg/core/src/tests/test-helpers/mocks/MockAdapter.sol
```diff
@@ -546,8 +550,9 @@ contract Mock4626CropsAdapter is ERC4626Adapter, Crops {
         // Check that token is neither the target nor the stake
         if (token == target || token == adapterParams.stake) revert Errors.TokenNotSupported();
         ERC20 t = ERC20(token);
-        t.safeTransfer(rewardsRecipient, t.balanceOf(address(this)));
-        emit RewardsClaimed(token, rewardsRecipient);
+        uint256 tBal = t.balanceOf(address(this));
+        t.safeTransfer(rewardsRecipient, tBal);
```
jparklev, line 554: style thing that's fine either way, but I personally liked the previous code that had `t.balanceOf(address(this))` inline!

fedealconada: me too but i changed it so I can use it on the event

jparklev: mm yep, that's fair enough

pkg/core/src/adapters/abstract/extensions/Crop.sol
jparklev: Hmm, possibly just because i'm early in the day, but what's the reason the bytecodes would be different (if we're assuming Trust is de-duped)?

pkg/core/src/adapters/abstract/BaseAdapter.sol
jparklev: OK, great! Not positive which is better, but let me give the latest code a re-review and see how it feels!

pkg/core/src/adapters/abstract/extensions/Crop.sol
fedealconada: So, what I'm comparing is two versions of `CAdapter` bytecode where I'm modifying `Crop`.

- On the first version, I remove all usages of `Trust` inside `Crop` but I'm leaving the inheritance (both the line that imports `Trust.sol` and the `contract Crop is Trust`.
- On the second version, I'm completely removing `Trust` from `Crop`. 

The idea would be that both `CAdapter` versions would "weight" the same because in both cases there's only one `Trust` contract being used. Though, the bytecode would differ because one `Crop` contract has the `contract Crop is Trust` and the other one just `contract Crop`.

Unless I'm mistaken.

If this does not make sense it might be better to have a call to explain it better haha.

jparklev (approved):
LGTM

PR #296, opened 2022-09-20
GEN-1475: remove passthrough functions

fedealconada:
## Motivation

Some adapters have trusted functions. If those adapters are deployed through a factory, then it is only the factory who can call those functions.

To fix this, we have added some passthrough methods on the factory contracts whose purpose is just to call the adapters method. 

We may want to remove these functions if we, on each adapter deployment, make the factory to add another trusted address (for example, an admin address). 

NOTE: We can also decide just not to do this and leave as it was. Feedback is welcome :)

## Solution

- Add `adapterAdmin` to `BaseFactory.sol`
- Create a method to change the current `adapterAdmin` (requires trust)
- On every `deployAdapter` method fomo factories, add `setIsTrusted(adapterAdmin, true)`
- Remove passthrough methods

## Implementer Checklist


pkg/deployments/hardhat.seed.js
```diff
@@ -364,28 +362,30 @@ const MAINNET_FACTORIES = [
...
     ifee: ethers.utils.parseEther("0.0025"),
     mode: 0,
+    rType: CROP,
     tilt: 0,
     reward: COMP_TOKEN.get(chainId),
     targets: CTARGETS(chainId),
-    crops: false,
     guard: ethers.utils.parseEther("100000"),
   }),
   // FFactory example
```
fedealconada: With this PR, FFactory became bigger in size so it's not "deployable". Since we are not using it I've commented it out but not sure if we want to solve this or not.

jparklev: Nah, I think this is right – I don't think we need to worry about the `FFactory`

pkg/core/src/adapters/abstract/BaseAdapter.sol
```diff
@@ -144,6 +154,38 @@ abstract contract BaseAdapter is IERC3156FlashLender {
         return ERC20(token).balanceOf(address(this));
     }
 
+    function setRewardsRecipient(address recipient) external requiresTrust {
```
jparklev: Same comment as https://github.com/sense-finance/sense-v1/pull/290#discussion_r975606904

pkg/core/src/adapters/abstract/BaseAdapter.sol
```diff
@@ -144,6 +154,38 @@ abstract contract BaseAdapter is IERC3156FlashLender {
...
+
+    /// @notice Transfers reward tokens from the adapter to Sense's reward container
+    /// @dev If adapter is either Crop or Crops, we check token is not any of the reward tokens
+    function extractToken(address token) external virtual {
+        if (adapterParams.rType == 1) {
+            if (token == Crop(address(this)).reward()) revert Errors.TokenNotSupported();
+        } else if (adapterParams.rType == 2) {
+            uint256 i = 0;
+            Crops adapter = Crops(address(this));
+            while (true) {
```
jparklev: I think we might also be able to do like a 
`for (uint256 i = 0; ;) { }`

to save ourselves from defining `uint256 i = 0`, but also i'm fine either way personally 🤷

pkg/core/src/adapters/abstract/BaseAdapter.sol
```diff
@@ -144,6 +154,38 @@ abstract contract BaseAdapter is IERC3156FlashLender {
...
+                    ++i;
+                }
+            }
+        }
+
+        // Check that token is neither the target nor the stake
+        if (token == target || token == adapterParams.stake) revert Errors.TokenNotSupported();
+        ERC20 t = ERC20(token);
+        t.safeTransfer(rewardsRecipient, t.balanceOf(address(this)));
+        emit RewardsClaimed(token, rewardsRecipient);
```
jparklev: Maybe can emit the `balance` claimed as well?

jparklev (commented):
A quick comment or two. Should this be compared against https://github.com/sense-finance/sense-v1/pull/290?

pkg/core/src/adapters/abstract/BaseAdapter.sol
fedealconada: yes! that's was the intention :)

pkg/core/src/adapters/abstract/BaseAdapter.sol
fedealconada: agree!

jparklev (approved):
LGTM! Only comment is that we might want to call it something besides `adapterAdmin` since that suggests that the adapter has an admin that can do whatever they want, when really their powers are a bit limited. Maybe `restrictedAdmin` or something?

fedealconada:
> LGTM! Only comment is that we might want to call it something besides `adapterAdmin` since that suggests that the adapter has an admin that can do whatever they want, when really their powers are a bit limited. Maybe `restrictedAdmin` or something?

done sir

jparklev:
> > LGTM! Only comment is that we might want to call it something besides `adapterAdmin` since that suggests that the adapter has an admin that can do whatever they want, when really their powers are a bit limited. Maybe `restrictedAdmin` or something?
> 
> done sir

Awesome 💯 Now let's ship these factories!

PR #301, opened 2022-09-28
GEN-1421: general quality work

fedealconada:
## Motivation

Just some general improvements and fixes here and there (details below).

## Solution

* add forge-std and remove DSTestPlus and DSTest
* remove un-needed scripts from package.json
* polish dependenciess on package.json
* fix husky hooks
* remove all turbo scripts
* update README
* update CI file
* remove just
* fix foundry files
* remove Hevm.sol
* remove LiquidityHelper.sol
* remove test.sol
* use assertApproxEqAbs
* remove test.sol

## Implementer Checklist


pkg/deployments/cli.js
```diff
@@ -1,54 +0,0 @@
-const { task } = require("hardhat/config");
```
fedealconada, line 1: I have removed this file, are we using it somewhere? I was not sure... cc @jparklev

package.json
```diff
@@ -10,38 +10,32 @@
...
+    "test:4626": "yarn workspace @sense-finance/v1-core run test:4626 | gnomon",
     "lint": "yarn workspaces foreach --parallel --verbose run lint",
     "fix": "yarn workspaces foreach --parallel --verbose run fix",
     "gas-snapshot": "yarn workspaces foreach --parallel --verbose --interlaced run gas-snapshot"
   },
   "packageManager": "yarn@3.1.0",
   "devDependencies": {
-    "@types/prettier": "^2",
-    "dotenv": "^10.0.0",
+    "gnomon": "^1.5.0",
```
jparklev: nice ~ had you used this before?

pkg/core/package-lock.json
```diff
@@ -0,0 +1,13 @@
+{
```
jparklev: do we need a package lock if we're using yarn?

pkg/core/src/tests/test-helpers/TestHelper.sol
```diff
@@ -624,11 +580,56 @@ contract TestHelper is DSTest {
...
+            vm.warp(block.timestamp + 1 days);
         } else {
             MockERC4626(t).asset();
             MockToken underlying = MockToken(address(MockERC4626(t).asset()));
             underlying.mint(t, underlying.balanceOf(t));
         }
     }
+
+    /// @notice Tries to read env vars and/or sets defaut value if not exists
+    function setEnv() internal {
```
jparklev, line 597: woah, cool idea

jparklev (approved):
hell ya!! we could also migrate to `assertApproxEqAbs ` or `assertApproxEqRel` in `Test` from our current `assertClose` now or in the future as well https://github.com/foundry-rs/forge-std/blob/cb69e9c07fbd002819c8c6c8db3caeab76b90d6b/src/Test.sol#L399

package.json
fedealconada: nope! first time, but makes the trick :)

pkg/core/package-lock.json
fedealconada: nope haha, i might have done some `npm install` at some point, removing it :)

fedealconada:
> hell ya!! we could also migrate to `assertApproxEqAbs ` or `assertApproxEqRel` in `Test` from our current `assertClose` now or in the future as well https://github.com/foundry-rs/forge-std/blob/cb69e9c07fbd002819c8c6c8db3caeab76b90d6b/src/Test.sol#L399

not sure I follow... i've already removed all of our `assertClose` in favour of `assertApproxEqAbs`

jparklev:
> not sure I follow... i've already removed all of our assertClose in favour of assertApproxEqAbs

Ha, yea thought that might've been the case after I sent that message 😅 in any case, sgtm

fedealconada:
friendly reminder @abylaw to review this PR :)

README.md
```diff
@@ -82,13 +82,18 @@ cargo install just
 ```bash
 # Get contract dependencies
 git submodule update --init --recursive
-yarn isntall # or npm install
+yarn install # or npm install
 
 # Run local tests
```
abylaw: Could we elaborate a little bit how local tests are different from mainnet fork tests?

README.md
```diff
@@ -82,13 +82,18 @@ cargo install just
...
-just turbo-test-local
+yarn test
 
 # Run mainnet fork tests
-just turbo-test-mainnet
+yarn test:mainnet
```
abylaw, line 94: reading the comments below, it's not clear to me which tests `yarn test` & `yarn test:mainnet` run, if not the full local test suite.

Is it "run smoke tests"? ideally we would list out the set of tests run. (similar to below)

pkg/core/src/tests/adapters/WstETHAdapter.tm.sol
```diff
@@ -68,11 +67,9 @@ contract WstETHAdapterTestHelper is LiquidityHelper, DSTest {
     uint8 public constant DEFAULT_MODE = 0;
     uint64 public constant DEFAULT_TILT = 0;
 
-    Hevm internal constant hevm = Hevm(HEVM_ADDRESS);
-
     function setUp() public {
-        giveTokens(AddressBook.WETH, 10e18, hevm);
-        giveTokens(AddressBook.WSTETH, 10e18, hevm);
+        deal(AddressBook.WETH, address(this), 1e18);
+        deal(AddressBook.WSTETH, address(this), 1e18);
```
abylaw, line 72: wow! so they built into forge a bunch of convenience functions?

pkg/core/src/tests/adapters/FAdapter.tm.sol
```diff
@@ -42,26 +41,16 @@ contract FAdapterTestHelper is LiquidityHelper, DSTest {
...
     function setUp() public {
-        // Get RPC url from the environment
-        string[] memory inputs = new string[](2);
-        inputs[0] = "just";
-        inputs[1] = "_forge_rpc_url";
-        string memory rpcUrl = hevm.envString("MAINNET_RPC");
-
-        // Create fork from block mined on Apr 18 2022 at 12:00:10 AM UTC
-        // (before Fuse bug)
-        mainnetFork = hevm.createFork(rpcUrl, 14605885);
```
abylaw, line 58: how come creating fork is no longer necessary?

pkg/core/src/tests/adapters/CAdapter.tm.sol
```diff
@@ -110,6 +107,16 @@ contract CAdapterTestHelper is LiquidityHelper, DSTest {
...
         ); // Compound adapter
     }
+
+    function mintCToken(address cToken, uint256 mintAmount) internal {
+        if (cToken == AddressBook.cETH) {
+            AddressBook.cETH.call{ value: mintAmount }("");
+        } else {
+            ERC20 underlying = ERC20(CTokenLike(cToken).underlying());
+            underlying.safeApprove(cToken, mintAmount);
+            CTokenLike(AddressBook.cDAI).mint(mintAmount);
```
abylaw, line 119:
```suggestion
            CTokenLike(cToken).mint(mintAmount);
```
?

Also, why can we not use `deal` here?

pkg/core/src/tests/adapters/CropAdapter.t.sol
```diff
@@ -1224,26 +1224,26 @@ contract CropAdapters is TestHelper {
...
-        hevm.assume(tBal > 0.1 ether);
-        hevm.assume(tBal < 1000 ether);
+        vm.assume(tBal > 0.1 ether);
+        vm.assume(tBal < 1000 ether);
     }
 
-    function assertClose(uint256 a, uint256 b) public override {
-        assertClose(a, b, 1500);
+    function assertApproxEqAbs(uint256 a, uint256 b) public override {
+        assertApproxEqAbs(a, b, 1500);
```
abylaw, line 1246: it might be a bit surprising that this `assertApproxEqAbs` 1) exists here locally, and 2) asserts for different value than the one in `TestHelpers.sol`. Maybe rename this one?

pkg/core/src/tests/test-helpers/TestHelper.sol
abylaw: i like this too!

pkg/core/src/tests/fuzz.sh
```diff
@@ -0,0 +1,49 @@
...
+    echo ""; \
+    echo "${BIYellow}Target is ERC4626 with ${combination[3]} decimals"; \
+    echo "${BIYellow}Underlying is ERC4626 with ${combination[4]} decimals"; \
+    echo "${BIYellow}--------------------------------------------------"; \
+
+    export ERC4626_TARGET=true; \
+    export TARGET_DECIMALS=${combination[0]}; \
+    export UNDERLYING_DECIMALS=${combination[1]}; \
+    forge test --match-path "**/*.t.sol" --no-match-path "**/Adapter.t.sol"; \
+done
```
abylaw, line 49: love a good shell script 🥲

pkg/core/src/tests/fuzz.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+# Bold High Intensity Green
+BIGreen=$'\e[1;92m'
+
+# go to pkg/core directory
+cd "$(dirname "$3")"; \
```
abylaw: i've never seen these backslashes as being necessary within a script? wondering why?

abylaw (approved):
Nice!!!

Quickly wondering: why is it that forge manages to replace adding liquidity? it seems very specific to our protocol.

pkg/core/src/tests/adapters/WstETHAdapter.tm.sol
fedealconada: yes! they are constantly improving and adding new features/convenience functions :)

pkg/core/src/tests/adapters/FAdapter.tm.sol
fedealconada: What it has changed is that we are now using the same fork but rolling back to the block number [here](https://github.com/sense-finance/sense-v1/pull/301/files#diff-76bd75ca37488f203c24f8295e1db78cf4b671a8ef13f7cb5a3ff2e45b2c64dcR47)

pkg/core/src/tests/adapters/CAdapter.tm.sol
fedealconada: Good catch! I'm adding a comment but we can't use `deal()` because it breaks Compound's `redeem()` function. `deal()` is probably just replacing some memory slot that indicates one's balance but probably Compound has some logic so then, when calling `redeem()` the math does not make sense to the compound contract and it reverts.

pkg/core/src/tests/adapters/CropAdapter.t.sol
fedealconada: sgtm! renaming it to `assertApproxRewardBal`

pkg/core/src/tests/fuzz.sh
fedealconada: totally! when having this script on the justfile i was needing the backslashes and the forgot to remove them!

fedealconada:
> Nice!!!
> 
> Quickly wondering: why is it that forge manages to replace adding liquidity? it seems very specific to our protocol.

we were just having the add liquidity to set balances on addresses, but since we have `deal()` now we can use that instead

package.json
fedealconada: had to remove this one :( it was nice but, as the scripts needed to be piped with gnomon like:
`"test": "forge test --no-match-path \"*.tm*\" | gnomon",`

if then i wanted to call `yarn test` but appending some other forge filter like `--match-contract "CAdapters"` whatever was appended was discarded since it was after the pipe, something like:

`forge test --no-match-path \"*.tm*\" | gnomon --match-contract "CAdapters"` which is wrong.

I've decided to use the linux command `time`, pretty basic but works.

PR #305, opened 2022-10-04
GEN-1566: fix redeem restriction

fedealconada:
## Motivation

Fix bug on Divider about redeem restriction.

## Implementer Checklist

### DIFF
--- a/pkg/core/src/tests/Divider.t.sol
+++ b/pkg/core/src/tests/Divider.t.sol
@@ -476,21 +476,26 @@ contract Dividers is TestHelper {
         uint256 maturity = getValidMaturity(2021, 10);
 
         // Should be possible to init series
-        (, address yt) = periphery.sponsorSeries(address(adapter), maturity, true);
-
-        vm.startPrank(bob);
-        target.approve(address(adapter), type(uint256).max);
+        (address pt, address yt) = periphery.sponsorSeries(address(adapter), maturity, true);
 
         // Can't issue directly through the divider
         vm.expectRevert(abi.encodeWithSelector(Errors.IssuanceRestricted.selector));
+        vm.prank(bob);
         divider.issue(address(adapter), maturity, 1e18);
 
         // Can issue through adapter
-        adapter.doIssue(maturity, 1e18);
+        target.transfer(address(adapter), 1e18);
+
+        hevm.startPrank(address(adapter));
+        divider.issue(address(adapter), maturity, 1e18);
+        ERC20(pt).transfer(bob, ERC20(pt).balanceOf(address(adapter)));
+        ERC20(yt).transfer(bob, ERC20(yt).balanceOf(address(adapter)));
+        hevm.stopPrank();
 
         // It should still be possible to combine
-        divider.combine(address(adapter), maturity, ERC20(yt).balanceOf(bob));
-        vm.stopPrank();
+        uint256 ytBal = ERC20(yt).balanceOf(bob);
+        vm.prank(bob);
+        divider.combine(address(adapter), maturity, ytBal);
     }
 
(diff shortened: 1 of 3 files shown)


pkg/core/src/tests/test-helpers/mocks/MockAdapter.sol
```diff
@@ -504,6 +532,13 @@ contract Mock4626CropAdapter is ERC4626Adapter, Crop {
         MockToken(yt).transfer(msg.sender, MockToken(yt).balanceOf(address(this)));
     }
 
+    function doRedeem(uint256 maturity, uint256 ptBal) external {
```
jparklev: We could consider doing a forge `prank` rather than adding these `doRedeem`s. I _think_ it'd be OK, since we're not trying to test a specific adapter <> divider interaction here, but rather testing an address restriction. WDYT mr @fedealconada?

pkg/core/src/tests/Divider.t.sol
```diff
@@ -493,6 +493,38 @@ contract Dividers is TestHelper {
...
+
+        uint256 uBal = ERC20(pt).balanceOf(address(this));
+
+        // Can't issue directly through the divider
+        hevm.expectRevert(abi.encodeWithSelector(Errors.RedeemRestricted.selector));
+        divider.redeem(address(adapter), maturity, uBal);
+
+        // Can redeem through adapter
+        ERC20(pt).approve(address(adapter), type(uint256).max);
+        adapter.doRedeem(maturity, uBal);
```
jparklev: Should we do a sanity check that the user had their PTs burned and acquired more Target?

jparklev (approved):
LG – clean test 👌

pkg/core/src/tests/Divider.t.sol
fedealconada: yes ser

pkg/core/src/tests/test-helpers/mocks/MockAdapter.sol
fedealconada: completely, i guess i was overengineering. I have removed this `doRedeem` but also `doIssue` and `doCombine`.

jparklev: Cool :D

pkg/core/src/tests/Divider.t.sol
```diff
@@ -476,21 +476,71 @@ contract Dividers is TestHelper {
...
+        divider.setGuard(address(adapter), type(uint256).max);
+        uint256 maturity = getValidMaturity(2021, 10);
+
+        // Should be possible to sponsor series
+        (address pt, ) = periphery.sponsorSeries(address(adapter), maturity, true);
+
+        // Issue PTs andn YTs from user
+        divider.issue(address(adapter), maturity, 1e18);
+
+        // Issue PTs andn YTs from adapter (to test redeeming from adapter)
```
abylaw:
```suggestion
        // Issue PTs and YTs from adapter (to test redeeming from adapter)
```

pkg/core/src/tests/Divider.t.sol
```diff
@@ -476,21 +476,71 @@ contract Dividers is TestHelper {
...
+        hevm.prank(address(adapter));
+        divider.issue(address(adapter), maturity, 1e18);
+
+        // Move to maturity and settle
+        hevm.warp(maturity);
+        divider.settleSeries(address(adapter), maturity);
+
+        uint256 uBal = ERC20(pt).balanceOf(address(this));
+
+        // Can't issue directly through the divider
```
abylaw: issue -> redeem?

abylaw (approved):

PR #308, opened 2022-10-20
Morpho review - misc: identify files for review

KentonPrescott:
For Morpho team review


pkg/core/src/adapters/abstract/extensions/ExtractableReward.sol
```diff
@@ -1,60 +0,0 @@
...
-
-    /// @notice Receives a token address and returns whether it is an
-    /// extractable token or not
-    /// @dev To be overriden by the inheriting contract
-    function _isValid(address _token) internal virtual returns (bool);
-
-    /// @notice Transfers reward tokens from the adapter to Sense's reward container
-    function extractToken(address token) external {
-        if (!_isValid(token)) revert Errors.TokenNotSupported();
-        ERC20 t = ERC20(token);
```
Rubilmax, line 41: Just an advice:
this is pushing a new variable to the stack at runtime whereas you could just cast `token` to `ERC20(token)` everytime you need to call an `ERC20` method, which translates to no additional opcode nor additional stack variable at compile time (thus saves gas and stack depth)

pkg/core/src/adapters/abstract/extensions/ExtractableReward.sol
```diff
@@ -1,60 +0,0 @@
...
-        uint256 tBal = t.balanceOf(address(this));
-        t.safeTransfer(rewardsRecipient, t.balanceOf(address(this)));
```
Rubilmax, line 43: you could replace `t.balanceOf(address(this))` with `tBal` which was queried the line just above

pkg/core/src/adapters/abstract/extensions/ExtractableReward.sol
Rubilmax: You could also expect an `ERC20` as input, which is expected as an address once compiled. But I understand it may imply friction when using this endpoint from Solidity (because you need to statically cast `token` to `ERC20`)

pkg/core/src/adapters/abstract/extensions/ExtractableReward.sol
```diff
@@ -1,60 +0,0 @@
...
-        emit RewardsClaimed(token, rewardsRecipient, tBal);
-    }
```
MerlinEgalite, line 45: This function has no modifier. For now, I don't find an exploitable possibility (using reentrancy or something like that), but someone can still pass any address and the `msg.sender` would then be the `ERC4626` contract. Perhaps not an issue, but I would have added the modifier `requiresTrust` for Safety measure.

MerlinEgalite (approved):
Nice work, the code is very clear! 🎉

I just left a comment about a concern I have, but not sure this is really worrying. You know better the codebase than me, I guess you'll be better to assess if it's an issue or not.

pkg/core/src/adapters/abstract/extensions/ExtractableReward.sol
Rubilmax: For example, this function could get called by another contract, during an unsafe execution flow (not following the check-effects-interaction pattern, for example) which would allow a reentrancy and some unsafe aftermaths

pkg/core/src/adapters/abstract/extensions/ExtractableReward.sol
fedealconada: good catch! that was actually the idea but forgot to replace it :)

pkg/core/src/adapters/abstract/extensions/ExtractableReward.sol
fedealconada: thanks for the comment!! yeah, we've decided to make it external since this would always transfer the rewards to the defined `rewardsRecipient` address (and the only way to modify this is through `setRewardsRecipient` which has the `requiresTrust` modifier). 

as per the reentrancy thing, i'm gonna think a bit more about it but, at first glance, I don't think it represents an issue

pkg/core/src/adapters/abstract/extensions/ExtractableReward.sol
fedealconada: This is interesting! My understanding was that declaring the variable locally would be more gas-efficient. But, seems like you are right. I've anyways, out of curiosity, run the diff and here there are the results hehe

image

fedealconada: As per your second comment, yeah, we've kind of took the decision to always expect addresses instead of `ERC20` across the codebase :)

fedealconada:
> Nice work, the code is very clear! 🎉
> 
> I just left a comment about a concern I have, but not sure this is really worrying. You know better the codebase than me, I guess you'll be better to assess if it's an issue or not.

Thanks a ton for reviewing the code and for the comments! 🙏🏻

pkg/core/src/adapters/abstract/extensions/ExtractableReward.sol
MerlinEgalite: Makes sense, I just wanted to be sure!

fedealconada:
@jparklev shall we close this?

jparklev:
> @jparklev shall we close this?

sure, sgtm

PR #314, opened 2022-10-28
GEN-1632 factories deployments

fedealconada:
## Motivation

Deploy fixed 4626 factories

## Solution

- Deploys fixed 4626 factories
- I've also added the task for the re-deployment of the Periphery (thought it my be scheduled for later)  

## Implementer Checklist


pkg/deployments/tasks/20221028-factories-and-morpho/input.js
```diff
@@ -0,0 +1,84 @@
...
+    stakeSize: ethers.utils.parseEther("0.25"), // 0.25 WETH
+    minm: ((365.25 * 24 * 60 * 60) / 12).toString(), // 1 month
+    maxm: (10 * 365.25 * 24 * 60 * 60).toString(), // 10 years
+    mode: 0, // 0 = monthly
+    tilt: 0,
+    guard: ethers.utils.parseEther("100000"), // $100'000
+    targets: SAMPLE_TARGETS,
+  },
+  {
+    contractName: "@sense-finance/v1-core/src/adapters/abstract/factories/ERC4626Factory.sol:ERC4626Factory",
```
jparklev: why is this a path?

pkg/deployments/tasks/20221029-LATER-periphery/index.js
```diff
@@ -0,0 +1,170 @@
...
+    console.log("Adapters to verify:", adaptersVerified);
+
+    const newPeriphery = new ethers.Contract(peripheryAddress, peripheryAbi, deployerSigner);
+
+    for (let adapter of adaptersOnboarded) {
+      console.log("Onboarding adapter", adapter);
+      await newPeriphery.onboardAdapter(adapter, false).then(t => t.wait());
+    }
+
+    for (let adapter of adaptersOnboarded) {
```
jparklev: I believe we'll have to do this with factories, too (add factories that are on the existing periphery). I don't know if we had during the previous periphery upgrade?

jparklev: Also don't remember if we've had to remove adapters and if we should filter those out of this list somehow?

pkg/deployments/tasks/20221028-factories-and-morpho/index.js
```diff
@@ -0,0 +1,188 @@
...
+        for (const t of targets) {
+          periphery = periphery.connect(deployerSigner);
+
+          console.log(`\n- Add target ${t.name} to the whitelist`);
+          await (await factoryContract.supportTarget(t.address, true)).wait();
+
+          console.log(`\n- Onboard target ${t.name} @ ${t.address} via ${factoryContractName}`);
+          const data = ethers.utils.defaultAbiCoder.encode(["address[]"], [[]]);
+          const adapterAddress = await periphery.callStatic.deployAdapter(factoryAddress, t.address, data);
+          await (await periphery.deployAdapter(factoryAddress, t.address, data)).wait();
```
jparklev: So this fails with USDT rn, right?

pkg/deployments/tasks/20221029-LATER-periphery/index.js
```diff
@@ -0,0 +1,170 @@
...
+      console.log("\nSet the periphery on the Divider");
+      await (await divider.setPeriphery(peripheryAddress)).wait();
+
+      console.log("\nRemove the old Periphery auth over the pool manager");
+      await (await poolManager.setIsTrusted(oldPeriphery.address, false)).wait();
+
+      console.log("\nGive the new Periphery auth over the pool manager");
+      await (await poolManager.setIsTrusted(peripheryAddress, true)).wait();
+
+      // TOOD: we don't need this right? It was on the '20220313-periphery' but don't remember why?
```
jparklev: I believe that's right. I think we tried to batch this in with the last update, but it was extra

pkg/deployments/tasks/20221029-LATER-periphery/input.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  // mainnet
+  1: {
+    divider: "0x86bA3E96Be68563E41c2f5769F1AF9fAf758e6E0",
+    periphery: "0xFff11417a58781D3C72083CB45EF54d79Cd02437",
+    spaceFactory: "0x5f6e8e9C888760856e22057CBc81dD9e0494aA34",
+    poolManager: "0xf01eb98de53ed964AC3F786b80ED8ce33f05F417",
```
jparklev: This should be the noop pool manager from task "noop-pool-manager", right? (`0x9BFcA4882c1D8CDE3b4f787d3d74eb0E1bb0ec1B`)

pkg/deployments/tasks/20221029-LATER-periphery/index.js
```diff
@@ -0,0 +1,170 @@
...
+      // impersonate multisig
+      await hre.network.provider.request({
+        method: "hardhat_impersonateAccount",
+        params: [senseAdminMultisigAddress],
+      });
+
+      const multisigSigner = await hre.ethers.getSigner(senseAdminMultisigAddress);
+      oldPeriphery = oldPeriphery.connect(multisigSigner);
+      divider = divider.connect(multisigSigner);
+      poolManager = poolManager.connect(multisigSigner);
```
jparklev: since we're using the noop pool manager, we can probably get rid of all of the pool manager calls here

jparklev (commented):
A few quick comments before executing ~

pkg/deployments/tasks/20221028-factories-and-morpho/input.js
fedealconada: should have clarified! since there's a contract from yield daddy called `ERC4626Factory`, it conflicts with our own `ERC4626Factory` so hardhat makes you specify which exactly you want to usse

pkg/deployments/tasks/20221029-LATER-periphery/index.js
fedealconada: ohh, you're right, good call, will add it

pkg/deployments/tasks/20221028-factories-and-morpho/index.js
fedealconada: yep

pkg/deployments/tasks/20221029-LATER-periphery/input.js
fedealconada: Ohh, yes! I actually took this address from the Observable thinking it was the last version, we should update it there too!

pkg/deployments/tasks/20221029-LATER-periphery/index.js
```diff
@@ -71,6 +69,28 @@ task("20221029-LATER-periphery", "Deploys and authenticates a new Periphery").se
...
+      ),
+    ];
+    console.log("Factories to onboard:", factoriesOnboarded);
+
+    const factoriesVerified = [
+      ...new Set(
+        (await oldPeriphery.queryFilter(oldPeriphery.filters.FactoryChanged(null))).map(e => e.args.factory),
+      ),
+    ];
+    console.log("Factories to verify:", factoriesVerified);
```
jparklev: Should we be doing something with these?

pkg/deployments/tasks/20221029-LATER-periphery/index.js
```diff
@@ -50,7 +48,7 @@ task("20221029-LATER-periphery", "Deploys and authenticates a new Periphery").se
         ),
       ),
     ];
-    console.log("Adapters to onborad:", adaptersOnboarded);
+    console.log("Adapters to onboard:", adaptersOnboarded);
```
jparklev: Have we removed any adapters from the divider? do we need to filter those out somehow?

pkg/deployments/tasks/20221029-LATER-periphery/index.js
```diff
@@ -71,6 +69,28 @@ task("20221029-LATER-periphery", "Deploys and authenticates a new Periphery").se
...
+        (await oldPeriphery.queryFilter(oldPeriphery.filters.AdapterOnboarded(null))).map(
+          e => e.args.adapter,
+        ),
+      ),
+    ];
+    console.log("Factories to onboard:", factoriesOnboarded);
+
+    const factoriesVerified = [
+      ...new Set(
+        (await oldPeriphery.queryFilter(oldPeriphery.filters.FactoryChanged(null))).map(e => e.args.factory),
```
jparklev: should we filter out factories that we've unverified?

jparklev (commented):
A few last qs for clarity on verification and unsetting adapters ~

pkg/deployments/tasks/index.js
```diff
@@ -15,3 +15,5 @@ require("./20220922-4626-crop-factory");
 require("./20220922-4626-factory");
 require("./20220926-euler");
 require("./20221007-morpho-adapters");
+require("./20221031-factories-and-morpho");
+require("./20221029-LATER-periphery");
```
abylaw: should we call this task `periphery-v1.4 (i think???)` instead? more context :)

pkg/deployments/tasks/20221029-LATER-periphery/index.js
```diff
@@ -0,0 +1,144 @@
...
+    console.log("Adapters to verify:", adaptersVerified);
+
+    const newPeriphery = new ethers.Contract(peripheryAddress, peripheryAbi, deployerSigner);
+
+    for (let adapter of adaptersOnboarded) {
+      console.log("Onboarding adapter", adapter);
+      await newPeriphery.onboardAdapter(adapter, false).then(t => t.wait());
+    }
+
+    for (let adapter of adaptersOnboarded) {
```
abylaw:
```suggestion
    for (let adapter of adaptersVerified) {
```
?

pkg/deployments/tasks/20221029-LATER-periphery/index.js
```diff
@@ -0,0 +1,144 @@
...
+      await newPeriphery.setFactory(factory, true).then(t => t.wait());
+    }
+
+    console.log("\nAdding admin multisig as admin on Periphery");
+    await newPeriphery.setIsTrusted(senseAdminMultisigAddress, true).then(t => t.wait());
+
+    console.log("\nRemoving deployer as admin on Periphery");
+    await newPeriphery.setIsTrusted(deployer, false).then(t => t.wait());
+
+    // If we're in a forked environment, check the txs we need to send from the multisig as well
```
abylaw: meganit: move comment down into `else`?

pkg/deployments/tasks/20221029-LATER-periphery/index.js
```diff
@@ -0,0 +1,144 @@
...
+      ]);
+    } else {
+      // fund multisig
+      console.log(`\nFund multisig to be able to make calls from that address`);
+      await (
+        await deployerSigner.sendTransaction({
+          to: senseAdminMultisigAddress,
+          value: ethers.utils.parseEther("1"),
+        })
+      ).wait();
```
abylaw: `.then(t => t.wait())` to be consistent?

pkg/deployments/tasks/20221031-factories-and-morpho/index.js
```diff
@@ -0,0 +1,189 @@
...
+          maxm,
+          mode,
+          oracle: masterOracleAddress,
+          tilt,
+          guard,
+        })}}\n`,
+      );
+
+      const factoryParams = [masterOracleAddress, stake, stakeSize, minm, maxm, ifee, mode, tilt, guard];
+      const { address: factoryAddress, abi } = await deploy("ERC4626Factory_Sense", {
```
abylaw: does this name get surfaced anywhere? should it be `factoryContractName`?

pkg/deployments/tasks/20221031-factories-and-morpho/index.js
```diff
@@ -0,0 +1,189 @@
...
+        contract: "@sense-finance/v1-core/src/adapters/abstract/factories/ERC4626Factory.sol:ERC4626Factory",
```
abylaw: oh! interesting syntax. does it refer to a specific contract within that file?

pkg/deployments/tasks/20221031-factories-and-morpho/index.js
```diff
@@ -0,0 +1,189 @@
...
+        await hre.network.provider.request({
+          method: "hardhat_stopImpersonatingAccount",
+          params: [senseAdminMultisigAddress],
+        });
+
+        console.log("\n-------------------------------------------------------");
+        console.log(`\nDeploy adapters for: ${factoryContractName}`);
+        console.log("\n-------------------------------------------------------");
+
+        for (const t of targets) {
```
abylaw, line 118: wonder if we should pull out deploying adapters for factories into a separate task.

we'd reduce code duplication by quite a bit, iic.
so we'd run:
1) `20221031-factories-and-morpho`
2) `deploy-adapters`

or maybe pull adapters into a function?

pkg/deployments/tasks/20221031-factories-and-morpho/index.js
fedealconada: will be removed from here, but we need this notation to specify exactly which contract we are referring to since there's another contract with that exact name from a lib. we are using (yield-daddy)
I've added a comment about this :)

pkg/deployments/tasks/20221031-factories-and-morpho/index.js
fedealconada: to be honest, this happens on hardhat fork only, as is more like a sanity check when deploying a factory, just to make sure we can deploy an adapter for t and sponsor a series. In mainnet or goerli, we would deploy the adapter using OZ Defender.

fedealconada: but, we could definitely extract this out (maybe on a utils file) because it seems like something we are constantly using and it's copied over and over from one task to another one (i'll mark is as a quality work)

pkg/deployments/tasks/20221031-factories-and-morpho/index.js
fedealconada: i've removed this! it was temporary while I was trying some things, sorry!

fedealconada:
Oka, re-bumping this since the etherscan verification issue blocked my for a while.
Would need a re-review from you both @abylaw and @jparklev 🙏🏻
- I've addressed your last comments on the PR (@jparklev I'm filtering non-used adapters and factories for the Periphery re-deployment)
- I've fixed the issue with the verification. Context: seems like `hardhat-preprocessor` plugin conflicts we the verification task we were using (which, somehow, was not the one built-in with `hardhat-deploy`. Switching to that one worked well but with some caveats:
  - It makes me use the `solcInput` which, according to the docs, some time was needed as a workaround for a bug that was affecting verification for contracts using Solidity version < 0.8. Since we are using >0.8 we should not need this but somehow it only works with this param.
  - Also, after successfully verifying, when going to Etherscan I was seeing a warning saying that a library called `__CACHE_BREAKER` was not verified. According to `https://github.com/wighawag/hardhat-deploy/issues/78#issuecomment-786914537` this is something added when using `hardhat-preprocessor` but it's safe to remove. So, I've added a little workaround that, before submitting for verifying, I'm removing this library from the `solcInput` file. 

I've already asked about these 2 things [here](https://github.com/wighawag/hardhat-deploy/issues/78) and [here](https://github.com/wighawag/hardhat-preprocessor/issues/12#issuecomment-1310159980)

pkg/deployments/hardhat.utils.js
```diff
@@ -128,22 +128,35 @@ exports.setStorageAt = setStorageAt;
...
+  // since this was a bug supposedly fixed from Solidity version +0.8
+  // (https://github.com/wighawag/hardhat-deploy#4-hardhat-etherscan-verify)
+  // Also, verifying contracts when using `hardhat-preprocessor` plugin
+  // generates a warning on Etherscan saying that there's a library (__CACHE_BREAKER)
+  // that has not been verified. Hence, the lines below are removing this library
+  // from the solcInput file (https://github.com/wighawag/hardhat-deploy/issues/78#issuecomment-786914537)
+
+  // Remove __CACHE_BREAKER library from solcInput
+  const path = `${__dirname}/deployments/${hre.network.name}/solcInputs`;
+  fs.readdirSync(path).forEach(file => {
```
jparklev, line 143: Do you know why this wasn't appearing for us before? I suppose we'll wait for the responses on https://github.com/wighawag/hardhat-deploy/issues/78?

pkg/deployments/tasks/20221029-LATER-periphery-v1.4/index.js
```diff
@@ -0,0 +1,131 @@
...
+const data = require("./input");
+
+const { SENSE_MULTISIG, CHAINS, VERIFY_CHAINS } = require("../../hardhat.addresses");
+
+const dividerAbi = require("./abi/Divider.json");
+const peripheryAbi = require("./abi/Periphery.json");
+
+const { verifyOnEtherscan } = require("../../hardhat.utils");
+
+task("20221029-LATER-periphery-v1.4", "Deploys and authenticates a new Periphery").setAction(
```
jparklev: Echo @abylaw's suggestion here! Screen Shot 2022-11-10 at 4 02 11 PM

pkg/deployments/tasks/20221031-factories-and-morpho/index.js
```diff
@@ -0,0 +1,185 @@
...
+
+      const { address: factoryAddress, abi } = await deploy(factoryContractName, {
+        contract: !contract ? factoryContractName : contract,
+        from: deployer,
+        args: [
+          divider.address,
+          restrictedAdmin,
+          rewardsRecipient,
+          factoryParams,
+          ...(factoryContractName === "ERC4626CropFactory" ? [ethers.constants.AddressZero] : []),
```
jparklev: why is the reward token the zero address?

pkg/deployments/tasks/20221031-factories-and-morpho/index.js
```diff
@@ -0,0 +1,185 @@
...
+          oracle: masterOracleAddress,
+          tilt,
+          guard,
+        })}}\n`,
+      );
+
+      const factoryParams = [masterOracleAddress, stake, stakeSize, minm, maxm, ifee, mode, tilt, guard];
+
+      const { address: factoryAddress, abi } = await deploy(factoryContractName, {
+        contract: !contract ? factoryContractName : contract,
```
jparklev:
```suggestion
        contract: contract ? contract : factoryContractName,
```
#nit

pkg/deployments/tasks/20221029-LATER-periphery-v1.4/index.js
```diff
@@ -0,0 +1,131 @@
...
+          e => e.args.adapter,
+        ),
+      ),
+    ];
+    console.log("Adapters to verify & onboard:", adaptersOnboarded);
+
+    for (let adapter of adaptersOnboarded) {
+      console.log("\nOnboarding adapter", adapter);
+      const params = await divider.adapterMeta(adapter);
+      if (parseInt(params[2]) > 0) {
```
jparklev: could potentially do something like `params[2].gt(0)` or smtg here as well

jparklev (approved):
Didn't go as deep on the hardhat testing stuff, but deployment code LGTM 🦾

pkg/deployments/hardhat.utils.js
```diff
@@ -128,22 +128,35 @@ exports.setStorageAt = setStorageAt;
...
+    console.log(file);
+    var m = JSON.parse(fs.readFileSync(`${path}/${file}`).toString());
+    delete m?.settings?.libraries[""]?.__CACHE_BREAKER__;
+    fs.writeFileSync(`${path}/${file}`, JSON.stringify(m));
   });
+
+  // Try verification
```
abylaw: nit: remove comment

pkg/deployments/hardhat.utils.js
```diff
@@ -128,22 +128,35 @@ exports.setStorageAt = setStorageAt;
...
+  // (https://github.com/wighawag/hardhat-deploy#4-hardhat-etherscan-verify)
+  // Also, verifying contracts when using `hardhat-preprocessor` plugin
+  // generates a warning on Etherscan saying that there's a library (__CACHE_BREAKER)
+  // that has not been verified. Hence, the lines below are removing this library
+  // from the solcInput file (https://github.com/wighawag/hardhat-deploy/issues/78#issuecomment-786914537)
+
+  // Remove __CACHE_BREAKER library from solcInput
+  const path = `${__dirname}/deployments/${hre.network.name}/solcInputs`;
+  fs.readdirSync(path).forEach(file => {
+    console.log(file);
```
abylaw: is this for debugging or is there something going on here i don't understand? :) 
if for debugging, we should remove it.

pkg/deployments/tasks/20221029-LATER-periphery-v1.4/index.js
abylaw: also +1 on removing `later` :) we know it's for later!

pkg/deployments/tasks/20221031-factories-and-morpho/index.js
```diff
@@ -0,0 +1,185 @@
...
+        // fund multisig
+        console.log(`\n - Fund multisig to be able to make calls from that address`);
+        await (
+          await deployerSigner.sendTransaction({
+            to: senseAdminMultisigAddress,
+            value: ethers.utils.parseEther("1"),
+          })
+        ).wait();
+
+        // impersonate multisig account
```
abylaw: nit: remove comment

pkg/deployments/tasks/20221031-factories-and-morpho/index.js
```diff
@@ -0,0 +1,185 @@
...
+      const factoryContract = new ethers.Contract(factoryAddress, abi, deployerSigner);
+
+      console.log(`${factoryContractName} deployed to ${factoryAddress}`);
+
+      // if not hardhat fork, we try verifying on etherscan
+      if (chainId !== CHAINS.HARDHAT) {
+        console.log("\n-------------------------------------------------------");
+        await verifyOnEtherscan(factoryContractName);
+      } else {
+        // fund multisig
```
abylaw: nit: remove comment. `console.log` is enough

pkg/deployments/tasks/20221031-factories-and-morpho/index.js
```diff
@@ -0,0 +1,185 @@
...
+          factoryParams,
+          ...(factoryContractName === "ERC4626CropFactory" ? [ethers.constants.AddressZero] : []),
+        ],
+        log: true,
+      });
+      const factoryContract = new ethers.Contract(factoryAddress, abi, deployerSigner);
+
+      console.log(`${factoryContractName} deployed to ${factoryAddress}`);
+
+      // if not hardhat fork, we try verifying on etherscan
```
abylaw: nit: remove comment

pkg/deployments/tasks/20221031-factories-and-morpho/index.js
abylaw: i also wish to know...

pkg/deployments/tasks/20221031-factories-and-morpho/index.js
abylaw: great ok! makes sense~

abylaw (commented):
Yay!!

abylaw (approved):

pkg/deployments/hardhat.utils.js
fedealconada: nice catch hehe, it was just me debugging

pkg/deployments/tasks/20221031-factories-and-morpho/index.js
fedealconada: yep, we can actually do `contract || factoryContractName`

pkg/deployments/hardhat.utils.js
fedealconada: It was not because we are using a different verification task and the one built-in on the `hardhat-deploy` plugin. And seems like when using the preprocessor we should be using the `hardhat-deploy` built-in verification task.
Before, it was working because we were not making use of the preprocessor

pkg/deployments/tasks/20221031-factories-and-morpho/index.js
fedealconada: To be honest, it's good that you are asking. The `ERC4626CropFactory.sol` inherits `CropFactory` which requires a `reward` on the constructor. But, actually, on `ERC4626CropFactory` we should not require `reward` on the constructor because we will never use that (we will never have a common reward token across ERC4626 tokens haha).
So, I've just removed that param from the constructor since it was useless!

pkg/deployments/tasks/20221029-LATER-periphery-v1.4/index.js
fedealconada: done! i thought it was a string

pkg/deployments/tasks/20221031-factories-and-morpho/index.js
abylaw: Could we leave a comment so next person reading is not surprised? :)

fedealconada: Hm.. but this is gone now :)

PR #321, opened 2022-11-03
GEN-1377-coverage

fedealconada:
## Motivation

Have test coverage so we know where we are standing at with regards to tests, which ones are missing and where.

## Solution

- [x] fix remappings so `forge coverage` works well
- [x] add codecov and badge to README  
- [x] remove `Trust.sol` that we were not using
- [x] ignore contracts that don't need to be covered
 
## Implementer Checklist


codecov.yml
```diff
@@ -0,0 +1,7 @@
+# interfaces are automatically ignored by forge coverage
+ignore:
+  - "**/tests/**/*"                         # ignore all files inside tests/ folders
```
jparklev, line 3: Nice one, we should add this to the auto roller as well

pkg/core/.gas-snapshot
```diff
@@ -1,154 +1,319 @@
...
-PeripheryTest:testAdminVerifyAdapterWhenPermissionless() (gas: 3460413)
-PeripheryTest:testCantDeployAdapterIfTargetIsNotSupported() (gas: 1408212)
-PeripheryTest:testCantDeployAdapterIfTargetIsNotSupportedOnSpecificAdapter() (gas: 7088783)
-PeripheryTest:testCantMigrateLiquidityIfTargetsAreDifferent() (gas: 7168805)
-PeripheryTest:testCantVerifyAdapterNonAdmin() (gas: 3285407)
-PeripheryTest:testCantVerifyAdapterNonAdminWhenPermissionless() (gas: 3297226)
-PeripheryTest:testDeployAdapter() (gas: 3652687)
-PeripheryTest:testDeployAdapterWhenPermissionless() (gas: 3659960)
-PeripheryTest:testDeployPeriphery() (gas: 4236538)
+DateTimeTest:testFormatEleventhAndMay() (gas: 14054)
```
jparklev: I wonder if it's time to retire these gas snapshots

jparklev (commented):
I see that CI is passing with the new remappings! Good stuff.

I'm currently seeing this "coverage data is unkown" message on codecov, is there anything I need to do to see the breakdown?
Screen Shot 2022-11-03 at 10 31 33 AM

pkg/core/.gas-snapshot
fedealconada: you mean removing the CI.yml step (as well as the files)?

codecov.yml
fedealconada: I'll add it

pkg/core/.gas-snapshot
jparklev: yea!

fedealconada: sure! i'll remove it

jparklev (approved):
Confirmed w fede that codecov is looking good

abylaw (approved):
nice!! curious why we need to move things out of `src`?

fedealconada:
> nice!! curious why we need to move things out of `src`?

haha, not sure exactly why but that was making `forge coverage` to run, otherwise, it was throwing an error. If you add the `/src` back, and remove it from the `foundry.toml` files, throws the error 🤷🏼‍♂️

PR #327, opened 2022-12-05
GEN-1683: consolidate Goerli tokens

fedealconada:
## Motivation

Re-deploying a bunch of Goerli contracts so all our factories/adapters use the same targets and underlyings and avoid having multiple ones for them.  

## Solution

- [x] redeploy Goerli contracts
- [x] created a task that redeploys Goerli contracts
- [x] created task that verifies adapters on Etherscan

NOTE: in the near future we may migrate these scripts to foundry   

## Implementer Checklist

- [ ]  [OPTIONAL] Create a reference implementation (in python or JS)
- [ ]  Create a short "Motivation" section for the PR
- [ ]  Write a spec for the feature and put it in the PR description – basic function names and expected state transitions is OK
- [ ]  Get the PR reviewed by at least two people

If this is your first time reviewing smart contract changes
- [ ]  Review the [Solcurity](https://github.com/Rari-Capital/solcurity) standard

If smart contract changes were made
- [ ]  Check all of the new revert paths with concrete tests
- [ ]  Check all of the new storage slot writes with concrete tests – do not make the passing test case the trivial case!
- [ ]  Go line-by-line through the new code and ensure it has all been covered by concrete tests – since we don’t have a coverage engine for foundry yet, you can imagine how one might check that each different code path is covered
- [ ]  Simplify the implementation and spend some time trying to minimize gas costs
- [ ]  Write fuzz tests for the feature – try everything to break the implementation (is this function monotonically in/decreasing, should it always be less than something else, etc)
- [ ]  Capture bugs discovered during the above steps in concrete tests (regression tests)
- [ ]  Add integration tests – fuzz or concrete tests that test how the new code affects the entire system, either locally or with a mainnet fork
- [ ]  [OPTIONAL] Integrate the feature into the deployment scripts
- [ ]  [OPTIONAL] Add new sanity checks to the deployment scripts

If your PR uses or interacts with prices of any kind
- [ ] Make sure forge is running the test on the latest block
- [ ] Compare the value returned with some other source (like Curve)
- [ ] Check the `updatedAt` (if available) value of the contract and confirm that the price we are sourcing is in fact not older than X
- [ ] Ensure oracle liveness from the oracle maintainer, either through a public status interface or through written confirmation from a developer.

PR #328, opened 2022-12-06
GEN-1728: 4626 factory

fedealconada:
## Motivation

Add deployment script using forge scripts for the `ERC4626Factory` contract.

## Solution

NOTES:
- ~I've tried adding the `--verify` on the `package.json` but it throws an error~
- ~I tried running it on Goerli but fails (have to dig in a bit more)~
- In fork from mainnet works fine (didn't try in mainnet)
- ~If using the scripts defined in `package.json`, it needs you to append `--private-key PRIVATE_KEY` at the end:~

```
 yarn script:fork ./scripts/2022-06-12-ERC4626Factory.s.sol:ERC4626FactoryScript --private-key PRIVATE_KEY`
 ```

To run on fork from mainnet:
```
yarn script:fork ./scripts/2022-06-12-ERC4626Factory.s.sol
```

The `AddressGuru` contract is not meant to be deployed and might be an initial step towards our central repo for addresses though  it could be done better (this was just a quick way).

## Implementer Checklist

### CODE REVIEW

File: pkg/core/scripts/2022-06-12-erc4626-factory/ERC4626Factory.s.sol [hunk 1]
```diff
@@ -0,0 +1,163 @@
...
+
+        console.log("-------------------------------------------------------");
+        console.log("Deploy factory");
+        console.log("-------------------------------------------------------");
+
+        // TODO: read factoryParms from input.json file (can't do this right now because of a foundry bug when parsing JSON)
+        BaseFactory.FactoryParams memory factoryParams = BaseFactory.FactoryParams({
+            stake: AddressBook.WETH,
+            oracle: AddressBook.SENSE_MASTER_PRICE_ORACLE,
+            ifee: 1000000000000000, // 0.1%
```
jparklev:
```suggestion
            ifee: 0.1e18, // 0.1%
```

pkg/core/scripts/AddressGuru.sol
```diff
@@ -0,0 +1,74 @@
...
+import { AddressBookGoerli } from "./AddressBookGoerli.sol";
+
+contract AddressGuru {
+    uint256 public chainId;
+    constructor(uint256 _chainId) {
+        chainId = _chainId;
+    }
+
+    function divider() public returns (address) {
+        return chainId == AddressBook.MAINNET ? AddressBook.DIVIDER_1_2_0 : AddressBookGoerli.DIVIDER;
```
jparklev: can also use the `chainId` opcode! https://ethereum.stackexchange.com/questions/56749/retrieve-chain-id-of-the-executing-chain-from-a-solidity-contract

fedealconada: nice!

abylaw (approved):
Ok, nice! Primarily curious, why move over to foundry scripts for deploy? is it in effort to reduce the number of tools we rely on? (foundry vs foundry & hardhat?)
Is there any particular advantage (aside from using one language) to using solidity as our scripting language vs js?

fedealconada:
> Ok, nice! Primarily curious, why move over to foundry scripts for deploy? is it in effort to reduce the number of tools we rely on? (foundry vs foundry & hardhat?) Is there any particular advantage (aside from using one language) to using solidity as our scripting language vs js?

- is it in effort to reduce the number of tools we rely on? yes
- Is there any particular advantage? yes
  - aside from using one language
  - foundry is faster
  - verifying seems to run smoother
  - txs get executed faster (this is just because foundry is more generous with gas though we could achieve the same with hardhat though)
  - no need for using hardhat pre-processor when having 2 contracts with same name or using Fully Qualified Names (e.g [this](https://github.com/sense-finance/sense-v1/pull/327/files#diff-99c497fec42b358354cb6dfc14778b160dd066656c38acf01ebc60cf5dd302c3R53)) for the contracts when deploying  
  - we won't need to add contracts on the [`index.sol`](https://github.com/sense-finance/sense-v1/blob/dev/pkg/deployments/src/index.sol) contract and we could even deprecate the whole `deployments` package

PR #330, opened 2022-12-14
GEN-1773: erc4626 review

fedealconada:
## Motivation

Whenever we onboard a new ERC4626 target, we want to make sure that this new target passes some basic tests.
This PR is part of the [GEN-1773](https://linear.app/sense-finance/issue/GEN-1773/create-an-informal-smart-contract-review-process-for-new-targets) which establishes a [review process for new targets](https://www.notion.so/sensefinance/Review-process-for-4626-targets-88574f5382314ce7aef901cbae047c37). 

## Solution

Inspired by a16z [generalised property tests for ERC4626 vaults](https://github.com/a16z/erc4626-tests) we use those tests and also the ones we've had built on `ERC4626Adapter.tm.sol` (which have been slightly modified) as the test suite for new targets.

## Implementer Checklist


pkg/core/src/tests/adapters/ERC4626Adapter.tm.sol
```diff
@@ -11,88 +11,218 @@ import { BaseAdapter } from "../../adapters/abstract/BaseAdapter.sol";
...
 
-    function testMainnetERC4626AdapterScale() public {
-        ERC4626 target = ERC4626(AddressBook.IMUSD);
+    function testMainnetScale() public {
         uint256 decimals = target.decimals();
         uint256 scale = target.convertToAssets(10**decimals) * 10**(18 - decimals);
         assertEq(erc4626Adapter.scale(), scale);
     }
 
+    // function testScale() public {
```
jparklev: whats the reason not to have this test?

jparklev (approved):
Nice, see the build is failing but otherwise this works IMO! Cool to see the a16z tests, maybe some patterns we can pick up (`vm.assume(_isEOA(user))`?)

pkg/core/src/tests/adapters/ERC4626Adapter.tm.sol
fedealconada: the reason was that it's hard to generalise a test because making the scale to change usually depends on the `totalAssets()` and adding underlying directly to the vault does not always trigger a change in the scale.
I've made a small modifications to have a generalised way (which is just transferring underlying to the vault) which may fail and, otherwise, custom implementations for the vault we are trying to test.

pkg/core/src/tests/adapters/ERC4626Adapter.tm.sol
```diff
@@ -11,88 +11,205 @@ import { BaseAdapter } from "../../adapters/abstract/BaseAdapter.sol";
...
-    MasterPriceOracle public masterOracle;
-
-    uint64 public constant ISSUANCE_FEE = 0.01e18;
-    uint256 public constant STAKE_SIZE = 1e18;
-    uint256 public constant MIN_MATURITY = 2 weeks;
-    uint256 public constant MAX_MATURITY = 14 weeks;
-    uint8 public constant MODE = 0;
-    uint16 public constant DEFAULT_LEVEL = 31;
-    uint256 public constant INITIAL_BALANCE = 1.25e18;
+    address public userWithAssets; // address of a userr with underlying balance
```
abylaw:
```suggestion
    address public userWithAssets; // address of a user with underlying balance
```

pkg/core/src/tests/adapters/ERC4626Adapter.tm.sol
```diff
@@ -105,50 +222,45 @@ contract ERC4626Adapters is Test {
...
 
-    function testMainnetUnwrapTarget() public {
-        uint256 uBalanceBefore = ERC20(AddressBook.MUSD).balanceOf(address(this));
-        uint256 tBalanceBefore = ERC20(AddressBook.IMUSD).balanceOf(address(this));
+    // utils
 
-        ERC20(AddressBook.IMUSD).approve(address(erc4626Adapter), tBalanceBefore);
-        uint256 uDecimals = ERC20(AddressBook.MUSD).decimals();
-        uint256 rate = ERC4626(AddressBook.IMUSD).convertToAssets(10**uDecimals);
+    function _mutateVault(uint256 amt) internal {
```
abylaw: hmmm! i find `mutate` a bit unintuitive. I understand you probably chose it because something like `increaseVaultBalance` isn't strictly correct...kind of wordy, but maybe `simulateVaultRewardsIncrease`? `increaseVaultYield`?

pkg/core/src/tests/adapters/ERC4626Adapter.tm.sol
```diff
@@ -105,50 +222,45 @@ contract ERC4626Adapters is Test {
...
-        uint256 uBalanceAfter = ERC20(AddressBook.MUSD).balanceOf(address(this));
+        if (address(target) == AddressBook.BB_wstETH4626) {
+            underlying.transfer(AddressBook.IDLE_CDO, amt);
+            return;
+        }
 
-        assertEq(tBalanceAfter, 0);
-        assertEq(uBalanceBefore + unwrapped, uBalanceAfter);
+        // try mutating vault by transfering underlying to the vault
+        underlying.transfer(address(target), amt);
```
abylaw, line 251: nit: slightly prefer an `if else` here  since you can ignore irrelevant branches & control flow is more obvious :)

abylaw (approved):
LGTM~

PR #331, opened 2023-01-11
GEN-1731: specify receiver

fedealconada:
## Motivation

We want the Periphery methods to be able to specify who the receiver of the assets will be and not to assume it is always the msg.sender. This can help integrators and it makes the Periphery more flexible.

[Full spec on for Periphery v2 here
](https://www.notion.so/sensefinance/22-12-19-Sense-Periphery-v2-0-Spec-da948b88bb45467aab543e9d177168b5)
## Implementer Checklist


pkg/core/src/Periphery.sol
```diff
@@ -790,16 +827,17 @@ contract Periphery is Trust, IERC3156FlashBorrower {
             userData: abi.encode(liq.amounts, liq.minBptOut),
             fromInternalBalance: false
         });
-        balancerVault.joinPool(poolId, address(this), msg.sender, request);
-        lpBal = ERC20(address(pool)).balanceOf(msg.sender);
+        balancerVault.joinPool(poolId, address(this), receiver, request);
+        lpBal = ERC20(address(pool)).balanceOf(receiver);
```
jparklev, line 854: Hmm it feels a little odd to me now looking at this that we take the lp balance of the user and return it to them. As far as I can tell, this is just a return value and isn't used for anything, so it's not a big problem, just seems slightly unexpected to me

pkg/core/src/Periphery.sol
```diff
@@ -790,16 +827,17 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
     }
 
     function _removeLiquidityFromSpace(
         bytes32 poolId,
         address pt,
         address target,
         uint256[] memory minAmountsOut,
-        uint256 lpBal
+        uint256 lpBal,
+        address receiver
```
jparklev: Since this is an internal function, I don't think we need to use the `receiver` pattern if it isn't necessary. Tripped me up for a sec that we were calling this function from `_removeLiquidity` with `this` as the receiver address

pkg/core/src/Periphery.sol
```diff
@@ -335,17 +358,20 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
         uint256[] memory minAmountsOut,
         uint256 minAccepted,
-        bool intoTarget
+        bool intoTarget,
+        address receiver
     ) external returns (uint256 tBal, uint256 ptBal) {
         (tBal, ptBal) = _removeLiquidity(adapter, maturity, lpBal, minAmountsOut, minAccepted, intoTarget);
-        ERC20(Adapter(adapter).target()).safeTransfer(msg.sender, tBal); // Send Target back to the User
+        ERC20(Adapter(adapter).target()).safeTransfer(receiver, tBal); // Send Target to the receiver
+        if (ptBal > 0) ERC20(divider.pt(adapter, maturity)).safeTransfer(receiver, ptBal); // Send PTs to the receiver
```
jparklev: What's the reason to move this out of `_removeLiquidity`?

pkg/core/src/Periphery.sol
```diff
@@ -488,7 +518,7 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         BalancerVault.FundManagement memory funds = BalancerVault.FundManagement({
             sender: address(this),
             fromInternalBalance: false,
-            recipient: payable(address(this)),
+            recipient: payable(receiver),
```
jparklev: Think we can put the `payable` modifier on the function param

jparklev (approved):
Had a q or two and a concern about doing `balanceOf(receiver)`, but otherwise LGTM (once the build is going again 🙂 )

pkg/core/src/Periphery.sol
abylaw: ideally this would be the number of minted lp shares - could we do a diff between before/after joining the pool and return that?

pkg/core/src/Periphery.sol
```diff
@@ -810,10 +848,10 @@ contract Periphery is Trust, IERC3156FlashBorrower {
             userData: abi.encode(lpBal),
             toInternalBalance: false
         });
-        balancerVault.exitPool(poolId, address(this), payable(address(this)), request);
+        balancerVault.exitPool(poolId, address(this), payable(receiver), request);
 
-        tBal = ERC20(target).balanceOf(address(this));
-        ptBal = ERC20(pt).balanceOf(address(this));
+        tBal = ERC20(target).balanceOf(receiver);
+        ptBal = ERC20(pt).balanceOf(receiver);
```
abylaw: same here: could we do a diff of `tBal`/`ptBal` and return the exact number of `tBal` & `ptBal` the receiver will get?

in fact, i believe this is incorrect...it might only work if the receiver has no existing target/pts.

pkg/core/src/Periphery.sol
```diff
@@ -564,13 +594,14 @@ contract Periphery is Trust, IERC3156FlashBorrower {
 
     /// @return tAmount if mode = 0, target received from selling YTs, otherwise, returns 0
     /// @return issued returns amount of YTs issued (and received) except first provision which returns 0
-    /// @return lpShares Space LP shares received given the liquidity added
+    /// @return lpShares Space LP shares received given the liquidity added TODO: this is WRONG. It is the current LP balance of the receiver, right? What do we want here?
```
abylaw: would prefer the  `Space LP shares received given the liquidity` please.

Maybe a reword too: `LP Shares received from adding liquidity to a Space.`

pkg/core/src/Periphery.sol
```diff
@@ -198,80 +204,93 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
         uBal = Adapter(adapter).unwrapTarget(tBal); // unwrap target into underlying
-        ERC20(Adapter(adapter).underlying()).safeTransfer(msg.sender, uBal); // transfer underlying to msg.sender
+        ERC20(Adapter(adapter).underlying()).safeTransfer(receiver, uBal); // transfer underlying to receiver
     }
 
     /// @notice Swap YT for Target of a particular series
     /// @param adapter Adapter address for the Series
     /// @param maturity Maturity date for the Series
     /// @param ytBal Balance of Yield Tokens to swap
+    /// @param receiver Address to receive the Target
```
abylaw, line 268: consistency: add docs for return values?

pkg/core/src/tests/Periphery.t.sol
```diff
@@ -722,12 +722,166 @@ contract PeripheryTest is TestHelper {
...
+        // unwrap target into underlying
+        uint256 uBal = tBal.fmul(adapter.scale());
+
+        // calculate underlying swapped to pt
+        uint256 ptBal = uBal.fdiv(balancerVault.EXCHANGE_RATE());
+
+        vm.expectEmit(true, false, false, false);
+        emit Swapped(address(this), "0", adapter.target(), address(0), 0, 0, msg.sig);
+
+        periphery.swapTargetForPTs(address(adapter), maturity, tBal, 0, bob);
```
abylaw: i wonder if we need separate tests for these periphery functions with different receivers since there aren't separate code paths in the functions we're testing? i know you wrote a lot of new tests here...but i'm worried this is quite a bit of code while not increasing the coverage significantly. thoughts?

could we somehow do a limited fuzz test, for example, as an alternate solution?

pkg/core/src/Periphery.sol
fedealconada: right! i agree

pkg/core/src/Periphery.sol
fedealconada: I initially removed it more like conceptually because of what the function is doing. This function has no transfers to the receiver except for this transfer of the PTs. So, now with the `receiver` feature, this function would need to receive that param. So I felt it could be moved out.  
Then, I thought this may lead to stack too deep errors... But I can move it back as it was and then, when adding the next PRs check if they are giving me troubles (specially with the permit2 which adds more params).

pkg/core/src/Periphery.sol
fedealconada: like it!

pkg/core/src/Periphery.sol
fedealconada: yes, I think there was a reasoning why we did this like that (maybe @jparklev remembers). But I'm down to changing all this return values to exactly what is being returned and not the balance.

pkg/core/src/tests/Periphery.t.sol
fedealconada: completely agree! i've modified them to do fuzzing for some but, unfortunately, for others i'm reaching the stack too deep error, tried to sort it out but couldn't (and don't want to spend too much time). In any case, for the ones that I couldn't I left the ones that test a different receiver and removed the ones where the from and the receiver are the same, because as you said, they were not increasing coverage :)

PR #333, opened 2023-01-11
GEN-1733: combine to underlying

fedealconada:
## Motivation

Add more flexibility to the Periphery by having the option both to combine PT and YTs to target (passthrough method to `divider.combine()` and to combine them to underlying (`divider.combine` + `adapter.unwrapTarget()`)

[Full spec here ](https://www.notion.so/sensefinance/22-12-19-Sense-Periphery-v2-0-Spec-da948b88bb45467aab543e9d177168b5)

## Implementer Checklist

### DIFF
--- a/pkg/core/src/Periphery.sol
+++ b/pkg/core/src/Periphery.sol
@@ -445,7 +445,7 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         (tAmount, issued, lpShares) = _addLiquidity(dstAdapter, dstMaturity, tBal, mode, minBptOut, msg.sender);
     }
 
-    /* ========== ISSUANCE UTILS ========== */
+    /* ========== UTILS ========== */
 
     /// @notice Mint PTs & YTs of a specific Series
     /// @param adapter Adapter address for the Series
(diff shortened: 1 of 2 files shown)


pkg/core/src/Periphery.sol
```diff
@@ -477,6 +477,43 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         ERC20(divider.yt(adapter, maturity)).safeTransfer(receiver, uBal); // Send YT to the receiver
     }
 
+    /// @notice Reconstitute Target by burning PT and YT
+    /// @dev Explicitly burns YTs before maturity, and implicitly does it at/after maturity through `_collect()`
```
jparklev: I think this comment would be more for explaining divider internals rather than on the periphery?

pkg/core/src/Periphery.sol
```diff
@@ -477,6 +477,43 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
+    /// @param adapter Adapter address for the Series
+    /// @param maturity Maturity date for the Series
+    /// @param amt Amount of PT and YT to burn
+    /// @param receiver Address where the resulting Target will be transferred to
```
```suggestion
    /// @param receiver Address where the resulting Target will be transferred
```

pkg/core/src/Periphery.sol
```diff
@@ -477,6 +477,43 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
+    function combine(
+        address adapter,
+        uint256 maturity,
+        uint256 amt,
```
jparklev: I suppose `uBal` would be if we wanted to stay consistent with the divider, right?

pkg/core/src/Periphery.sol
```diff
@@ -477,6 +477,43 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
+        address receiver
+    ) external returns (uint256 tBal) {
+        ERC20(divider.pt(adapter, maturity)).safeTransferFrom(msg.sender, address(this), amt); // Pull PTs
+        ERC20(divider.yt(adapter, maturity)).safeTransferFrom(msg.sender, address(this), amt); // Pull YTs
+        uint256 tBal = divider.combine(adapter, maturity, amt);
+        ERC20(Adapter(adapter).target()).safeTransfer(receiver, tBal); // Send Target to the receiver
```
```suggestion
        ERC20(Adapter(adapter).target()).safeTransfer(receiver, tBal = divider.combine(adapter, maturity, amt)); // Send Target to the receiver
```

as an option

jparklev: if we don't do this option, we can still remove the `uint256` in front of `tBal`

pkg/core/src/Periphery.sol
```diff
@@ -477,6 +477,43 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
+    }
+
+    /// @notice Reconstitute Target by burning PT and YT and unwrap it
```
```suggestion
    /// @notice Reconstitute Target by burning PT and YT and unwraping it
```

pkg/core/src/Periphery.sol
```diff
@@ -477,6 +477,43 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
+    /// @param maturity Maturity date for the Series
+    /// @param amt Amount of PT and YT to burn
+    /// @param receiver Address where the resulting Underlying will be transferred to
+    function combineToUnderlying(
+        address adapter,
+        uint256 maturity,
+        uint256 amt,
+        address receiver
+    ) external returns (uint256 uBal) {
+        ERC20(divider.pt(adapter, maturity)).safeTransferFrom(msg.sender, address(this), amt); // Pull PTs
```
jparklev: Similar to the issuance PR, might not need the `safeTransferFrom` here

jparklev (commented):
Same q about generalized zaps as in the issue PR, but overall LG ~

jparklev:
also do we know why coverage took so long?
Screen Shot 2023-01-12 at 12 11 02 PM

pkg/core/src/Periphery.sol
fedealconada: yep

pkg/core/src/Periphery.sol
fedealconada: yes, true, you prefer `uBal` instead of `amt`?

pkg/core/src/Periphery.sol
fedealconada: i like the option!

fedealconada:
> ERC20(divider.pt(adapter, maturity)).safeTransferFrom(msg.sender, address(this), amt); // Pull PTs

no clue, it's happening on all these last PRs, have to investigate

pkg/core/src/tests/Periphery.t.sol
```diff
@@ -2260,6 +2261,49 @@ contract PeripheryTest is TestHelper {
...
+        uint256 maturity = getValidMaturity(2021, 10);
+        (address pt, address yt) = periphery.sponsorSeries(address(adapter), maturity, true);
+        uint256 uBal = periphery.issue(address(adapter), maturity, tBal, alice);
+
+        uint256 ptBalBefore = ERC20(pt).balanceOf(alice);
+        uint256 ytBalBefore = ERC20(yt).balanceOf(alice);
+        uint256 uBalBefore = ERC20(adapter.underlying()).balanceOf(alice);
+
+        ERC20(pt).approve(address(periphery), ptBalBefore);
+        ERC20(yt).approve(address(periphery), ytBalBefore);
```
abylaw, line 1559: this double combine might not be great ux on the consumer-side - are there any clever tricks we can do to reduce the number of approves?

abylaw (approved):
LGTM!

pkg/core/src/tests/Periphery.t.sol
jparklev: Yea, this is actually a drawback of doing this via the Periphery. Since the Divider is the owner of these tokens, it can pull them from the user w.o. approvals, but with the Periphery, i'm not sure if there's anyway around it (short of permit-style stuff!). Right @fedealconada?

pkg/core/src/Periphery.sol
jparklev: No strong preference, but perhaps `uBal` for consistency?

PR #334, opened 2023-01-11
GEN-1732: auto redeem

fedealconada:
## Motivation

Using `swapPTsForTarget` or `swapPTsForUnderlying` should automatically try redeeming PTs on the Divider instead of executing the `_swap` method when a Series has already been settled (since PTs would be 1:1 with the underlying). 

[Full spec here](https://www.notion.so/sensefinance/22-12-19-Sense-Periphery-v2-0-Spec-da948b88bb45467aab543e9d177168b5#8281f494e53a490da7e8cd85beb06bdb)

## Implementer Checklist

### DIFF
--- a/pkg/core/src/tests/Periphery.t.sol
+++ b/pkg/core/src/tests/Periphery.t.sol
@@ -693,8 +693,10 @@ contract PeripheryTest is TestHelper {
         vm.prank(from);
         ERC20(pt).approve(address(periphery), ptBalBefore);
 
-        vm.expectEmit(true, false, false, false);
-        emit PTRedeemed(address(adapter), maturity, 0);
+        (, , , , , , , uint256 mscale, ) = divider.series(address(adapter), maturity);
+        uint256 tBalRedeemed = ptBalBefore.fdiv(mscale);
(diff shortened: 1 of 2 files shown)


pkg/core/src/tests/Periphery.t.sol
```diff
@@ -662,6 +662,35 @@ contract PeripheryTest is TestHelper {
...
+        // settle series
+        vm.warp(maturity);
+        divider.settleSeries(address(adapter), maturity);
+
+        uint256 tBalBefore = ERC20(adapter.target()).balanceOf(alice);
+        uint256 ptBalBefore = ERC20(pt).balanceOf(alice);
+
+        ERC20(pt).approve(address(periphery), ptBalBefore);
+
+        vm.expectEmit(true, false, false, false);
```
jparklev:
```suggestion
        vm.expectEmit(true, true, false, false);
```
and possibly we could pre-calculate and check the redemption amount emitted as well?

jparklev (approved):
Nice, simple enough

abylaw (approved):
nice!! makes sense to me

PR #335, opened 2023-01-23
GEN-1735: permit2

fedealconada:
## Motivation

Improve Sense’s UX by using Permit2 which would allow users not to have to make `approve()` calls in order for the Periphery to spend their tokens.

## Solution

Methods on our `Periphery` contract making use of a `transferFrom` (or `safeTransferFrom`) will need to be modified to include two additional params:
    - `IPermit2.PermitTransferFrom calldata permit`
    - `bytes memory signature`

Because of constraints (aka `stack too deep` error), I have decided that instead of receiving these 2 params, it receives just one (which is the bytes encoded version of the 2 mentioned above) and then we decode it on the method. 

## Implementer Checklist


pkg/core/script/2022-06-12-ERC4626Factory.s.sol
```diff
@@ -5,6 +5,7 @@ import "forge-std/Script.sol";
 import "forge-std/StdCheats.sol";
 
 import { AddressGuru, Constants } from "@sense-finance/v1-utils/addresses/AddressGuru.sol";
```
jparklev, line 7: We could get this from https://github.com/sense-finance/sense-addresses when we're all merged in there, right? :D

pkg/core/src/external/IPermit2.sol
```diff
@@ -0,0 +1,57 @@
+// Minimal Permit2 interface, derived from
+// https://github.com/Uniswap/permit2/blob/main/src/interfaces/IPermit2.sol
```
jparklev: Dead link? Maybe `interfaces/ISignatureTransfer.sol`?

pkg/fuse/src/tests/NoopPoolManager.tm.sol
```diff
@@ -42,6 +43,10 @@ contract NoopPoolManagerTest is ForkTest {
 
     BaseAdapter.AdapterParams internal adapterParams;
 
+    // user
+    uint256 internal bobPrivKey = _randomUint256();
+    address internal bob = vm.addr(bobPrivKey);
```
jparklev, line 48: Nice, haven't seen this cheatcode before

pkg/fuse/src/tests/NoopPoolManager.tm.sol
```diff
@@ -116,8 +122,12 @@ contract NoopPoolManagerTest is ForkTest {
         mockAdapter.scale();
         divider.setAdapter(address(mockAdapter), true);
 
+        setPermit2(AddressBook.PERMIT2);
         stake.mint(address(this), 2e18);
-        stake.approve(address(periphery), 2e18);
+        stake.mint(bob, 2e18);
+        stake.approve(address(permit2), 2e18);
```
jparklev, line 128:
```suggestion
```

pkg/core/src/tests/test-helpers/mocks/MockSpace.sol
```diff
@@ -259,8 +267,8 @@ contract MockSpaceFactory {
 
         pool = new MockSpacePool(address(vault), _target, pt, _adapter);
         pools[_adapter][_maturity] = address(pool);
-
-        vault.setYieldSpace(address(pool));
+        poolsByPoolId[pool.getPoolId()] = address(pool);
+        // vault.setYieldSpace(address(pool));
```
jparklev:
```suggestion
```

pkg/core/src/Periphery.sol
```diff
@@ -90,12 +96,22 @@ contract Periphery is Trust, IERC3156FlashBorrower {
     function sponsorSeries(
         address adapter,
         uint256 maturity,
-        bool withPool
+        bool withPool,
+        bytes memory permit
```
jparklev: What's the reason to decode bytes rather than do a struct arg like:
```
struct Permit {
   IPermit2.PermitTransferFrom permitMsg;
   bytes signature;
}
```
gas savings, perhaps?

pkg/core/src/Periphery.sol
```diff
@@ -149,9 +165,20 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         uint256 maturity,
         uint256 tBal,
         uint256 minAccepted,
-        address receiver
+        address receiver,
+        bytes memory permit
     ) external returns (uint256 ptBal) {
-        ERC20(Adapter(adapter).target()).safeTransferFrom(msg.sender, address(this), tBal); // pull target
+        // pull underlying
```
jparklev, line 177: target, right?

jparklev: This gets me thinking, should we verify that `pmsg.permitted.token` is what we expect? I guess the function would just fail b/c it would e.g try to make a swap with target it doesn't have, but i'm not sure if there's an edge case somewhere. WDYT?

pkg/core/src/Periphery.sol
```diff
@@ -768,19 +951,36 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
+                pmsg,
+                IPermit2.SignatureTransferDetails({ to: address(this), requestedAmount: lpBal }),
+                msg.sender,
+                signature
+            );
+        }
 
         // (1) Remove liquidity from Space
         {
+            // address target = Adapter(adapter).target();
```
```suggestion
```

pkg/core/src/Periphery.sol
```diff
@@ -706,7 +888,8 @@ contract Periphery is Trust, IERC3156FlashBorrower {
             // issue = 0 means that we are on the first pool provision or that the pt:target ratio is 0:target
             if (mode == 0) {
                 // (2) Sell YTs
-                tAmount = _swapYTsForTarget(address(this), adapter, maturity, issued);
+                tAmount = _swapYTsForTarget(address(this), adapter, maturity, issued, permit);
```
jparklev: Hmm is this an empty permit? Since I think this call won't trigger `sender != address(this)` and the following transfer

pkg/core/src/tests/test-helpers/Permit2Helper.sol
```diff
@@ -0,0 +1,191 @@
...
+
+// Local bytecode clone of the canonical Permit2 contract deployed to mainnet.
+contract Permit2Clone is IPermit2 {
+    error InvalidNonce();
+    error InvalidSigner();
+
+    constructor() {
+        // Deployed Permit2 bytecode at
+        // https://etherscan.io/address/0x000000000022D473030F116dDEE9F6B43aC78BA3
+        bytes
```
jparklev, line 167: This is to avoid copy/pasting?

jparklev (commented):
Lots of new concepts and paths of work here – great job. Left a few comments with suggestions, questions, and anything potentially unexpected ~

Also, I see there's a `Permit2Library` [here](https://github.com/Uniswap/permit2/blob/ca6b6ff2b47afc2942f3c67b0d929ca4f0b32631/src/libraries/Permit2Lib.sol), would it be useful at all to us?

Also also, I dug into the universal router for abit reviewing this, since I was curious about what they're doing. Looks like they have this command multicall pattern where you queue up batch actions and the contract executes them in sequence. And they use it to queue up a permit transfer from the user and another call (e.g. swap) in the same TX. I think we could do something similar, but it might require a bigger re-architecting... don't think it's worth it, but lets' jam about it on our next call

pkg/core/script/2022-06-12-ERC4626Factory.s.sol
fedealconada: yess! that's the idea, I need to go through each repo we have and create PRs to integrate sense-addresses

pkg/core/src/Periphery.sol
fedealconada: hmm... didn't really think about the struct to be honest, i will try to see if it works with the struct, it would be more readable

pkg/core/src/Periphery.sol
fedealconada: I thought about it but then just went with the idea of letting it fail 🤷🏼‍♂️

pkg/core/src/Periphery.sol
fedealconada: Ohhh, good catch, yeah. Nothing happens though, but I should call it with "0x" to be correct.

pkg/core/src/tests/test-helpers/Permit2Helper.sol
fedealconada: I think so! I took it from [here](https://github.com/dragonfly-xyz/useful-solidity-patterns/blob/main/test/Permit2Vault.t.sol#L249)

fedealconada:
> Lots of new concepts and paths of work here – great job. Left a few comments with suggestions, questions, and anything potentially unexpected ~
> 
> Also, I see there's a `Permit2Library` [here](https://github.com/Uniswap/permit2/blob/ca6b6ff2b47afc2942f3c67b0d929ca4f0b32631/src/libraries/Permit2Lib.sol), would it be useful at all to us?
> 

I'll check! Ty!

> Also also, I dug into the universal router for abit reviewing this, since I was curious about what they're doing. Looks like they have this command multicall pattern where you queue up batch actions and the contract executes them in sequence. And they use it to queue up a permit transfer from the user and another call (e.g. swap) in the same TX. I think we could do something similar, but it might require a bigger re-architecting... don't think it's worth it, but lets' jam about it on our next call

Right! I've seen that as well, but not sure how it would help us. Happy to jam about it :)

pkg/core/src/tests/test-helpers/Permit2Helper.sol
```diff
@@ -0,0 +1,192 @@
...
+    function generatePermit(
+        uint256 ownerPrivKey,
+        address spender,
+        address token
+    ) public returns (Periphery.PermitData memory permit) {
+        IPermit2.PermitTransferFrom memory permit = IPermit2.PermitTransferFrom({
+            permitted: IPermit2.TokenPermissions({ token: ERC20(token), amount: type(uint256).max }),
+            nonce: _randomUint256(),
+            deadline: block.timestamp
+        });
```
abylaw, line 49: from a real consumer, how would we get a nonce and set the deadline?
here we can control when a block is mined so setting the deadline as the block.timestamp makes sense - would we make a reasonable estimate as to when the tx will be mined, and set that? would we need a nonce that is unique per spender?

pkg/core/src/Periphery.sol
```diff
@@ -213,13 +255,22 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
+            msg.sender,
+            permit.sig
+        );
         // Wrap Underlying into Target and swap it for YTs
-        uint256 targetIn = Adapter(adapter).wrapUnderlying(underlyingIn);
-        (targetBal, ytBal) = _flashBorrowAndSwapToYTs(adapter, maturity, targetIn, targetToBorrow, minOut);
+        {
+            uint256 targetIn = Adapter(adapter).wrapUnderlying(underlyingIn);
+            (targetBal, ytBal) = _flashBorrowAndSwapToYTs(adapter, maturity, targetIn, targetToBorrow, minOut);
+        }
```
abylaw, line 272: for my own education: curious why you're wrapping these two lines in a new scope (if that's what `{}` is in solidity)

abylaw:
just a few questions~i think this is a little out of my depth when it comes to making coding design decisions :)

jparklev:
gentle bump on the qs here ^ @fedealconada ~

pkg/core/src/Periphery.sol
fedealconada: when you see some random `{}` it's because we were getting the "[stack too deep](https://soliditydeveloper.com/stacktoodeep)" error, and one of the ways to avoid this is by using "block scoping".

pkg/core/src/tests/test-helpers/Permit2Helper.sol
fedealconada: > how would we get a nonce and set the deadline?

we can use sequential nonces

> set the deadline?
> would we make a reasonable estimate as to when the tx will be mined, and set that?

yes! we could be generous to avoid the tx to revert

> would we need a nonce that is unique per spender?
not per `sender` but per token owner

fedealconada:
> gentle bump on the qs here ^ @fedealconada ~

sorry, i missed those ones, ive just answered them :)

pkg/core/src/tests/test-helpers/Permit2Helper.sol
abylaw: tyty :)

fedealconada:
Closing since it was merged on `GEN-1888-zaps`

PR #336, opened 2023-01-24
prep for sanFRAX_EUR Wrapper integration

fedealconada:
## Motivation

sanFRAX_EUR Wrapper integration (Angle)

## Solution

- Create OwnableERC4626CropFactory (GEN-1898)
- Create OwnableERC4626CropAdapter (GEN-1898)
- Create PingPongClaimer (this is a contract used by the adapter to claim rewards. As in sanFRAX_EUR Wrapper rewards are automatically claimed when there's a token transfer, this claimer is just returning the target back (GEN-1899)
- Added Morpho as a dependency to use RewardsDistributor (GEN-1914)
- Modified ERC4626Adapter.tm.sol as part of the review of sanFRAX_EUR Wrapper ([GEN-1776](https://linear.app/sense-finance/issue/GEN-1776/review-angle-frax-product-before-integration)) 
- Create deployment script

## Implementer Checklist


pkg/core/src/tests/adapters/ERC4626Adapter.tm.sol
```diff
@@ -247,10 +273,23 @@ contract ERC4626Adapters is Test {
...
+
+            vm.warp(block.timestamp + 1 days);
+            return;
+        }
+
         // try mutating vault by transfering underlying to the vault
         underlying.transfer(address(target), amt);
     }
 
+    event Hola(uint256);
```
abylaw: hola!

pkg/core/src/adapters/implementations/claimers/PingPongClaimer.sol
```diff
@@ -0,0 +1,27 @@
...
+// External references
+import { ERC20 } from "solmate/tokens/ERC20.sol";
+
+// Internal references
+import { IClaimer } from "../../../adapters/abstract/IClaimer.sol";
+
+/// @title ExtractableReward
+/// @dev This claimer only returns the received target back to the adapter. This is because
+/// target has automatic rewards claiming which is triggered when transferring the target.
+contract PingPongClaimer is IClaimer {
```
abylaw, line 14: what a cute name. what's the reasoning behind it? :)
is the ping pong claimer specific to Angle? if so, should it have Angle in the name?

pkg/core/src/tests/adapters/ERC4626Adapter.tm.sol
```diff
@@ -247,10 +273,23 @@ contract ERC4626Adapters is Test {
             return;
         }
 
+        if (address(target) == AddressBook.sanFRAX_EUR_Wrapper) {
+            deal(AddressBook.FRAX, address(this), amt);
+            address sanToken = Token(address(target)).sanToken();
+            address stableMaster = Token(sanToken).stableMaster();
+            MockToken(AddressBook.FRAX).approve(stableMaster, amt);
+            Token(stableMaster).mint(amt, address(this), 0x6b4eE7352406707003bC6f6b96595FD35925af48, 0);
```
abylaw: nit: address -> constant?

pkg/core/src/adapters/abstract/factories/OwnableERC4626CropFactory.sol
```diff
@@ -0,0 +1,85 @@
...
+import { BaseAdapter } from "../../abstract/BaseAdapter.sol";
+import { OwnableERC4626CropAdapter } from "../erc4626/OwnableERC4626CropAdapter.sol";
+import { ERC4626CropFactory } from "./ERC4626CropFactory.sol";
+import { ExtractableReward } from "../../abstract/extensions/ExtractableReward.sol";
+import { Errors } from "@sense-finance/v1-utils/libs/Errors.sol";
+
+// External references
+import { Bytes32AddressLib } from "solmate/utils/Bytes32AddressLib.sol";
+
+/// @notice Ownable Crop Factoy contract that deploys Ownable Crop Adapters for Rolling Liquidity Vaults
```
abylaw:
```suggestion
/// @notice Ownable Crop Factory contract that deploys Ownable Crop Adapters for Rolling Liquidity Vaults
```

pkg/core/src/adapters/abstract/factories/OwnableERC4626CropFactory.sol
```diff
@@ -0,0 +1,85 @@
...
+            mode: factoryParams.mode,
+            tilt: factoryParams.tilt,
+            level: DEFAULT_LEVEL
+        });
+
+        // Use the CREATE2 opcode to deploy a new Adapter contract.
+        // This will revert if an ERC4626 adapter with the provided target has already
+        // been deployed, as the salt would be the same and we can't deploy with it twice.
+        adapter = address(
+            new OwnableERC4626CropAdapter{ salt: _target.fillLast12Bytes() }(
```
abylaw, line 57: wondering if we should be allowing the same erc4626 adapters but with varying `rewardRecipients` & `reward`s. (by including them in the salt?)

pkg/core/src/adapters/abstract/erc4626/OwnableERC4626CropAdapter.sol
```diff
@@ -0,0 +1,50 @@
...
+    function getMaturityBounds() external view override returns (uint256, uint256) {
+        return open == 2 ? (adapterParams.minm, adapterParams.maxm) : (0, 0);
+    }
+
+    function notify(
+        address _usr,
+        uint256 amt,
+        bool join
+    ) public override(BaseAdapter, Crop) {
+        super.notify(_usr, amt, join);
```
abylaw: for my own understanding: if we are calling `super.notify` with the same arguments, wondering why we need to override?

pkg/core/script/2023-01-24-OwnableERC4626CropFactory.s.sol
```diff
@@ -0,0 +1,223 @@
...
+            console.log("- Deploy sanFRAX_EUR_Wrapper rewards claimer");
+            vm.broadcast(deployer);
+            PingPongClaimer claimer = new PingPongClaimer(address(adapter));
+
+            console.log("- Set claimer to adapter");
+            vm.broadcast(senseMultisig); // broadcast following tx from multisig
+            adapter.setClaimer(address(claimer));
+
+            // create RLV
+            if (chainId != Constants.MAINNET) {
```
abylaw: unnecessary check since we're already in an `if (chainId != Constants.MAINNET)`

abylaw (approved):
looks good to me!!

as an FYI, UI won't support crop adapters, since it doesn't accept `data` (i'll add it tomorrow), and claimers will have to be manually added afterwards.

pkg/core/src/adapters/implementations/claimers/PingPongClaimer.sol
fedealconada: > what a cute name. what's the reasoning behind it? :)

credit to @jparklev for this name :) so, sanFRAX_EUR claims rewards automatically when transferring the token to someone (it does it via a hook). So, in order to claim the rewards on the adapter, we only need to do a transfer call. We've actually evaluated not having a claimer contract and just making the adapter to transfer the tokens to himself (which will triggers the claim) but we've decided not to do that since that would mean modifying `Crop.sol` and all adapters would be doing this transfer (even if that does not trigger the claim).
So, we've done this claimer which only returns the transferred tokens, this is why it's a Ping Pong: adapter transfers tokens to ping pong claimer and claimer returns it back.
   
> is the ping pong claimer specific to Angle? if so, should it have Angle in the name?

it is not. potentially, there may be other tokens with this same hook idea, so we could re-use it.

pkg/core/src/adapters/abstract/erc4626/OwnableERC4626CropAdapter.sol
fedealconada: this is because `BaseAdapter` defines this function as `virtual` and expects a contract to override it. `ERC4626Adapter` inherits `BaseAdapter` but does not override the function. So, this contract (`OwnableERC4626CropAdapter`) has to.

pkg/core/src/tests/adapters/CropsAdapter.t.sol
```diff
@@ -1402,6 +1403,19 @@ contract CropsAdapters is TestHelper {
...
+        uint256 tBal = 100e18;
+        uint256 maturity = getValidMaturity(2021, 10);
+        (, address yt) = periphery.sponsorSeries(address(cropsAdapter), maturity, true);
+
+        PingPongClaimer claimer = new PingPongClaimer();
+        vm.prank(Constants.RESTRICTED_ADMIN);
+        cropsAdapter.setClaimer(address(claimer));
+
+        // calling issue does not revert
+        divider.issue(address(cropsAdapter), maturity, tBal);
```
jparklev, line 1422: Thoughts on adding a check that the target transfer events are being emitted (to and from the claimer)?

pkg/core/src/adapters/abstract/IClaimer.sol
```diff
@@ -3,5 +3,5 @@ pragma solidity 0.8.15;
 
 interface IClaimer {
     /// @dev Claims rewards on protocol.
-    function claim() external;
+    function claim(address adapter) external;
```
jparklev, line 6: So this means that we can't use claimers that conform to the new interface with old crop(s) adapters, right?

pkg/core/src/adapters/implementations/claimers/PingPongClaimer.sol
```diff
@@ -0,0 +1,19 @@
...
+// Internal references
+import { IClaimer } from "../../../adapters/abstract/IClaimer.sol";
+import { BaseAdapter } from "../../../adapters/abstract/BaseAdapter.sol";
+
+/// @title ExtractableReward
+/// @dev This claimer only returns the received target back to the adapter. This is because
+/// target has automatic rewards claiming which is triggered when transferring the target.
+contract PingPongClaimer is IClaimer {
+    function claim(address adapter) external virtual {
+        ERC20 target = ERC20(BaseAdapter(adapter).target());
```
jparklev: Could also do like `ERC20(BaseAdapter(msg.sender).target())`

pkg/core/src/adapters/abstract/erc4626/OwnableERC4626CropAdapter.sol
```diff
@@ -0,0 +1,50 @@
...
+    uint256 internal open = 1;
+
+    constructor(
+        address _divider,
+        address _target,
+        address _rewardsRecipient,
+        uint128 _ifee,
+        AdapterParams memory _adapterParams,
+        address _reward
+    ) ERC4626Adapter(_divider, _target, _rewardsRecipient, _ifee, _adapterParams) Crop(_divider, _reward) {}
```
jparklev: What's the reason we don't inherit `ERC4626CropAdapter`? (which has `notify` and `_isValid`)

pkg/core/foundry.toml
```diff
@@ -14,5 +14,7 @@ remappings = [
     '@openzeppelin/contracts=../../node_modules/@openzeppelin/contracts',
     '@yield-daddy=lib/yield-daddy',
     '@auto-roller=lib/auto-roller',
-    'sense-v1-utils=../utils/src/'
+    'sense-v1-utils=../utils/src/',
+    '@morpho=lib/morpho-core-v1',
+    '@rari-capital/solmate/=../../node_modules/solmate'
```
jparklev, line 19: We'd moved to the `transmissions11/` solmate, right? Was there a legacy import relying on rari?

pkg/core/script/2023-01-24-OwnableERC4626CropFactory.s.sol
```diff
@@ -0,0 +1,221 @@
...
+        console.log("Deploying from:", deployer);
+        
+        address senseMultisig = chainId == Constants.GOERLI ? deployer : addressGuru.multisig();
+        console.log("Sense multisig is:", senseMultisig);
+
+        console.log("-------------------------------------------------------");
+        console.log("Deploy Ownable ERC4626 Crop Factory");
+        console.log("-------------------------------------------------------");
+
+        // TODO: read factoryParms from input.json file (can't do this right now because of a foundry bug when parsing JSON)
```
jparklev, line 53: 😞

pkg/core/script/2023-01-24-OwnableERC4626CropFactory.s.sol
```diff
@@ -0,0 +1,221 @@
...
+        BaseFactory.FactoryParams memory factoryParams = BaseFactory.FactoryParams({
+            oracle: addressGuru.oracle(),
+            stake: addressGuru.weth(),
+            stakeSize: 25e16, // 0.25 WETH
```
jparklev:
```suggestion
            stakeSize: 0.25e18, // 0.25 WETH
```

pkg/core/script/2023-01-24-OwnableERC4626CropFactory.s.sol
```diff
@@ -0,0 +1,221 @@
...
+        console.log("- Deploy sanFRAX_EUR_Wrapper rewards claimer");
+        vm.broadcast(deployer);
+        PingPongClaimer claimer = new PingPongClaimer();
+
+        // deploy the rewards distributor
+        RewardsDistributor distributor = _deployRewardsDistributor(reward);
+
+        if (chainId != Constants.MAINNET) {
+            // TODO: read targets from input.json file (can't do this right now because of a foundry bug when parsing JSON)
+            address sanFRAX_EUR_Wrapper = chainId == Constants.FORK || chainId == Constants.MAINNET ? AddressBook.sanFRAX_EUR_Wrapper : AddressBookGoerli.cUSDC;
```
jparklev: I think `chainId == Constants.MAINNET` is impossible based on the conditional above

pkg/core/script/2023-01-24-OwnableERC4626CropFactory.s.sol
```diff
@@ -0,0 +1,221 @@
...
+        console.log("  -> adapter guard: %s", guard);
+
+        vm.stopBroadcast();
+    }
+
+    function _createRLV(OwnableERC4626CropAdapter adapter, AutoRollerFactory rlvFactory, address distributor) internal returns (AutoRoller rlv) {
+        // create RLV for ownable adapter
+        console.log("- Create RLV for %s", OwnableERC4626CropAdapter(adapter).name());
+        
+        // TODO: replace address(0xFED) for rewards distributor!!!!!!!!!
```
jparklev: This has been done, right?

pkg/core/script/2023-01-24-OwnableERC4626CropFactory.s.sol
```diff
@@ -0,0 +1,221 @@
...
+
+            vm.stopBroadcast();
+        }
+
+        if (chainId == Constants.MAINNET) {
+            console.log("-------------------------------------------------------");
+            console.log("ACTIONS TO BE DONE ON DEFENDER: ");
+            console.log("1. Set factory on Periphery (multisig)");
+            console.log("2. Set factory as trusted on Divider (multisig)");
+            console.log("3. Deploy ownable adapter on Periphery using factory (relayer)");
```
jparklev, line 146: Why does the relayer need to do this piece, again? I might just be forgetting the significance

pkg/core/script/2023-01-24-OwnableERC4626CropFactory.s.sol
```diff
@@ -0,0 +1,221 @@
...
+            vm.broadcast(senseMultisig); // broadcast following tx from multisig
+            adapter.setClaimer(address(claimer));
+
+            // create RLV
+            AutoRoller rlv = _createRLV(adapter, AutoRollerFactory(rlvFactory), address(distributor));
+            // roll first series
+            _roll(rlv);
+        }
+
+        if (deployer != senseMultisig) {
```
jparklev, line 125: Will the deployer ever be the same as the multisig?

pkg/core/src/adapters/abstract/factories/OwnableERC4626CropFactory.sol
jparklev: good q

I think it's OK to leave `rewardRecipients` out since the `restrictedAdmin` can update it after the fact

For `reward`, hmm yea, it feels like a rare but valid use case. I think I like the property that "each factory can only deploy one adapter per target" and i'd be hesitant to do away with it

WDYT @fedealconada

jparklev (commented):
A few comments and questions around alternatives. Overall looks good and I like the script you made for this one – feels good to see some more complex stuff done with foundry's scripting

Also, should we cover these lines in the `Claimer`?
Screen Shot 2023-01-31 at 5 26 12 PM

pkg/core/src/adapters/abstract/factories/OwnableERC4626CropFactory.sol
fedealconada: yeah, definitely good question, made me think as well, we can also update the `reward` on `Crop.sol` and I think I'm good with keeping it as is.

pkg/core/src/adapters/implementations/claimers/PingPongClaimer.sol
fedealconada: ohhh, good one! and i can remove the `address adapter` param so i don't change the interface of this function!

pkg/core/src/adapters/abstract/erc4626/OwnableERC4626CropAdapter.sol
fedealconada: oh, yeah, no reason, just missed it! ty!

pkg/core/foundry.toml
fedealconada: oh, no no, it's just that we I think morpho is using that version so that's why we need to add the remapping

pkg/core/script/2023-01-24-OwnableERC4626CropFactory.s.sol
fedealconada: It's just what we were doing. But, actually, this time we are gonna use the Adapters UI to do this, so probably it will be the deployer. We don't really need the relayer to do this.

pkg/core/script/2023-01-24-OwnableERC4626CropFactory.s.sol
fedealconada: on Goerli it is

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
```diff
@@ -130,6 +130,15 @@ abstract contract BaseFactory is Trust {
         rewardsRecipient = _recipient;
     }
 
+    /// @notice sets trusted address for an adapter
+    function setIsTrusted(
```
fedealconada, line 134: What do you think about adding this @jparklev ? it's just convenient to have this method because when we deploy an adapters via factories, the owner is the factory. In this case, for example, I was missing a `setClaimer` function on the factory which would be a passthrough to the `setClaimer` of the adapter.
To avoid missing these kind of passthrough methods, we could just have this `setIsTrusted` which is triggered by the owner of the factory and allows it to add a new trusted user to the adapter.

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
```diff
@@ -130,6 +130,16 @@ abstract contract BaseFactory is Trust {
         rewardsRecipient = _recipient;
     }
 
+    /// @notice sets trusted address for an adapter
+    /// @dev factory must already be a trusted address for the adapter 
+    function setAdapterTrusted(
```
fedealconada, line 135: What do you think about adding this @jparklev ? it's just convenient to have this method because when we deploy an adapters via factories, the owner is the factory. In this case, for example, I was missing a `setClaimer` function on the factory which would be a passthrough to the `setClaimer` of the adapter.
To avoid missing these kind of passthrough methods, we could just have this `setAdapterTrusted` which is triggered by the owner of the factory and allows it to add a new trusted user to the adapter.

jparklev: Yes, and only ppl trusted on the factory could call this? Yea, I think this makes sense to have

jparklev (approved):
Looks good to go to me once the build is passing 👍

pkg/core/src/adapters/abstract/factories/BaseFactory.sol
fedealconada: yes!

pkg/core/src/adapters/implementations/claimers/PingPongClaimer.sol
abylaw: Got it!! Thank you, that makes sense ❤️

PR #337, opened 2023-02-08
GEN-2022: add function to set factory params

fedealconada:
## Motivation

Make our factories more flexible.

## Solution

Add `setFactoryParams` to BaseFactory.sol

## Implementer Checklist


jparklev (approved):
Smol but mighty

Also i've never seen an event with a struct! What does that even look like in the data

fedealconada:
> Smol but mighty
> 
> Also i've never seen an event with a struct! What does that even look like in the data

lol me neither, but it worked, and it looks like this:
image

abylaw (approved):
Yay! Nice

PR #338, opened 2023-02-09
GEN-2027: fix ifee

fedealconada:
## Motivation

Avoid mis-typing the ifee

## Solution

- Redeploy OwnableERC4626CropFactory
- Add deployment scripts helpers (both for forge scripts and hardhat) to help on calculating the ifee

## Implementer Checklist


pkg/deployments/tasks/20230102-ownable-erc4626-crop-factory/index.js
```diff
@@ -253,10 +257,9 @@ task(
...
     console.log("2. Set factory as trusted on Divider (multisig)");
-    console.log("3. Deploy ownable adapter on Periphery using factory (relayer)");
-    console.log("4. Go back to script, and deploy the claimer");
-    console.log("5. Set claimer to adapter (multisig)");
-    console.log("6. Create RLV using AutoRollerFactory and the rewards distributor address (relayer)");
-    console.log("7. Roll first series (relayer)");
+    console.log("3. Deploy ownable adapter on Periphery (Adapters UI using deployer address)");
+    console.log("4. Set claimer to adapter (deployer)");
+    console.log("5. Set multisig as trusted of adapter via factory `setAdapterIsTrusted` (multisig)");
+    console.log("5. Roll first series (Angle team or relayer)");
```
abylaw:
```suggestion
    console.log("6. Roll first series (Angle team or relayer)");
```

abylaw (approved):
Nice!! This makes sense to me. It's much clearer what the fees are now.

pkg/core/src/tests/test-helpers/TestHelper.sol
```diff
@@ -659,4 +659,20 @@ contract TestHelper is Test {
     function assertApproxEqAbs(uint256 a, uint256 b) public virtual {
         assertApproxEqAbs(a, b, 100);
     }
+
+    function percentageToDecimal(uint128 percentageFee) public returns (uint128) {
+        return percentageFee / 100;
```
jparklev: Thoughts on logging what the percentage will be for spot checks? e.g. `percentage: 0.05%`

We could use [log_named_decimal_uint](https://github.com/dapphub/ds-test/blob/cd98eff28324bfac652e63a239a60632a761790b/src/test.sol#L32) to get it [formatted nicely in solidity](https://github.com/foundry-rs/book/blob/c388d3541e2ca6ecdf526746a0539bd8b89f84ff/src/reference/ds-test.md?plain=1#L130)

pkg/core/script/2022-06-12-ERC4626Factory.s.sol
```diff
@@ -52,7 +54,7 @@ contract ERC4626FactoryScript is Script, StdCheats {
             stakeSize: 0.25e18, // 0.25 WETH
             minm: 2629800, // 1 month
             maxm: 315576000, // 10 years
-            ifee: 1e15, // 0.001%
+            ifee: th.percentageToDecimal(0.05e18), // 0.05%
```
jparklev: Perhaps since we are using this in `script` contract as well, it should be in a contract other than `TestHelper`? Like maybe we should make a general "utilities" contract of some kind. WDYT?

jparklev (approved):
Comment on logging for sanity checks and where to put the functionality, but overall agree this feels like a great step towards clarity and DX

pkg/core/src/tests/test-helpers/TestHelper.sol
fedealconada: sure! i'm logging it [here](https://github.com/sense-finance/sense-v1/pull/338/files#diff-ee8e939b6bb46fbe45268c19ac5a070f0b1d48ae0750bd26ea25f44128f27db3R70) on the script but I guess I could have it already included on each function so it always logs, that's your idea, right?

pkg/core/script/2022-06-12-ERC4626Factory.s.sol
fedealconada: sure!

fedealconada: pushed, thoughts?

pkg/core/src/tests/test-helpers/TestHelper.sol
jparklev: Yep!

pkg/core/script/2022-06-12-ERC4626Factory.s.sol
jparklev: LGTM

jparklev:
Anything we still need here before merging @fedealconada?

fedealconada:
> Anything we still need here before merging @fedealconada?

nothing! ill merge

PR #339, opened 2023-02-13
GEN-1888: zaps

fedealconada:
## Motivation

Add more flexibility to the Periphery with the use of zaps to:

- swap any token for PTs or YTs
- swap PTs or YTs for any token
- issue from any token
- combine to any token
- sponsor from any token

This PR is based on the [permit2 PR](https://github.com/sense-finance/sense-v1/pull/335) and it adds up the zap features.

## Solution

We use [0x API](https://docs.0x.org/0x-api-swap/api-references/get-swap-v1-quote) to generate a quote for a swap. For example, if we want to sponsor a series using `DAI` and our stake token is `WETH`, the portal will ping `https://api.0x.org/swap/v1/quote?sellToken=DAI&buyToken=0xae7ab96520DE3A18E5e111B5EaAb095312D7fE84&sellAmount=250000000000000000`

Then, we will use some of the properties of the response to create our quote and send it as a param to our `sponsorSeries` function, which receives both the quote and also the permit2 message to be able to pull the tokens (`DAI`) from the user. Some examples can be found [here](https://github.com/sense-finance/sense-v1/pull/339/files#diff-0247642c13110fa2468a1ca4090095e332f3d9f9b4f0308a3d35b87bd7fd37cfR145)

We are always doing swaps to or from `underlying` since we believe there will be either more liquidity, more liquidity sources on 0x than if we choose to use the `target` token.  

When doing actions from/to target or underlying (e.g swap PTs for target) there's no need to do any swap on 0x, we will just use the target as it is or wrap the underlying.

## Implementer Checklist


pkg/core/src/Periphery.sol
```diff
@@ -46,6 +58,12 @@ contract Periphery is Trust, IERC3156FlashBorrower {
     /// @notice Sense core Divider address
     BalancerVault public immutable balancerVault;
 
+    /// @notice Permit2 contract
+    IPermit2 public immutable permit2; // TODO: do we want this to be mutable?
```
jparklev: I think it's ok to have this immutable, seems like something that'd be quite disruptive to change for the ecosystem

pkg/core/src/Periphery.sol
```diff
@@ -950,6 +940,119 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         ptBal = ERC20(pt).balanceOf(address(this)) - ptBal;
     }
 
+    // @dev Swaps ETH->ERC20, ERC20->ERC20 or ERC20->ETH held by this contract using a 0x-API quote
+    function _fillQuote(SwapQuote memory quote) internal returns (uint256 boughtAmount) {
+        if (quote.sellToken == quote.buyToken) return 0; // No swap if the tokens are the same.
+        if (quote.swapTarget != exchangeProxy) revert Errors.InvalidExchangeProxy();
+
+        // Track our balance of the buyToken to determine how much we've bought.
+        boughtAmount = address(quote.buyToken) == ETH ? address(this).balance : quote.buyToken.balanceOf(address(this));
```
jparklev: This is tracking the initial balance of the buy token, right? so we can check the change. when would we expect some buy token to already be in the periphery?

pkg/core/src/Periphery.sol
```diff
@@ -950,6 +940,119 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
+        // Refund any unspent protocol fees to the sender.
+        uint256 refundAmt = address(this).balance;
+        if (address(quote.buyToken) == ETH) refundAmt = refundAmt - boughtAmount;
+        payable(msg.sender).transfer(refundAmt);
+
+        emit BoughtTokens(address(quote.sellToken), address(quote.buyToken), boughtAmount);
+        return boughtAmount;
+    }
+
+    function _getRevertMsg(bytes memory _returnData) internal pure returns (string memory) {
```
jparklev: Ohh nice, how did u figure this out?

pkg/core/src/Periphery.sol
```diff
@@ -950,6 +940,119 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
+            // TODO: return non-swapped underlying if there is?
+            // sell tokens for underlying and wrap into target
+            Adapter(adapter).unwrapTarget(_amt);
+            amt = _fillQuote(quote);
+        } else {
+            amt = _amt;
+        }
+    }
+
+    function _transferFrom(
```
jparklev, line 977: Similar to the library?

jparklev (commented):
Quick questions, but more will come! Should we put this PR up against the `permit` stuff to minimize the diff?

Also I wonder if we even still need the migrate liquidity function with RLVs being the default now 🤔

pkg/core/src/Periphery.sol
fedealconada: hmm... never, you mean that we could just use the whole total balance that is in the Periphery?

pkg/core/src/Periphery.sol
fedealconada: stole it from somewhere haha

pkg/core/src/Periphery.sol
fedealconada: indeed!

fedealconada:
> Quick questions, but more will come! Should we put this PR up against the `permit` stuff to minimize the diff?

that was the idea but i forgot xD
> 
> Also I wonder if we even still need the migrate liquidity function with RLVs being the default now 🤔

yes! had the same thought, that's why it was I commented it out but wanted to check with you.

pkg/core/src/Periphery.sol
jparklev: tbh I think I might not be understanding this part of the code yet heh. My understanding was that `boughtAmount` would always be `0` here since nothing had been purchased yet?

fedealconada: i think i understand your point and we can just remove that line.
basically, the idea there was that we init `boughtAmount` with the current periphery's `buyToken` balance (which yes, will be 0 unless someone has airdropped some tokens haha, so this is why we could actually remove that line)
then (line 906), we execute the swap so now the periphery has received some buy tokens and we do `boughtAmount = buyToken.balanceOf(periphery) - boughtAmount` to calculate how many tokens we've just bought.

pkg/core/src/Periphery.sol
```diff
@@ -107,17 +131,12 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
-
-        // Transfer stakeSize from sponsor into this contract
-        permit2.permitTransferFrom(
-            permit.msg, // permit message.
-            IPermit2.SignatureTransferDetails({ to: address(this), requestedAmount: stakeSize }), // transfer recipient and amount.
-            msg.sender, // owner of the tokens
-            permit.sig // packed permit.sig that was the result of signing the EIP712 hash of `permit`.
-        );
+        if (address(quote.sellToken) != ETH) _transferFrom(permit, stake, stakeSize);
+        if (address(quote.sellToken) != stake) _fillQuote(quote);
```
jparklev, line 125: So this would be if they just transfer the stake in directly rather than swapping for it, right?

pkg/core/src/Periphery.sol
```diff
@@ -1113,6 +892,115 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
+        // allowance to 0 before being able to update it.
+        // TODO: add tests for USDT!!
+        if (address(quote.sellToken) != ETH)
+            ERC20(address(quote.sellToken)).safeApprove(quote.spender, type(uint256).max);
+
+        // Call the encoded swap function call on the contract at `swapTarget`,
+        // passing along any ETH attached to this function call to cover protocol fees.
+        (bool success, ) = quote.swapTarget.call{ value: msg.value }(quote.swapCallData);
+        // if (!success) revert(_getRevertMsg(res));
+        if (!success) revert Errors.ZeroExSwapFailed();
```
jparklev: Maybe can give this error an argument and pass in the raw `res`

pkg/core/src/Periphery.sol
```diff
@@ -1113,6 +892,115 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
+
+        // Use our current buyToken balance to determine how much we've bought.
+        boughtAmount = (
+            address(quote.buyToken) == ETH ? address(this).balance : quote.buyToken.balanceOf(address(this))
+        );
+
+        // TODO: add tests for this! and check if we actually need it
```
jparklev: 👍

pkg/core/src/Periphery.sol
```diff
@@ -1147,4 +1038,5 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         uint256 amountOut,
         bytes4 indexed sig
     );
+    event BoughtTokens(address indexed sellToken, address indexed buyToken, uint256 indexed boughtAmount);
```
jparklev: Should we include the sold amount so that a price could be backed out?

pkg/core/src/Periphery.sol
```diff
@@ -1113,6 +892,115 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
+    /// simply unwrap tokens or do nothing
+    function _fromTarget(
+        address adapter,
+        uint256 _amt,
+        SwapQuote calldata quote
+    ) internal returns (uint256 amt) {
+        if (address(quote.buyToken) == Adapter(adapter).underlying()) {
+            amt = Adapter(adapter).unwrapTarget(_amt);
+        } else if (address(quote.buyToken) != Adapter(adapter).target()) {
+            // TODO:the issue here is that the quote needs to calculate off-chain the amount of underlying that will be received from the unwrapTarget
```
jparklev: I guess there's no way to dynamically update the `swapCallData` on-chain to reflect the resulting underlying the contract now has?

jparklev (commented):
Slowly the comments and questions continue coming in :p

pkg/core/src/Periphery.sol
fedealconada: yep

pkg/core/src/Periphery.sol
fedealconada: Not sure I follow you

pkg/core/src/tests/test-helpers/TestHelper.sol
```diff
@@ -667,4 +668,17 @@ contract TestHelper is Test, Permit2Helper {
     function assertApproxEqAbs(uint256 a, uint256 b) public virtual {
         assertApproxEqAbs(a, b, 100);
     }
+
+    function _getQuote(
+        address adapter,
+        address fromToken,
+        address toToken
+    ) public returns (Periphery.SwapQuote memory quote) {
+        MockAdapter adapter = MockAdapter(adapter);
```
jparklev: Unused, perhaps?

pkg/core/src/Periphery.sol
```diff
@@ -159,246 +178,153 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
     /// @param ytBal Balance of YTs to swap
     /// @param receiver Address to receive the Target
-    /// @return tBal amount of Target received
-    function swapYTsForTarget(
+    /// @param permit Permit to pull YTs
+    /// @param quote Quote with swap details
+    /// @return amt amount of Target received
+    /// @dev if quote.buyToken is neither target nor underlying, it will unwrap target
+    /// and swap it on 0x
+    // TODO: why we dont have minAccepted here? Should we?
```
jparklev: ...probably. Could be a simple check like `amt > minAccepted` at the end

pkg/core/src/Periphery.sol
jparklev: Like a way to update `swapCallData` with the result of `unwrapTarget` on-chain, which would remove the need to calculate the amount off-chain?

pkg/core/src/tests/test-helpers/TestHelper.sol
```diff
@@ -667,4 +668,17 @@ contract TestHelper is Test, Permit2Helper {
     function assertApproxEqAbs(uint256 a, uint256 b) public virtual {
         assertApproxEqAbs(a, b, 100);
     }
+
+    function _getQuote(
```
jparklev: I might be missing something, but are `swapCallData` and `swapTarget` set to null when we use this function? :) Are we testing a `quote` with non-null values somewhere as well?

pkg/core/src/Periphery.sol
```diff
@@ -159,246 +178,153 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
-            permit.msg,
-            IPermit2.SignatureTransferDetails({ to: address(this), requestedAmount: tBal }),
-            msg.sender,
-            permit.sig
-        );
-        (tAmount, issued, lpShares) = _addLiquidity(adapter, maturity, tBal, mode, minBptOut, receiver, permit);
+        PermitData calldata permit,
+        SwapQuote calldata quote
+    ) private returns (uint256 amt) {
+        if (address(quote.sellToken) != address(0) && address(quote.sellToken) != Adapter(adapter).underlying())
```
jparklev: What happens if `sellToken` is `address(0)`

pkg/core/src/Periphery.sol
```diff
@@ -159,246 +178,153 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
-    /// @notice Adds liquidity providing target
-    /// @param adapter Adapter address for the Series
-    /// @param maturity Maturity date for the Series
-    /// @param tBal Balance of Target to provide
-    /// @param mode 0 = issues and sell YT, 1 = issue and hold YT
-    /// @param minBptOut Minimum BPT the user will accept out for this transaction
-    /// @param receiver Address to receive the BPT
-    /// @dev see return description of _addLiquidity
-    function addLiquidityFromTarget(
+    function _swap(
```
jparklev: thoughts on `swapFixedTermToken` or `swapSenseToken`?

pkg/core/src/Periphery.sol
```diff
@@ -82,18 +96,28 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         bytes sig;
     }
 
+    struct SwapQuote {
+        IERC20 sellToken;
+        IERC20 buyToken;
```
```suggestion
        ERC20 buyToken;
```
same for the above, if you're down

pkg/core/src/Periphery.sol
fedealconada: sgtm! i'm anyways waiting for the final version and see if we have space to add the _getRevertMsg

pkg/core/src/Periphery.sol
fedealconada: ohhh, hmm.. i guess it's technically possible, but would be little bit complex and will increase the size of the contract. I think it would be something like [this](https://sourcegraph.com/github.com/cl2089/etherscan-contract-crawler/-/blob/contracts/0x777c22F63A83ee27D375865C65AfdaBaf96e213a_CrosschainForwarder/20_35_SwapCalldataUtils.sol).
The problem is that if 0x decides to change however they are returning the swaCallData, then we would need to adjust that and so on...

pkg/core/src/tests/test-helpers/TestHelper.sol
fedealconada: Oh, I think I should have clarified. For local tests, I'm not really testing the swapping (I could potentially do a 0x Router Mock or something like that, maybe add some expectCall), so all the tests work with swapping either fromUnderlying, fromTarget, toUnderlying or toTarget (where we only wrap or unwrap if necessary).
On mainnet tests I am effectively testing the fillQuote with real quotes.

pkg/core/src/Periphery.sol
fedealconada: For example: User wants to swap PTs for Target
- sellToken = address(0)
- buyToken = address(target)
This use case will not do any swap on 0x, so we don't care about the `sellToken` value, we just care about the `buyToken`. If `sellToken = address(0)` that means that our only options for `buyToken` are underlying or target. So here we will just swap PTs for Target and return the target.

pkg/core/src/Periphery.sol
fedealconada: sounds good!

pkg/core/src/Periphery.sol
fedealconada: yess, it's legacy from my first implementation :)

pkg/core/src/Periphery.sol
jparklev: > If sellToken = address(0) that means that our only options for buyToken are underlying or target

Ahh I think I follow. This is the `quote` for what to do with the Target we get from swapping PTs in, so if `sellToken` is `address(0)`, then we're saying "we won't be selling anything on 0x", in which case it only makes sense to have Target or Underlying for the `buyToken` (ie the "out" token, since those are the only tokens we can get w.o. swapping on 0x)

fedealconada: Exactly!

jparklev: I guess `sellToken` could also be `address(target)` here?

jparklev: Or, this brings up a follow up question of – why do we check the `sellToken` here but not in e.g. `removeLiquidity` (where we're also in a position where we can only sell Target or Underlying. or am I missing something?)

jparklev: Like, when we're selling Sense tokens, we need the `sellToken` to be either Target or Underlying, b/c that's what we'll have after swapping on Space

Then when we're buying Sense tokens, we need `buyToken` to be either Target or Underlying, b/c that's what we need  to swap on Space

Is that all correct?

pkg/core/src/Periphery.sol
```diff
@@ -637,6 +635,7 @@ contract Periphery is Trust, IERC3156FlashBorrower {
                     adapter,
                     maturity,
                     issued,
+                    0, // TODO: do we want to use 0 here or have yet another param? (minAccepted?)
```
jparklev: safe thing is probably to add another param -_-

pkg/core/src/Periphery.sol
fedealconada: > Or, this brings up a follow up question of – why do we check the sellToken here but not in e.g. removeLiquidity (where we're also in a position where we can only sell Target or Underlying. or am I missing something?)

I'll add tests but, we don't really need to check the `quote.sellToken` for a `removeLiquidity`.
Let's say I want to removeLiquidity to DAI:
- pull tokens (whichever token is in the permit message). We should be technically pulling BPT tokens but if it is any other token we will try to pull those tokens.
- execute _removeLiquidityFromSpace which will return target (unless quote.buyToken is PTs in which case just returns PTs and tBal). This would fail if the Periphery does not have the BPT tokens. 
- unwrap target into underlying 
- Periphery now has underlying which we need to swap for buyToken (e.g DAI). If the fillQuote's sellToken is other than underlying (e.g `address(0)` or `USDC`), it will just fail because the Periphery is not holding that other token

pkg/deployments/hardhat.utils.js
```diff
@@ -246,3 +246,82 @@ exports.approve = async (token, owner, spender) => {
     await token.approve(spender, ethers.constants.MaxUint256).then(tx => tx.wait());
   }
 };
+
+// TODO: is there any better way to calculate nonces? should I use last transaction and have ordered nonces?
```
alexfertel, line 263: In case it is helpful. There is a `Smart Nonces` section [here](https://github.com/dragonfly-xyz/useful-solidity-patterns/tree/main/patterns/permit2#advanced-integrations). @fedealconada @jparklev

jparklev:
Thoughts on renaming `Periphery` to `PeripheryV2` @fedealconada?

pkg/deployments/hardhat.utils.js
fedealconada: yeah, i read it but it does not say much, there's no example nor nothing. In any case, I checked this out on the Uniswap channel and they told me I can use normal sequential nonces

pkg/deployments/hardhat.config.js
```diff
@@ -124,7 +124,7 @@ module.exports = {
             enabled: true,
             // The Balancer Vault becomes too large to deploy with more optimizer runs,
             // but we only need to deploy it on sim runs (we use the existing mainnet Vault on prod runs)
-            runs: process.env.SIM === "true" ? 100 : 500,
+            runs: process.env.SIM === "true" ? 100 : 100,
```
abylaw, line 127:
```suggestion
            runs: 100,
```

pkg/deployments/hardhat.config.js
```diff
@@ -113,7 +113,7 @@ module.exports = {
         settings: {
           optimizer: {
             enabled: true,
-            runs: process.env.SIM === "true" ? 100 : 500,
+            runs: process.env.SIM === "true" ? 100 : 100,
```
abylaw, line 116:
```suggestion
            runs: 100,
```

pkg/deployments/hardhat.config.js
```diff
@@ -91,7 +91,7 @@ module.exports = {
       // saveDeployments: false,
       // blockGasLimit: 21000000,
       // blockNumber: 13491969,
-      // allowUnlimitedContractSize: false,
+      allowUnlimitedContractSize: true,
```
abylaw: still intended to be set to `true`? :)

abylaw:
@fedealconada just have a little more to review!! will be done tmr for sure -- ty for patience 🙏

pkg/deployments/deploy/prod/06-series.js
```diff
@@ -31,36 +44,740 @@ module.exports = async function () {
     await sponsor(adapter(chainId).target);
   }
 
+  // PTs SWAPS
+  async function _swapForPTs(adapter, maturity, quote) {
```
abylaw, line 49: would slightly prefer these functions to have more arguments (like signer) and not pull in variables via closure, as the function becomes difficult to move, understand & refactor

pkg/core/src/tests/Periphery.tm.sol
```diff
@@ -642,4 +1289,356 @@ contract PeripheryMainnetTests is PeripheryTestHelper {
...
...24316b9ae028f1497c275eb9192a3ea0f670223df021240000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001c000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000e00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000a00000000000000000000000000000000000000000000000000000000000000002000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2000000000000000000000000eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee0000000000000000000000000000000000000000000000000000000000000000869584cd0000000000000000000000001000000000000000000000000000000000000011000000000000000000000000000000000000000000000099788145ed63f48648";
```
abylaw: nit: pull out the swap call data into constants for readability

pkg/core/src/tests/Periphery.tm.sol
```diff
@@ -239,7 +470,13 @@ contract PeripheryMainnetTests is PeripheryTestHelper {
...
+        (address pt, address yt) = periphery.sponsorSeries(
+            address(fadapter),
+            maturity,
+            false,
+            data,
+            _getQuote(address(fadapter), AddressBook.DAI, AddressBook.DAI)
+        );
 
         // Check pt and yt deployed
         assertTrue(pt != address(0));
```
abylaw, line 495: add test for edge-case where `sellToken` is `0x0` or quote is otherwise malformed?

pkg/core/src/tests/Periphery.tm.sol
```diff
@@ -642,4 +1289,356 @@ contract PeripheryMainnetTests is PeripheryTestHelper {
...
+            quote.sellToken = ERC20(fromToken);
+            quote.buyToken = ERC20(toToken);
+            return quote;
+        }
+        MockAdapter adapter = MockAdapter(adapter);
+        if (fromToken == address(0)) {
+            if (toToken == adapter.underlying() || toToken == adapter.target()) {
+                // Create a quote where we only fill the buyToken (with target or underlying) and the rest
+                // is empty. This is used by the Periphery so it knows it does not have to perform a swap.
+                quote.buyToken = ERC20(toToken);
```
abylaw: i found it a bit hard to follow the nested if functions, could we break up this function a bit?

maybe `getSellQuote`/`getBuyQuote`?

pkg/core/src/tests/test-helpers/TestHelper.sol
```diff
@@ -667,4 +668,12 @@ contract TestHelper is Test, Permit2Helper {
     function assertApproxEqAbs(uint256 a, uint256 b) public virtual {
         assertApproxEqAbs(a, b, 100);
     }
+
+    function _getQuote(address fromToken, address toToken) public returns (Periphery.SwapQuote memory quote) {
+        if (fromToken == address(0)) {
+            quote.buyToken = ERC20(toToken);
+        } else {
+            quote.sellToken = ERC20(fromToken);
```
abylaw, line 676: `quote.buyToken = ERC(toToken)`?

pkg/core/src/Periphery.sol
```diff
@@ -1113,6 +914,114 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
+    }
+
+    /// @notice Given an amount and a quote, decides whether it needs to unwrap and make a swap on 0x,
+    /// simply unwrap tokens or do nothing
+    function _fromTarget(
+        address adapter,
+        uint256 _amt,
+        SwapQuote calldata quote
+    ) internal returns (uint256 amt) {
+        if (address(quote.buyToken) == Adapter(adapter).underlying()) {
```
abylaw: mega nit: maybe slightly easier to understand reversed:

```
if (address(quote.buyToken) == Adapter(adapter.target()) {
  amt = amt
} else if (address(quote.buyToken) == Adapter(adapter).underlying()) {
  amt = Adapter(adapter).unwrapTarget(_amt);
} else {
 ...
}

File: pkg/core/src/Periphery.sol [hunk 27]
```diff
@@ -459,140 +360,63 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
-                mode,
-                minBptOut,
-                msg.sender,
-                permit
-            );
-        }
+        amt = _fromTarget(adapter, amt, quote);
+        address(quote.buyToken) == ETH
+            ? payable(receiver).transfer(amt)
+            : ERC20(address(quote.buyToken)).safeTransfer(receiver, amt); // transfer bought tokens to receiver
```
abylaw: since this occurs a few times, `_transferTo(reveiver, token, amt)`?

abylaw: mega nit: remove comment, code is self-explanatory

pkg/core/src/Periphery.sol
```diff
@@ -159,297 +174,183 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
+        (tAmount, issued, lpShares) = _addLiquidity(
             adapter,
             maturity,
-            lpBal,
-            minAmountsOut,
+            _toTarget(adapter, amt, quote),
             minAccepted,
-            intoTarget,
+            minBptOut,
+            mode,
```
abylaw, line 351: do we want to support selling YTs into something other than target as well?

pkg/core/src/Periphery.sol
```diff
@@ -459,140 +360,63 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
+    /// and swap it on 0x
+    /// if quote.buyToken is PT, it will return target and PTs
+    function removeLiquidity(
         address adapter,
         uint256 maturity,
         uint256 lpBal,
         uint256[] memory minAmountsOut,
         uint256 minAccepted,
-        bool intoTarget,
+        bool swapPTs,
```
abylaw, line 377: similar to `addLiquidity`, do we want to also swap the target earned from swapping PTs into `quote.buyToken`?

pkg/core/src/Periphery.sol
```diff
@@ -459,140 +360,63 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
     /// @param maturity Maturity date for the Series [unix time]
-    /// @param targetIn Amount of Target to issue with
+    /// @param amt Amount to issue with
     /// @dev The balance of PTs and YTs minted will be the same value in units of underlying (less fees)
     /// @param receiver Address where the resulting PTs and YTs will be transferred to
+    /// @param permit Permit to pull tokens
+    /// @param quote Quote with swap details
+    /// @return uBal Amount of PTs and YTs minted
+    /// @dev if quote.sellToken is neither target nor underlying, it will unwrap target
+    /// and swap it on 0x
```
abylaw:
```suggestion
    /// @dev if quote.sellToken is neither target nor underlying, it will swap on 0x and wrap to target
```

pkg/core/src/Periphery.sol
```diff
@@ -602,43 +426,29 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
-        sigs[0] = IPermit2.SignatureTransferDetails({ to: address(this), requestedAmount: amt });
-        sigs[1] = IPermit2.SignatureTransferDetails({ to: address(this), requestedAmount: amt });
-
-        permit2.permitTransferFrom(permit.msg, sigs, msg.sender, permit.sig);
-        uint256 tBal = divider.combine(adapter, maturity, amt);
-        ERC20(Adapter(adapter).underlying()).safeTransfer(receiver, uBal = Adapter(adapter).unwrapTarget(tBal)); // Send Underlying to the receiver
+        amt = _fromTarget(adapter, divider.combine(adapter, maturity, uBal), quote);
+        address(quote.buyToken) == ETH
+            ? payable(receiver).transfer(amt)
+            : ERC20(address(quote.buyToken)).safeTransfer(receiver, amt); // transfer bought tokens to receiver
```
abylaw: `_transferTo(...)`

pkg/core/src/tests/Periphery.tm.sol
```diff
@@ -275,6 +512,160 @@ contract PeripheryMainnetTests is PeripheryTestHelper {
...
+        _addLiquidityFromToken(adapter, maturity, quote, amt);
+
+        // 4. Add liquidity from Underlying
+        token = ERC20(MockAdapter(adapter).underlying());
+        amt = 10**token.decimals(); // 1 Target
+        // Create quote only with sellToken as target. We don't care about the other params
+        // since no swap on 0x will be done
+        quote = _getQuote(adapter, address(token), address(0));
+        _addLiquidityFromToken(adapter, maturity, quote, amt);
+    }
```
abylaw, line 614: ditto above: add test for when quote is malformed

pkg/core/src/tests/Periphery.tm.sol
```diff
@@ -275,6 +512,160 @@ contract PeripheryMainnetTests is PeripheryTestHelper {
...
+        // will be ignored in this case. When removing liquidity, sellToken would always be the LP token
+        // TODO: modify _removeLiquidityToToken so we can test this
+        // quote = _getQuote(adapter, AddressBook.DAI, MockAdapter(adapter).target());
+        // _removeLiquidityToToken(adapter, maturity, quote, amt);
+
+        // 4 Remove liquidity to Underlying
+        // Create quote only with sellToken as target. We don't care about the other params
+        // since no swap on 0x will be done
+        quote = _getQuote(adapter, address(0), MockAdapter(adapter).underlying());
+        _removeLiquidityToToken(adapter, maturity, quote, amt);
```
abylaw, line 660: ditto above: add test for when quote is malformed

(applies to swap to/from pts/yts as well: e.g. swap token for pts with fromToken = `0x0` or `pt/yt.address`

abylaw (commented):
nice!!!! so excited for the feature :)

One thing I did notice is that one side of the `SwapQuote` is always ignored, so perhaps it can be simplified to only take one token, which could simplify the interface.

pkg/deployments/hardhat.config.js
fedealconada: no no haha, it's just to let me try the tests, once I manage to reduce the code it would have to be set to `false` xD

pkg/core/src/tests/test-helpers/TestHelper.sol
fedealconada: It must be `ERC20` since it's the name of the interface/contract. Just `ERC` would not compile since it's no related to any interface/contract defined.

pkg/core/src/Periphery.sol
fedealconada: totallyl!

pkg/core/src/Periphery.sol
fedealconada: yess

pkg/core/src/Periphery.sol
fedealconada: I don't think so, either you want to add the more Target you can to the pool, or somehow you prefer to keep those YTs.
I'd actually wonder if we really need to option to keep the YTs @jparklev ?  Because if not, we could remove that, saving some code size :)

pkg/core/src/Periphery.sol
fedealconada: similar to my prev answer:
I don't think so, either you want to remove the more liquidity you can from the pool (in the same asset), or somehow you also prefer keeping those PTs.
I'd actually wonder if we really want to offer the option to keep the PTs @jparklev ? Because if not, we could remove that, saving some code size :)

fedealconada: if we remove this option and the also the one of my prev message, we are reducing the size of the contract and amount of kb that matches more or less the addition of the `function swapYTsForTarget` to keep compatibility with the RLVs @jparklev

pkg/core/src/tests/test-helpers/TestHelper.sol
abylaw: Ah apologies :) That was a typo. I meant to suggest that when `fromToken != address(0)`, should we be setting the `buyToken` on the quote as well?

But I think it was still based on a misunderstanding - we ignore `buyToken` if `fromToken` is set :b

pkg/core/src/Periphery.sol
abylaw: true, true. 

iirc keeping YTs were for larger liquidity providers - selling their YTs would affect the market rate too much + it has a better return profile? @jparklev probably knows better~

pkg/core/src/Periphery.sol
abylaw: hmm!

say i have a cDAI PT / cDAI pool...and I want to withdraw into ETH.
say i don't want to keep my PTs - so they get swapped for cDAI. wouldn't we want to also convert that cDAI into ETH as well?

pkg/core/src/Periphery.sol
jparklev: > iirc keeping YTs were for larger liquidity providers - selling their YTs would affect the market rate too much + it has a better return profile? @jparklev probably knows better~

Partially true, yes! Though when the yt swap is triggered, it's after the LP has issued and joined the pool, so there's almost always actually enough liquidity to do the swap (their own!). However, holding YTs does give one a different exposure (often better, as u mentioned) and makes exiting liquidity positions earlier easier (combine YTs with exited PTs). Does that make sense?

B/c of that I might look elsewhere first for cuts, though I know we've gotten all of the low hanging fruit. I'm increasingly thinking we can pull out pool manager stuff... but let's chat about it mon

pkg/core/src/Periphery.sol
```diff
@@ -148,308 +153,204 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
         ERC20(divider.yt(adapter, maturity)).safeTransfer(receiver, ytBal);
     }
 
-    /// @notice Swap PTs for Target of a particular series
+    /// @notice Swap PTs of a particular series
     /// @param adapter Adapter address for the Series
     /// @param maturity Maturity date for the Series
     /// @param ptBal Balance of PT to sell
-    /// @param minAccepted Min accepted amount of Target
+    /// @param minAccepted Min accepted amount of Target when selling them on Space // TODO: are we good with this? Or should be minAccepted amount of tokens?
```
fedealconada: @jparklev  what do you think about this? Should we have 2 params or only one? Maybe just `minAccepted` in Token terms?

jparklev: I think we only need 1 param, and having it in token terms makes sense to me 👍

jparklev (approved):
Latest changes check out. Had the thought that we could consider removing the entire concept of `verified` as well, but... if we're fine size-wise then I wouldn't worry about it at this point (how far off are we?). Other things on the horizon for us to work on :)

So, other than the build and the `minAccepted` thread, LGTM. Pushing forward with sherlock, think we're ready?

(oh! also! feel free to resolve threads when they've been addressed, might be good for decluttering a bit ~)

pkg/core/src/Periphery.sol
fedealconada: i missed this one!  this is actually what we are doing now, when removing liquidity: we get `cDAI` and `PTs` but since `swapPTs` is `true` we swapped those PTs for target, so we end up with all target (`cDAI`) and then we unwrap all of that target to underlying (`DAI`) and finally swap it (on 0x) for `ETH`.

What it's **not** possible is to decide to keep the target from the PTs as target. I don't think it's something we want but curious about your thoughts @abylaw and @jparklev

fedealconada:
> Latest changes check out. Had the thought that we could consider removing the entire concept of `verified` as well, but... if we're fine size-wise then I wouldn't worry about it at this point (how far off are we?). Other things on the horizon for us to work on :)
> 
well, we are on the very edge haha but it's working 🤷🏼‍♂️ we can do that if we have any issues or, if after the changes from audit we are exceeding the size 

> So, other than the build and the `minAccepted` thread, LGTM. Pushing forward with sherlock, think we're ready?
> 
pushed! you can take a look :)

> (oh! also! feel free to resolve threads when they've been addressed, might be good for decluttering a bit ~)

done!

pkg/core/src/Periphery.sol
abylaw: Ah got it! That seems fine to me from a UX perspective, thoughts @jparklev ?

abylaw (approved):
LGTM as well :)

PR #340, opened 2023-02-15
GEN-2034: add idleUSDCJunior4626 to 4626 tests

fedealconada:
## Motivation

Add support for idleUSDCJunio4626 on our 4626 test suite. 

## Implementer Checklist

### DIFF
--- a/pkg/core/src/tests/adapters/ERC4626Adapter.tm.sol
+++ b/pkg/core/src/tests/adapters/ERC4626Adapter.tm.sol
@@ -200,6 +200,7 @@ contract ERC4626Adapters is ForkTest {
 
     /// @notice scale() and scaleStored() have the same implementation
     /// @dev given the fact that totalAssets() might be implemented differently by each protocol,
+    /// you may need to implement a custom case for the token given on _increaseVaultYield() function
     function testScaleAndScaleStored() public {
         // Sanity check: vault is already initialised and scale > 0
         uint256 scale = erc4626Adapter.scale();
--- a/pkg/core/src/tests/adapters/ERC4626Adapter.tm.sol
+++ b/pkg/core/src/tests/adapters/ERC4626Adapter.tm.sol
@@ -47,6 +47,18 @@ interface Token {
     ) external;
 }
 
+interface IDLE {
+    function idleToken() external returns (address);
+
+    function mintIdleToken(
+        uint256 _amount,
+        bool,
+        address _referral
+    ) external returns (uint256 mintedTokens);
+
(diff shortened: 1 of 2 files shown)


pkg/core/src/tests/adapters/ERC4626Adapter.tm.sol
```diff
@@ -74,9 +86,10 @@ contract ERC4626Adapters is ForkTest {
         // set `userWithAssets` if exists
         try vm.envAddress("USER_WITH_ASSETS") returns (address _userWithAssets) {
```
jparklev, line 87: For my own edification, where are we setting this currently?

jparklev (approved):

pkg/core/src/tests/adapters/ERC4626Adapter.tm.sol
fedealconada: This can be set as an env (but it's not mandatory to use it), for example, you could call the script like:
`export ERC4626_ADDRESS=0x79F05f75df6c156B2B98aC1FBfb3637fc1e6f048 USER_WITH_ASSETS=0x1982b2f5814301d4e9a8b0201555376e62f82428 && yarn test:4626:integration`

jparklev: Got it – are we doing that with CI, or just manually for now?

fedealconada: So, these are tests we run on mainnet fork and, when running on CI, we just don't use the `USER_WITH_ASSETS` since it's optional. But, we also use the command `yarn test:4626:integration` when reviewing a new 4626 target we are gonna be integrating with. When doing so, we manually run this command, passing the `ERC4626_ADDRESS` (mandatory) and the  `USER_WITH_ASSETS`.
This PR is just to add support for the `idleUSDCJunior4626` since we have [this](https://github.com/sense-finance/sense-v1/blob/GEN-2034-idleUSDCJunior4626/pkg/core/src/tests/adapters/ERC4626Adapter.tm.sol#L202) test in the test suite which requires customising the logic of the [_increaseVaultYield()](https://github.com/sense-finance/sense-v1/blob/GEN-2034-idleUSDCJunior4626/pkg/core/src/tests/adapters/ERC4626Adapter.tm.sol#L278)

jparklev: Yep, I see it in the 4626 review process doc. Makes sense, ty for the context :)

abylaw (approved):
LGTM :)

PR #343, opened 2023-03-16
GEN-2119: integrate aura vaults

fedealconada:
## Motivation

Integrate Aura vaults to Sense Finance

## Solution

To integrate Aura we have to work on 2 limitations:
- Even though the vaults are ERC4626, the asset:share **ratio is always 1:1**, so we can’t really use them as they are.
- The vaults shares are **non-transferrable** (they have `transfer` and `transferFrom` overridden so they always revert.

The proposed solution is to:
- create an `AuraVaultWrapper` which wraps an Aura vault: this way, we will bypass the `transfer`/`transferFrom` restriction.
    - set this wrapper as our `target` which represents 1:1 the Aura vault share. It will contain all Sense's users Aura tokens of a particular vault, representing a big position and each user will have a share of it.
    - the underlying of this wrapper will be the base asset of the Balancer pool (e.g for `auraB-rETH-WETH-STABLE-vault`, whose Balancer pool is `B-rETH-WETH` the underlying will be `WETH`.
- for the exchange rate (scale value), we will use the `getRate()` function from the Balancer pool which: *returns the appreciation of one BPT relative to the underlying tokens*

We will then create an `AuraAdapter` whose target is a wrapped Aura vault and its underlying is the base asset of the balancer pool that underlies the Aura vault. 

### Wrap/Unwrap
TODO: join/exit
TODO: deposit with BPT

### Rewards

`AuraAdapter` will inherit from Crops since it will be receiving `BAL` and `AURA` rewards. When rewards are distributed by Aura, they will go to where the Aura position is held, which is the wrapper contract. 

It will also override the `_claimRewards` function to be able to claim rewards on behalf of the wrapper (which is who maintains the aura vault position) and also to be able to extract those rewards (from the wrapper to the adapter). This is possible because `AuraVaultWrapper` inherits the `ExtractableReward`.

### Aura yield mechanism 
_from Aura's Discord_

Basically, each day Aura pools get harvested (by arbers who get a small portion of the fees). When harvested, all the unclaimed `BAL` is claimed, `AURA` is minted based on claimed `BAL` (aura minted per bal formula on Aura docs), fees are taken for auraBAL.
The remaining rewards are then built into a queue(or buffer) that pays out the earned BAL + minted aura less fees over a 7 day period at an even rate per block.
This means that the current yield on AURA is made up of 1/7th of each of the last 7 days of deposits. If everyone deposits on day 0, and no one deposits or withdraws for 7 days, than the projected and actual yields should match.  If more capital has left the pool than joined over the last 7 days, the current yield should be above projected, and if capital is joining the pool every day, the current will remain below the projected until inflows taper off.

## Implementer Checklist


jparklev (commented):
Nice, need to review the wrapper in more detail, but teh adapter is looking good so far to me (other than vestige comments and things)!

pkg/core/src/tests/adapters/AuraAdapter.tm.sol
```diff
@@ -434,4 +469,33 @@ contract AuraAdapters is AuraAdapterTestHelper {
...
+        // load wallet
+        deal(s, address(this), sSize);
+        deal(t, address(this), 10**tDecimals);
+
+        // roll 1st series
+        rlv.roll();
+        console.log("- First series sucessfully rolled!");
+
+        // check we can deposit
+        rlv.deposit(5 * 1**tDecimals, address(this)); // deposit 1 target
```
jparklev: 5 target? and should it be `10**tDecimals`?

pkg/core/src/adapters/abstract/extensions/Crops.sol
```diff
@@ -107,6 +107,8 @@ abstract contract Crops is Trust {
         }
     }
 
+    // TODO: should we remove the reconcile?
```
jparklev: Could u expand on this?

pkg/deployments/hardhat.config.js
```diff
@@ -113,7 +113,7 @@ module.exports = {
         settings: {
           optimizer: {
             enabled: true,
-            runs: process.env.SIM === "true" ? 100 : 100,
+            runs: process.env.SIM === "true" ? 100 : 500,
```
jparklev: I wonder if we still need this distinction? (ie maybe we could always use the same # of runs)

pkg/core/src/tests/adapters/AuraAdapter.tm.sol
```diff
@@ -0,0 +1,531 @@
...
+
+    function testMainnetAuraAdapterScale() public {
+        uint256 aTokenUnderlying = _getRate();
+        assertEq(adapter.scale(), aTokenUnderlying);
+    }
+
+    function testMainnetAuraAdapterPriceFeedIsValid() public {
+        // Check last updated timestamp is less than 1 hour old
+        (address oracle, , , , , , , ) = adapter.adapterParams();
+        (, int256 ethPrice, , uint256 ethUpdatedAt, ) = PriceOracleLike(oracle).latestRoundData(); // ETH:USD
```
jparklev: We don't need this test for ETH underlying assets, right?

pkg/core/src/adapters/implementations/aura/AuraVaultWrapper.sol
```diff
@@ -0,0 +1,296 @@
...
+        request = IVault.ExitPoolRequest(poolAssets, minAmountsOut, userData, false);
+    }
+
+    /// @dev should be overriden if and only if BPT is one of the pool tokens
+    function _getBPTIndex() internal view virtual returns (uint256) {
+        return type(uint256).max;
+    }
+
+    /// @notice return index of the element if found, else return uint256.max
+    function find(IAsset[] memory array, address element) internal pure returns (uint256 index) {
```
jparklev, line 254: Feels like this should be in a library?

pkg/core/src/adapters/implementations/aura/AuraVaultWrapper.sol
```diff
@@ -0,0 +1,296 @@
...
+
+        // must drop BPT index as well
+        exitTokenIndex = _getBPTIndex() < exitTokenIndex ? exitTokenIndex - 1 : exitTokenIndex;
+        // 0 = EXACT_BPT_IN_FOR_ONE_TOKEN_OUT
+        bytes memory userData = abi.encode(0, lpAmt, exitTokenIndex);
+
+        request = IVault.ExitPoolRequest(poolAssets, minAmountsOut, userData, false);
+    }
+
+    /// @dev should be overriden if and only if BPT is one of the pool tokens
```
jparklev, line 248: can that happen 🙀

pkg/core/src/adapters/implementations/aura/AuraVaultWrapper.sol
```diff
@@ -0,0 +1,296 @@
...
+
+    function _assembleExitRequest(uint256 lpAmt) internal view virtual returns (IVault.ExitPoolRequest memory request) {
+        uint256[] memory minAmountsOut = new uint256[](poolAssets.length);
+
+        // encode user data
+        uint256 exitTokenIndex = find(poolAssets, address(asset));
+
+        // must drop BPT index as well
+        exitTokenIndex = _getBPTIndex() < exitTokenIndex ? exitTokenIndex - 1 : exitTokenIndex;
+        // 0 = EXACT_BPT_IN_FOR_ONE_TOKEN_OUT
```
jparklev: Maybe have this as a constant?

pkg/core/src/tests/adapters/AuraAdapter.tm.sol
```diff
@@ -0,0 +1,531 @@
...
+        divider.initSeries(address(oAdapter), maturity, msg.sender);
+
+        // Mint some stake to sponsor Series
+        deal(AddressBook.DAI, divider.periphery(), STAKE_SIZE);
+
+        // Periphery approves divider to pull stake to sponsor series
+        vm.prank(divider.periphery());
+        ERC20(AddressBook.DAI).approve(address(divider), STAKE_SIZE);
+
+        // Open can open sponsor window
```
jparklev:
```suggestion
        // Opener can open sponsor window
```

pkg/core/src/tests/adapters/AuraAdapter.tm.sol
```diff
@@ -0,0 +1,531 @@
...
+        // Approvals
+        underlying.approve(address(adapter), type(uint256).max);
+        target.approve(address(adapter), type(uint256).max);
+
+        // Full cycle
+        adapter.unwrapTarget(adapter.wrapUnderlying(wrapAmt));
+
+        uint256 postbal = underlying.balanceOf(address(this));
+
+        // NOTE: we can't expect that prebal == postbal because when we do a `wrapUnderlying` we are
```
jparklev, line 390: Should we expect them within a certain percentage of each other?

pkg/core/src/tests/adapters/AuraAdapter.tm.sol
```diff
@@ -0,0 +1,531 @@
...
+            sender: receiver,
+            fromInternalBalance: false,
+            recipient: receiver,
+            toInternalBalance: false
+        });
+
+        amountOut = balancerVault.swap(request, funds, minAccepted, type(uint256).max);
+    }
+
+    function _increaseScale() internal {
```
jparklev: Nice, this seems like a good way to do it. Would this work for a smaller swap as well?

pkg/core/src/tests/adapters/AuraAdapter.tm.sol
jparklev, line 1: ooh i can comment on a file now. ty github

pkg/core/src/tests/adapters/AuraAdapter.tm.sol
```diff
@@ -0,0 +1,531 @@
...
+// Test references
+import { AddressBook } from "@sense-finance/v1-utils/addresses/AddressBook.sol";
+import { Constants } from "../test-helpers/Constants.sol";
+import { DateTimeFull } from "../test-helpers/DateTimeFull.sol";
+import { ForkTest } from "@sense-finance/v1-core/tests/test-helpers/ForkTest.sol";
+import { MockFactory } from "../test-helpers/mocks/MockFactory.sol";
+
+import "hardhat/console.sol";
+
+// AURA: HOW YIELD WORKS?
```
jparklev, line 36: 😵 thanks for this writeup, i think i'm like 75% of the way there on this mechanism

pkg/core/src/adapters/implementations/aura/AuraAdapter.sol
```diff
@@ -0,0 +1,128 @@
...
+        bool join
+    ) public override(BaseAdapter, Crops) {
+        super.notify(_usr, amt, join);
+    }
+
+    function _claimRewards() internal override {
+        address aToken = address(AuraVaultWrapper(target).aToken());
+        uint256 extraRewardsLen = IRewards(aToken).extraRewardsLength();
+        if (extraRewardsLen == 0) {
+            IRewards(aToken).getReward(address(target), false);
```
jparklev, line 107: We can do `false` here because we know there are no extra rewards – and this will save gas (which would otherwise be spent looping through an empty list in the `getReward` functions). Is that all correct?

pkg/core/src/adapters/implementations/aura/AuraVaultWrapper.sol
```diff
@@ -0,0 +1,296 @@
...
+    function depositFromBPT(uint256 bptIn, address receiver) external returns (uint256 shares) {
+        // transfer BPT from user to here
+        ERC20(address(pool)).safeTransferFrom(msg.sender, address(this), bptIn);
+
+        _mint(receiver, bptIn);
+
+        // lock BPT into Aura Vault
+        auraBooster.deposit(auraPID, bptIn, true);
+
+        //TODO: modify event for DepositFromBPT
```
jparklev: Or a new event?

pkg/core/src/adapters/implementations/aura/AuraVaultWrapper.sol
```diff
@@ -0,0 +1,296 @@
...
+
+        bptOut = aToken.withdraw(shares, address(this), address(this));
+
+        ERC20(address(pool)).safeTransfer(receiver, bptOut);
+    }
+
+    /* ========== ExtractableReward overrides ========== */
+
+    function _isValid(address _token) internal virtual override returns (bool) {
+        return (_token != address(aToken)); // TODO: aToken is anyways non-transferable, maybe this check is not needed?
```
jparklev: Will that always be the case? If so then I agree

pkg/core/src/adapters/implementations/aura/AuraVaultWrapper.sol
```diff
@@ -0,0 +1,296 @@
...
+        // max amounts in
+        uint256 amountsLength = _getBPTIndex() < type(uint256).max ? poolAssets.length - 1 : poolAssets.length;
+
+        uint256[] memory amountsIn = new uint256[](amountsLength);
+        uint256 index = find(poolAssets, address(asset)); // find index of underlying
+        // uint256 indexSkipBPT = index > _getBPTIndex() ? index - 1 : index;
+        amountsIn[index] = amt;
+
+        // encode user data
+        uint256 minBptOut = 0; // TODO: fine to be 0?
```
jparklev: This should be fine as long as users are using the peripheries, which have slippage protection built-in

pkg/core/src/adapters/implementations/aura/AuraVaultWrapper.sol
```diff
@@ -0,0 +1,296 @@
...
+        request = IVault.JoinPoolRequest({
+            assets: poolAssets,
+            maxAmountsIn: amountsIn,
+            userData: userData,
+            fromInternalBalance: false
+        });
+    }
+
+    function _redeemFromBalancer(uint256 lpAmt) internal virtual returns (uint256) {
+        uint256 balanceBefore = asset.balanceOf(address(this));
```
jparklev, line 225: Are we expecting tokens to be sitting in this contract?

jparklev (commented):
Man, this one was pretty non-trivial (where's the meme of alan showing you the portal with both of you crying hehe). But good job here, seems like we're quite close. I saw a few ToDos still, and I had a few questions, but it was mostly small things – think we're looking good overall

pkg/core/src/adapters/abstract/extensions/Crops.sol
fedealconada: nvm, i thought for a moment that the reconcile was linked to that bug Space had the we then fixed, but it had nothing to do with it

pkg/deployments/hardhat.config.js
fedealconada: agree

pkg/core/src/tests/adapters/AuraAdapter.tm.sol
fedealconada: right! we don't, i'll remove it

pkg/core/src/adapters/implementations/aura/AuraVaultWrapper.sol
fedealconada: i took this from a Pendle library actually, but since it was only one method we are using, i've added it here. I've just made a gas diff between moving that to a library vs leaving it as it is and, by little, but seems like the latter is more gas efficient 🤷🏼‍♂️
image

pkg/core/src/adapters/implementations/aura/AuraVaultWrapper.sol
fedealconada: i actually didn't know, but saw this message in one of Pendle's contract. I don't think we will have that use case but wanted to leave it there just in case we then re-use this contract to wrap some other aura vault that somehow has that condition...

pkg/core/src/tests/adapters/AuraAdapter.tm.sol
fedealconada: yes! do you want me somehow to change the _increaseScale to receive a param that determines the size of the swap?

pkg/core/src/tests/adapters/AuraAdapter.tm.sol
fedealconada: i got that from their Discord and found it useful as a summary of how yield works

pkg/core/src/adapters/implementations/aura/AuraAdapter.sol
fedealconada: exactly that

pkg/core/src/adapters/implementations/aura/AuraVaultWrapper.sol
fedealconada: yeah lol, that's what i meant actually, adding it now

fedealconada: waht do you think if this method does not return anything? since we now it's the ratio is always 1:1 it's obvious that the return value is the same as `bptIn`

fedealconada: same for `withdrawToBPT`

pkg/core/src/adapters/implementations/aura/AuraVaultWrapper.sol
fedealconada: seems like! https://github.com/aurafinance/convex-platform/blob/lite/contracts/contracts/BaseRewardPool4626.sol#L253

pkg/core/src/adapters/implementations/aura/AuraVaultWrapper.sol
fedealconada: right, but if they call directly either the wrapper or the adapter, you're not protected against slippage...

pkg/core/src/adapters/implementations/aura/AuraVaultWrapper.sol
fedealconada: yes, the big aura position will be sitting in this contract while rewards, when collected (e.g `notify()`), are transferred to this contract but immediately (on the same tx) extracted to the adapter

pkg/core/src/adapters/implementations/aura/AuraVaultWrapper.sol
jparklev: Ahh got it. Yea, since it's only one thing + it's cheaper + contract size isn't a big issue, I think it's fine as is 👍

pkg/core/src/adapters/implementations/aura/AuraVaultWrapper.sol
jparklev: OK, cool. Yea, don't think we will run into this, then, but doesn't hurt

pkg/core/src/tests/adapters/AuraAdapter.tm.sol
jparklev: I don't think you need to unless you think it'd be good to test with different values more particular to each test(?) Just wanted to make sure our assumptions held for even very minor increases in scale (assuming scale still increases for v small swaps)

pkg/core/src/adapters/implementations/aura/AuraVaultWrapper.sol
jparklev: Yes, since this is an out-of-spec function anyway, I think that's fine 👌 (is there a return value from `auraBooster.deposit`?)

pkg/core/src/adapters/implementations/aura/AuraVaultWrapper.sol
jparklev: Will that ever be required? I know it's the same with the RLV, so we just always encourage ppl to use the Periphery, since there's no need to use the RLV directly

pkg/core/src/tests/adapters/AuraAdapter.tm.sol
```diff
@@ -0,0 +1,539 @@
...
+        _issue(address(adapter), maturity, tBal);
+
+        uint256 scale = adapter.scale(); // current scale
+
+        // roll to period finish
+        vm.warp(IRewards(address(target.aToken())).periodFinish() + 100);
+
+        // distribute rewards
+        uint256 AURA_PID = 15; // 15 is the aura PID
+        IBooster(AddressBook.AURA_BOOSTER).earmarkRewards(AURA_PID);
```
abylaw, line 418: could we check whether the reward amount is proportional to the amount of `target` the user is holding?

pkg/core/src/tests/adapters/AuraAdapter.tm.sol
```diff
@@ -0,0 +1,539 @@
...
+        ERC20(AddressBook.B_RETH_STABLE).approve(address(target), bptAmt);
+
+        // deposit from BPTs (converts BPTs into wrapped token)
+        vm.expectEmit(true, true, true, true);
+        emit DepositFromBPT(address(this), address(this), bptAmt);
+
+        target.depositFromBPT(bptAmt, address(this));
+        assertEq(target.balanceOf(address(this)), bptAmt);
+        assertEq(ERC20(AddressBook.B_RETH_STABLE).balanceOf(address(this)), 0);
+
```
abylaw, line 610: it'd be nice to add tests to check that BPT have been deposited into the Aura Booster and the Wrapper now holds aura tokens? (as well as checking balances on withdraw)

abylaw (approved):
lgtm~ some suggestions for more complete tests!

pkg/core/src/tests/adapters/AuraAdapter.tm.sol
fedealconada: added!

pkg/core/src/tests/adapters/AuraAdapter.tm.sol
fedealconada: i can quickly add it!

pkg/core/src/adapters/implementations/aura/AuraVaultWrapper.sol
fedealconada: it returns a boolean indicating whether the deposit succeeded or not which, actually, i could add to check everything went well

pkg/core/src/adapters/implementations/aura/AuraVaultWrapper.sol
fedealconada: nope, not required

pkg/core/src/tests/adapters/AuraAdapter.tm.sol
fedealconada: yes!

pkg/deployments/tasks/20230321-aura/index.js
```diff
@@ -0,0 +1,284 @@
...
+      console.log("\n-------------------------------------------------------");
+      await verifyOnEtherscan("AuraVaultWrapper");
+    }
+
+    console.log(`AuraVaultWrapper deployed to ${targetAddress}\n`);
+    return new ethers.Contract(targetAddress, abi, deployerSigner);
+  };
+
+  const _deployRLV = async (adapter, rlvFactory) => {
+    // TODO: we don't need a rewards distributor for this RLV, right?
```
jparklev: Yes! Is there a reason we wouldn't?

jparklev (approved):
Other than the script TODO, recent changes LGTM (in addition to @abylaw's comments ✨ )

pkg/deployments/tasks/20230321-aura/index.js
fedealconada: I've added this but I've realised that we can't have multiple rewards distributor on one RLV, right? So i've created a `MultiRewardsDistributor.sol`, mind taking a look at it? or is there any other alternative? @jparklev cc @abylaw

pkg/core/src/external/MultiRewardsDistributor.sol
```diff
@@ -0,0 +1,112 @@
...
+    /// @param _account The address of the claimer.
+    /// @param _claimable The overall claimable amount of token rewards.
+    /// @param _proof The merkle proof that validates this claim.
+    function claim(
+        ERC20 _reward,
+        address _account,
+        uint256 _claimable,
+        bytes32[] calldata _proof
+    ) external {
+        bytes32 candidateRoot = MerkleProof.processProof(_proof, keccak256(abi.encodePacked(_account, _claimable)));
```
jparklev: I think the `_reward` token should be in the `keccack` here, right? Otherwise the `claimable` balance isn't associated with any token in particular, and a user could claim some `claimable` balance for any of the tokens? Or am I missing smtg?

jparklev (commented):
Just one question on the `claim` logic

pkg/deployments/tasks/20230321-aura/index.js
jparklev: Another option is to have the `recipient` address be some contract that can only move tokens to multiple independent reward distributors (e.g. it has functions like `moveAura` that moves all of the aura tokens in the contract to an aura rd).

However, having a single multi is better overall IMO

pkg/core/src/external/MultiRewardsDistributor.sol
fedealconada: ohhh, good catch! yes indeed!

jparklev (approved):

pkg/core/src/adapters/implementations/aura/AuraAdapter.sol
```diff
@@ -13,6 +13,7 @@ import { Crops } from "../../abstract/extensions/Crops.sol";
 import { Errors } from "@sense-finance/v1-utils/libs/Errors.sol";
 import { ExtractableReward } from "../../abstract/extensions/ExtractableReward.sol";
 import { FixedMath } from "../../../external/FixedMath.sol";
+import "hardhat/console.sol";
```
jparklev, line 16: leftover?

pkg/core/src/tests/adapters/AuraAdapter.tm.sol
```diff
@@ -300,17 +301,20 @@ contract AuraAdapters is AuraAdapterTestHelper {
     }
 
     function testMainnetGetUnderlyingPrice() public {
-        assertEq(adapter.getUnderlyingPrice(), 1e18);
+        (, int256 rethPrice, , , ) = PriceOracleLike(RETH_ETH_PRICEFEED).latestRoundData();
+        // within 1%
+        uint256 rETHPrice = adapter.getUnderlyingPrice();
+        assertApproxEqAbs(rETHPrice, uint256(rethPrice), rETHPrice.fmul(0.050e18));
```
jparklev: This variable naming confused me for a sec! maybe `rETHPriceAdapter`?

pkg/core/src/external/ComposableStablePreview.sol
```diff
@@ -0,0 +1,543 @@
...
+    using ComposableStableMath for uint256;
+    using StablePoolUserData for bytes;
+    using FixedPoint for uint256;
+
+    struct ImmutableData {
+        address[] poolTokens;
+        address[] rateProviders;
+        uint256[] rawScalingFactors;
+        bool[] isExemptFromYieldProtocolFee;
+        //
```
jparklev: comment here?

pkg/core/src/adapters/implementations/aura/AuraVaultWrapper.sol
```diff
@@ -250,15 +263,25 @@ contract AuraVaultWrapper is ERC4626, ExtractableReward {
...
+        res.LP = address(pool);
+        res.noTokensExempt = NO_TOKENS_EXEMPT;
+        res.allTokensExempt = ALL_TOKENS_EXEMPT;
+        res.bptIndex = BPT_INDEX;
+        res.totalTokens = res.poolTokens.length;
 
         return abi.encode(res);
     }
 
+    function _getExemption() internal view virtual returns (bool[] memory res) {
```
jparklev: unused?

pkg/core/src/adapters/implementations/aura/AuraVaultWrapper.sol
```diff
@@ -193,16 +204,18 @@ contract AuraVaultWrapper is ERC4626, ExtractableReward {
         uint256 amountsLength = _getBPTIndex() < type(uint256).max ? poolAssets.length - 1 : poolAssets.length;
 
         uint256[] memory amountsIn = new uint256[](amountsLength);
+        uint256[] memory maxAmountsIn = new uint256[](poolAssets.length);
         uint256 index = find(poolAssets, address(asset)); // find index of underlying
-        amountsIn[index] = amt;
+        uint256 indexSkipBPT = index > _getBPTIndex() ? index - 1 : index;
```
jparklev, line 209: I'm not sure I understand why `maxAmountsIn` and `amountsIn` would use a different index, but must just be something I don't understand about this pool type

jparklev (approved):
Other than a few small comments, these updates look good to me. I don't have a full understanding of this pool type yet though, so i'd rely on extra testing to improve confidence. What are your thoughts on adding some of our basic 4626 tests, or at least the ones relevant to the functions implemented, to the vault wrapper?

And maybe adding something like our `testScaleIsExRate` test (we have for our 4626 adapter) to the `AuraAdapter` suite as well?

pkg/core/src/adapters/implementations/aura/AuraAdapter.sol
fedealconada: lol, yeah

fedealconada:
> Other than a few small comments, these updates look good to me. I don't have a full understanding of this pool type yet though, so i'd rely on extra testing to improve confidence. What are your thoughts on adding some of our basic 4626 tests, or at least the ones relevant to the functions implemented, to the vault wrapper?
> 
> And maybe adding something like our `testScaleIsExRate` test (we have for our 4626 adapter) to the `AuraAdapter` suite as well?

yeah, that sounds reasonable!

pkg/core/src/tests/adapters/AuraAdapter.tm.sol
fedealconada: hm... i don't know, it;s the price of the underlying, the underlying is `rETH`, then `rETHPrice` makes sense to me, am i missing anything? i don't mind if it's `rETHPriceAdapter` though

pkg/core/src/external/ComposableStablePreview.sol
fedealconada: no comments xD

pkg/core/src/tests/adapters/AuraAdapter.tm.sol
jparklev: Oh, just that `rETHPrice` and `rethPrice` are very similar

fedealconada: ohhhh, trueeee

fedealconada:
> > Other than a few small comments, these updates look good to me. I don't have a full understanding of this pool type yet though, so i'd rely on extra testing to improve confidence. What are your thoughts on adding some of our basic 4626 tests, or at least the ones relevant to the functions implemented, to the vault wrapper?
> > And maybe adding something like our `testScaleIsExRate` test (we have for our 4626 adapter) to the `AuraAdapter` suite as well?
> 
> yeah, that sounds reasonable!

pushed this!

pkg/core/src/tests/adapters/ERC4626Adapter.tm.sol
```diff
@@ -310,6 +326,14 @@ contract ERC4626Adapters is ForkTest {
             return;
         }
 
+        if (
+            keccak256(abi.encodePacked(target.symbol())) ==
+            keccak256(abi.encodePacked("w-aurawstETH-rETH-sfrxETH-BPT-vault"))
+        ) {
+            IAuraAdapterTestHelper(msg.sender).increaseScale(uint64(amt));
```
jparklev: Who is `msg.sender` in this case, if this is being called from within this contract? (presumably whoever is calling the external test function, which, is that the test runner itself?)

pkg/core/src/tests/adapters/AuraAdapter.tm.sol
```diff
@@ -629,7 +628,10 @@ contract AuraAdapters is AuraAdapterTestHelper {
 
     // override testMainnetWrapUnwrap to avoid swapping very small amounts on balancer
     function testMainnetWrapUnwrap(uint256 wrapAmt) public override {
-        vm.assume(wrapAmt > 1e6);
+        vm.assume(wrapAmt > 1e10);
+        if (wrapAmt < 1e10) {
```
jparklev: If we have the `assume`, we don't need this extra check, right?

jparklev (approved):
Question about how `_increaseVaultYield` works, but overall the recent commits LGTM

pkg/core/src/tests/adapters/AuraAdapter.tm.sol
fedealconada: no haha, this is me just trying to debug why the build is failing... but will remove when i work this out

pkg/core/src/tests/adapters/ERC4626Adapter.tm.sol
fedealconada: this is actually a vestige from when I added this tests differently. It can be removed!

PR #346, opened 2023-04-05
fix: refund excess tokens

fedealconada:
fixes https://github.com/sherlock-audit/2023-03-sense-judging/issues/29


pkg/core/src/Periphery.sol
```diff
@@ -446,6 +482,10 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         permit2.permitTransferFrom(permit.msg, sigs, msg.sender, permit.sig);
         amt = _fromTarget(adapter, divider.combine(adapter, maturity, uBal), quote);
         _transfer(quote.buyToken, receiver, amt);
+
+        // refund any remaining underlying to receiver
+        ERC20 underlying = ERC20(Adapter(adapter).underlying());
+        _transfer(underlying, receiver, underlying.balanceOf(address(this)));
```
jparklev: Why not use `quote.sellToken` rather than calling the `underlying` getter on `adapter` here?

jparklev (commented):
Thoughts on adding these excess refunds to `_toTarget` and `_fromTarget` directly?

pkg/core/src/Periphery.sol
fedealconada: yeah, good call!

fedealconada: changed!

jparklev: awesome!

jparklev:
> Thoughts on adding these excess refunds to `_toTarget` and `_fromTarget` directly?

gentle bump on this @fedealconada ^

fedealconada:
> > Thoughts on adding these excess refunds to `_toTarget` and `_fromTarget` directly?
> 
> gentle bump on this @fedealconada ^

sorry! didn't see this. definitely! would be more optimal :) just pushed!

pkg/core/src/Periphery.sol
```diff
@@ -1010,7 +1023,7 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         address receiver,
         uint256 amt
     ) internal {
-        address(token) == ETH ? payable(receiver).transfer(amt) : token.safeTransfer(receiver, amt);
+        if (amt > 0) address(token) == ETH ? payable(receiver).transfer(amt) : token.safeTransfer(receiver, amt);
```
IAm0x52: The should be call rather than transfer for ETH

IAm0x52 (approved):
Fixes look good. Excess sell tokens are now returned inside _toTarget and _fromTarget

IAm0x52 (approved):
Fix looks good. ETH transfer changed to call

PR #347, opened 2023-04-05
fix: _transferFrom on sponsorSeries

fedealconada:
fixes https://github.com/sherlock-audit/2023-03-sense-judging/issues/36

### DIFF
--- a/pkg/core/src/Periphery.sol
+++ b/pkg/core/src/Periphery.sol
@@ -87,6 +87,7 @@ contract Periphery is Trust, IERC3156FlashBorrower {
     struct SwapQuote {
         ERC20 sellToken;
         ERC20 buyToken;
+        uint256 amount;
         address spender;
         address payable swapTarget;
         bytes swapCallData;
@@ -121,7 +122,7 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         SwapQuote calldata quote
     ) external payable returns (address pt, address yt) {
         (, address stake, uint256 stakeSize) = Adapter(adapter).getStakeAndTarget();
-        if (address(quote.sellToken) != ETH) _transferFrom(permit, stake, stakeSize);
+        if (address(quote.sellToken) != ETH) _transferFrom(permit, address(quote.sellToken), quote.amount);
         if (address(quote.sellToken) != stake) _fillQuote(quote);
 
         // Approve divider to withdraw stake assets
--- a/pkg/core/src/tests/Periphery.t.sol
+++ b/pkg/core/src/tests/Periphery.t.sol
@@ -41,6 +41,7 @@ contract PeripheryTest is TestHelper {
     function testSponsorSeries() public {
         uint256 maturity = getValidMaturity(2021, 10);
 
+        vm.expectEmit(false, false, false, false);
         emit SeriesSponsored(address(adapter), maturity, address(this));
 
         Periphery.PermitData memory data = generatePermit(bobPrivKey, address(periphery), address(stake));
--- a/pkg/core/src/tests/Periphery.t.sol
+++ b/pkg/core/src/tests/Periphery.t.sol
@@ -41,8 +41,8 @@ contract PeripheryTest is TestHelper {
     function testSponsorSeries() public {
         uint256 maturity = getValidMaturity(2021, 10);
 
-        vm.expectEmit(false, false, false, false);
-        emit SeriesSponsored(address(adapter), maturity, address(this));
+        vm.expectEmit(true, true, true, true);
+        emit SeriesSponsored(address(adapter), maturity, bob);
 
         Periphery.PermitData memory data = generatePermit(bobPrivKey, address(periphery), address(stake));
         vm.prank(bob);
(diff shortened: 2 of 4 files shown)


pkg/core/src/Periphery.sol
```diff
@@ -87,6 +87,7 @@ contract Periphery is Trust, IERC3156FlashBorrower {
     struct SwapQuote {
         ERC20 sellToken;
         ERC20 buyToken;
+        uint256 amount;
```
fedealconada, line 90: this would only be used on `sponsorSeries` and when not using permit but normal approvals. curious on what you guys think? should we rather send an `amount` param on `function sponsorSeries()`?

jparklev: Hmm it's not used for permit approvals? I see `quote.amount` is passed into `_transferFrom`, which I think would be used for normal transfers or permit? (think i'm missing smtg, so pls set me straight!)

fedealconada: My bad, you're right, somehow I thought I was ignoring that param on permit :)

pkg/core/src/tests/Periphery.t.sol
```diff
@@ -41,6 +41,7 @@ contract PeripheryTest is TestHelper {
     function testSponsorSeries() public {
         uint256 maturity = getValidMaturity(2021, 10);
 
+        vm.expectEmit(false, false, false, false);
```
jparklev: Any reason not to match check on some of the topics here?

pkg/core/src/tests/Periphery.tm.sol
```diff
@@ -286,7 +286,60 @@ contract PeripheryMainnetTests is PeripheryTestHelper {
...
+        });
+        vm.prank(bob);
+        (address pt, address yt) = periphery.sponsorSeries(address(cadapter), maturity, false, data, quote);
+
+        // Check pt and yt deployed
+        assertTrue(pt != address(0));
+        assertTrue(yt != address(0));
+
+        // Check PT and YT onboarded on PoolManager (Fuse)
+        // (PoolManager.SeriesStatus status, ) = PoolManager(poolManager).sSeries(address(cadapter), maturity);
```
jparklev, line 338: Vestige

pkg/core/src/tests/Periphery.tm.sol
```diff
@@ -286,7 +286,60 @@ contract PeripheryMainnetTests is PeripheryTestHelper {
         // assertTrue(status == PoolManager.SeriesStatus.QUEUED);
     }
 
-    function testMainnetSponsorSeriesFromToken() public {
+    function testMainnetSponsorSeriesFromTokenWithApprove() public {
+        // Roll to Feb-08-2023 09:12:23 AM +UTC where we have a real adapter (wstETH adapter)
```
jparklev: Why do we need to do this if we're deploying our own adapter?

pkg/core/src/tests/Periphery.tm.sol
```diff
@@ -320,6 +373,7 @@ contract PeripheryMainnetTests is PeripheryTestHelper {
 
         Periphery.PermitData memory data = generatePermit(bobPrivKey, address(periphery), AddressBook.USDC);
         Periphery.SwapQuote memory quote = Periphery.SwapQuote({
+            amount: 1000000e6, // we a large amount of USDC (which will be converted into stake token)
```
jparklev: This might be good as an excess refund test as well? from https://github.com/sense-finance/sense-v1/pull/346/files

jparklev (approved):
Few smol questions around testing, but this looks like it solves the issue to me

pkg/core/src/tests/Periphery.tm.sol
fedealconada: Comment is wrong. We fix the block number to make sure that we use a valid 0x `swapCallData`.

pkg/core/src/tests/Periphery.t.sol
fedealconada: Nope, adding it!

pkg/core/src/tests/Periphery.tm.sol
fedealconada: I left this on purpose since we've decided not to remove the code in case we wanted to re-use it later on. Many things related to the pool manager are only commented and not removed. But happy to remove them, would you prefer that?

pkg/core/src/tests/Periphery.tm.sol
fedealconada: yeah, i was actually gonna do that but, since the repos are separate, we don't have the refund excess feature on this PR. if you want, I could add this tests once we merge all these PRs?

pkg/core/src/tests/Periphery.tm.sol
jparklev: Ahh I see, I don't think we'll have a a pool manager any time soon, but I think it's fine to leave in as long as its intentional

pkg/core/src/tests/Periphery.tm.sol
jparklev: > if you want, I could add this tests once we merge all these PRs?

Sure, if it's not too much extra work ~

fedealconada: of course not! i've created a [task](https://linear.app/sense-finance/issue/GEN-2229/add-refund-excess-tests) already so we don't forget

IAm0x52 (approved):
Fix looks good. _transferFrom now correctly pulls the sellToken instead of stake

PR #350, opened 2023-04-05
fix: add deadline to balancer swaps

fedealconada:
fixes https://github.com/sherlock-audit/2023-03-sense-judging/issues/55

notes:

adding `deadline` as a new param triggered some `stack too deep` error here and there, so most of the changes you will see are just workaround to bypass that.

downside: some added gas costs (see below the gas diff)

if you see some other way of making this pass and reducing gas a bit more, please let me know!

image


pkg/core/src/Periphery.sol
```diff
@@ -282,22 +299,24 @@ contract Periphery is Trust, IERC3156FlashBorrower {
             IPermit2.PermitTransferFrom(IPermit2.TokenPermissions(ERC20(address(0)), 0), 0, 0),
             "0x"
         );
-        amt = this._swapYTsForTarget(msg.sender, adapter, maturity, ytBal, permit);
+        uint256 deadline = block.timestamp + 14400; // 4 hours
```
jparklev: Why 4 hours?

pkg/core/src/Periphery.sol
```diff
@@ -747,23 +765,24 @@ contract Periphery is Trust, IERC3156FlashBorrower {
     /// @notice Initiates a flash loan of Target, swaps target amount to PTs and combines
     /// @param adapter adapter
     /// @param maturity maturity
-    /// @param ytBalIn YT amount the user has sent in
-    /// @param amountToBorrow target amount to borrow
+    /// @param data abi.encoded data:
+    /// - YT amount the user has sent in
+    /// - target amount to borrow
```
jparklev, line 772: & deadline?

pkg/core/src/Periphery.sol
```diff
@@ -92,6 +92,18 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
     }
 
+    struct AddLiquidityData {
+        uint256 minAccepted; // Min accepted amount of Target (from the sell of YTs)
+        uint256 minBptOut; // Minimum BPT the user will accept out for this transaction
+        uint256 deadline;
+    }
+
+    struct RemoveLiquidityData {
+        uint256 minAccepted; // min accepted when swapping PTs to underlying (only used when removing liquidity on/after maturity)
```
jparklev: Nit: capitalization consistence with `AddLiquidityData`

pkg/core/src/tests/Periphery.t.sol
```diff
@@ -630,49 +631,74 @@ contract PeripheryTest is TestHelper {
...
+        uint256 maturity = getValidMaturity(2021, 10);
+
+        Periphery.PermitData memory data = generatePermit(bobPrivKey, address(periphery), address(stake));
+        vm.prank(bob);
+        periphery.sponsorSeries(address(adapter), maturity, true, data, _getQuote(address(stake), address(stake)));
+
+        // add liquidity to mockBalancerVault
+        addLiquidityToBalancerVault(maturity, 1000e18);
+
+        // roll to deadline + 1 so the swap should fail
```
jparklev: Should there be an `expectRevert`?

pkg/core/src/Periphery.sol
```diff
@@ -92,6 +92,18 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         bytes swapCallData;
     }
 
+    struct AddLiquidityData {
```
jparklev: Consider `AddLiquidityParams`, copying uni's pattern here: https://github.com/Uniswap/v3-core/blob/d8b1c635c275d2a9450bd6a78f3fa2484fef73eb/contracts/UniswapV3Pool.sol#L306

jparklev (approved):
Small observations and a question about the 4 hours hardcoded, but otherwise looks reasonable to me, given that we don't want to make too many changes after everything has been audited

pkg/core/src/Periphery.sol
fedealconada: ohh, haha, i forgot to bring this up for discussion! in this case, we need to hardcode a deadline, so i was not sure what would it be reasonable? 1 hour? cc @stevenvaleri

pkg/core/src/tests/Periphery.t.sol
fedealconada: good catch, this is a vestige from when i started testing but local tests use a `MockSpace.sol` and we are not checking for the deadline there. I'm removing this line.

pkg/core/src/Periphery.sol
jparklev: Gotcha, do we even need assign a non-zero deadline here, if the swap is going to be executed within the same block?

fedealconada: oh, so this fix is actually wrong, because `block` will be the block when the tx gets executed and not the block when the tx was submitted, right? so we don't actually have a solution for this...

jparklev: Correct. Our problem here is that we can't add a `deadline` param to this function because we wanted to keep the interface stable for auto roller compatibility, right? If that's the case, I think it's OK to not handle deadline support for this function and note in it in the comments, and then we'll use `swapYTs` on the frontend

fedealconada: pushed!

PR #351, opened 2023-04-05
feat: alt implementation without structs as params

stevenvaleri:
An alternate implementation of the deadline without using structs in the params for external functions


fedealconada:
Nice! IMHO, it looks cleaner. Seems like it consumes a little bit more gas though:

image

pkg/core/src/Periphery.sol
```diff
@@ -332,7 +320,9 @@ contract Periphery is Trust, IERC3156FlashBorrower {
     /// @param adapter Adapter address for the Series
     /// @param maturity Maturity date for the Series
     /// @param amt Amount to provide
-    /// @param data AddLiquidityData struct with data for the addLiquidity call
+    /// @param minAccepted Min accepted amount of Target (from the sell of YTs)
+    /// @param minBptOut Minimum BPT the user will accept out for this transaction
+    /// @param deadline Time in seconds when this transaction must be executed before
```
stevenvaleri, line 325: Is this param defined elsewhere? If so let's replace the definition to keep it consistent

pkg/core/src/Periphery.sol
```diff
@@ -375,7 +372,9 @@ contract Periphery is Trust, IERC3156FlashBorrower {
     /// @param adapter Adapter address for the Series
     /// @param maturity Maturity date for the Series
     /// @param lpBal Balance of LP tokens to provide
-    /// @param data RemoveLiquidityData struct with data for the removeLiquidity call
+    /// @param minAmountsOut minimum accepted amounts of PTs and Target given the amount of LP shares provided
```
stevenvaleri, line 375: Should we re-order minAmountsOut & minAccepted to match add liquidity

pkg/core/src/Periphery.sol
```diff
@@ -713,53 +720,67 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
-        address pt = divider.pt(adapter, maturity);
-        uint256 _ptBal;
-        (tBal, _ptBal) = _removeLiquidityFromSpace(
-            pool.getPoolId(),
-            pt,
+        {
+            BalancerPool pool = BalancerPool(spaceFactory.pools(adapter, maturity));
+            _transferFrom(permit, address(pool), lpBal);
+            poolId = pool.getPoolId();
+        }
```
stevenvaleri, line 749: blocking makes the pool variable clear from the stack once executed, clearing space for one

pkg/core/src/Periphery.sol
```diff
@@ -713,53 +720,67 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
         uint256 maturity,
         uint256 lpBal,
-        RemoveLiquidityData memory data,
+        uint256[] memory minAmountsOut,
+        uint256 minAccepted,
+        uint256 deadline,
         bool swapPTs,
         address receiver,
         PermitData calldata permit
     ) internal returns (uint256 tBal, uint256 ptBal) {
```
stevenvaleri, line 740: I think we leave the return variables even though they are overridden to maintain consistency with the rest of the contract even though they are overridden. Removing the names may be safer.

stevenvaleri:
Given Significant Gas increase of Remove Liquidity, I'm hesitant of this approach. The worst case doesn't have that much of a difference, so if the worst case is the most common path, then maybe it would make more sense.

Gas Comparison 

Add Liquidity - slight improvement
This branch --- 234987 | 165554 | 388553
Original ------- 235042 | 165620 | 388618

Remove Liquidity - significantly worse
This branch --- 126844 | 129071 | 168028
Original ------- 120990 | 111369 | 167985  

This branch
image

Original
image

pkg/core/package.json
```diff
@@ -25,7 +25,8 @@
         "script:goerli": "cp ../../.env ./.env && time forge script --rpc-url https://eth-goerli.g.alchemy.com/v2/Ef0GN4VITrvOPk8qVRASZLqoCyMWo_Pd --chain-id 5 --optimize --optimizer-runs 500 -vvv --broadcast --verify",
         "coverage": "time forge coverage --match-path \"**/*.{t.sol,tm.sol}\" --no-match-contract \"ERC4626StdTest\" --report lcov",
         "gas-snapshot": "forge snapshot --no-match-path \"*.tm*\"",
-        "gas-snapshot:diff": "forge snapshot --no-match-path \"*.tm*\" --diff"
+        "gas-snapshot:diff": "forge snapshot --no-match-path \"*.tm*\" --diff",
+        "gas-report": "forge test --gas-report --match-path \"**/*.{t.sol,tm.sol}\" --no-match-contract \"ERC4626StdTest\""
```
jparklev, line 29: Nice, we should add this regardless of whether we merge this PR or not

pkg/core/src/Periphery.sol
jparklev: > Removing the names may be safer

For my own curiosity, could u expand on your thoughts on this? Safer in that we're less likely to accidentally flip them?

pkg/core/src/Periphery.sol
jparklev: ^ Another piece we may be able to pull into the original PR either way @fedealconada?

jparklev (commented):
Woot! A Steven PR! V exciting, and appreciate being able to see an alternative. Haven't given it too deep of a look yet, but the `localVars` stuff makes me think you have some experience with stack too deep errors before hah

In any case, your conclusion makes sense to me here https://github.com/sense-finance/sense-v1/pull/351#issuecomment-1499689094, and it seems like there are some pieces we could pull in regardless of whether or not we take the whole impl

pkg/core/src/Periphery.sol
stevenvaleri: I think in this implementation it is safe. The returned values override the named returns. But having two named, un-used variables, with nearly identical names of used variables is a little uncomfortable. It introduces future risk of these being used or set inappropriately in the future and returning one value when you intend to return another.

pkg/core/src/Periphery.sol
fedealconada: hmm.. it's the `deadline` we send to Balancer. Their definition is "The UNIX timestamp at which our trade must be completed by - if the transaction is confirmed after this time then the transaction will fail."
You say we should use that definition?

pkg/core/src/Periphery.sol
fedealconada: completely!

pkg/core/package.json
fedealconada: added this on [original PR](https://github.com/sense-finance/sense-v1/pull/350/files#diff-8891cee254f4a23b9817fab2d5f63f2ccb626e75e29f35e1650c579f4ce11307R27)

pkg/core/src/Periphery.sol
fedealconada: added this on [original PR](https://github.com/sense-finance/sense-v1/pull/350/commits/0c89e3512da4a1c856cb791e08cb3154d795dc12#diff-64da87194dd7275be5f56cebbe0684684471e7df8ec9d2385fc9f322aa4a7818R98)

pkg/core/src/Periphery.sol
fedealconada: added this on [original PR](https://github.com/sense-finance/sense-v1/pull/350/commits/0c89e3512da4a1c856cb791e08cb3154d795dc12#diff-64da87194dd7275be5f56cebbe0684684471e7df8ec9d2385fc9f322aa4a7818R728)

fedealconada:
> Woot! A Steven PR! V exciting, and appreciate being able to see an alternative. Haven't given it too deep of a look yet, but the `localVars` stuff makes me think you have some experience with stack too deep errors before hah
> 
> In any case, your conclusion makes sense to me here [#351 (comment)](https://github.com/sense-finance/sense-v1/pull/351#issuecomment-1499689094), and it seems like there are some pieces we could pull in regardless of whether or not we take the whole impl

Same! Nice to see this PR :) 🎉

jparklev:
Closing as the other deadline PR https://github.com/sense-finance/sense-v1/pull/350 has been merged, and compatible bits from this pass have been incorporated

PR #352, opened 2023-04-06
fix: swap PTs when settled and redeem restricted

fedealconada:
fixes https://github.com/sherlock-audit/2023-03-sense-judging/issues/32


jparklev (approved):
LGTM. Matches my understanding of what we'd talked about

IAm0x52 (approved):
Fix looks good. Periphery will swap instead of redeeming if redeem is restricted

PR #354, opened 2023-04-21
feat: move external balanceOf calls to an internal function

stevenvaleri:
## Motivation

Reduce the Periphery Contract Size

## Solution

move all external ERC20.balanceOf() calls to be called from an internal function. As is, the larger code to perform this external call is repeated throughout the contract. by using an internal function, the code to move to an internal function is quite small and that will occur often, while the code for the external call will only occur once.

## Implementer Checklist

This is a relatively minor re-factor, with no expected changes in functionality. All existing tests work.

### DIFF
--- a/pkg/core/src/Periphery.sol
+++ b/pkg/core/src/Periphery.sol
@@ -132,12 +132,14 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         PermitData calldata permit,
         SwapQuote calldata quote
     ) external payable returns (address pt, address yt) {
-        (, address stake, uint256 stakeSize) = Adapter(adapter).getStakeAndTarget();
+        (, address stakeAddress, uint256 stakeSize) = Adapter(adapter).getStakeAndTarget();
         if (address(quote.sellToken) != ETH) _transferFrom(permit, address(quote.sellToken), quote.amount);
-        if (address(quote.sellToken) != stake) _fillQuote(quote);
+        if (address(quote.sellToken) != stakeAddress) _fillQuote(quote);
+
+        ERC20 stake = ERC20(stakeAddress);
 
         // Approve divider to withdraw stake assets
-        ERC20(stake).safeApprove(address(divider), stakeSize);
+        stake.safeApprove(address(divider), stakeSize);
 
         (pt, yt) = divider.initSeries(adapter, maturity, msg.sender);
 
@@ -152,13 +154,13 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         }
 
         // refund any excess stake assets
-        ERC20(stake).safeTransfer(msg.sender, ERC20(stake).balanceOf(address(this)));
+        stake.safeTransfer(msg.sender, _balanceOf(stake, address(this)));
 
         // refund any remaining quote.sellToken to receiver
         _transfer(
             quote.sellToken,
             msg.sender,
-            address(quote.sellToken) == ETH ? address(this).balance : quote.sellToken.balanceOf(address(this))
+            address(quote.sellToken) == ETH ? address(this).balance : _balanceOf(quote.sellToken, address(this))
         );
 
         emit SeriesSponsored(adapter, maturity, msg.sender);
@@ -217,7 +219,7 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         _transfer(
             quote.sellToken,
             receiver,
-            address(quote.sellToken) == ETH ? address(this).balance : quote.sellToken.balanceOf(address(this))
+            address(quote.sellToken) == ETH ? address(this).balance : _balanceOf(quote.sellToken, address(this))
         );
     }
 
@@ -803,7 +805,7 @@ contract Periphery is Trust, IERC3156FlashBorrower {
             abi.encode(adapter, uint256(maturity), ytBalIn, ytBalIn - acceptableError, deadline, true)
         );
         if (!result) revert Errors.FlashBorrowFailed();
-        tBal = target.balanceOf(address(this));
+        tBal = _balanceOf(target, address(this));
     }
 
     /// @notice Initiates a flash loan of Target, issues PTs/YTs and swaps the PTs to Target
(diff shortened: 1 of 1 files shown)


stevenvaleri:
I tried two other iterations of this change. The current PR both improves on gas and results in the smallest contract size.

One - the internal `balanceOf()` excepted an address rather than an ERC20
Two - the internal `balanceOf()` excepted an ERC20 and does not cache possible values
Three (current) - the internal `balanceOf()` excepted an ERC20 and caches where possible

image

pkg/core/src/Periphery.sol
```diff
@@ -931,13 +937,17 @@ contract Periphery is Trust, IERC3156FlashBorrower {
             userData: abi.encode(lpBal),
             toInternalBalance: false
         });
-        tBal = ERC20(target).balanceOf(address(this));
-        ptBal = ERC20(pt).balanceOf(address(this));
+
+        ERC20 targeErc20 = ERC20(target);
```
fedealconada: very smol thing: i noticed that while in some lines above you changed `stake` for `stakeAddress` so you could then define `ERC20 stake` you have decided here to keep the var name as is (e.g with the `target` param) and defined `ERC20 targetErc20` instead of renaming it to `targetAddress` and defining `ERC20 target`. 

Was it because you didn't want to rename the var name because it was a param?

Same applies for param `pt`.

fedealconada:
image

fedealconada (approved):
🥇
image

pkg/core/src/Periphery.sol
stevenvaleri: This is a good point. My main goal was to minimize changes, including param names - but as you point out this creates and extends some inconsistencies.

pretty much in functions where we have two variables holding an address and ERC20 for the same contract, we should have consistent naming for contracts and addresses. I think, ideally, this naming would also carry through the rest of the contract, but if we want to keep it to just functions where this overlap occurs, we can do that too.

The least disruptive approach would likely be renaming the ERC20s to end with `Token` and leaving the addresses with just the contract name.
+ using xyzToken would be consistent with the SwapQuote struct with the `buyToken` and `sellToken`
+ it is a little messy for the `ptToken` though - principaltokenToken

pkg/core/src/Periphery.sol
```diff
@@ -911,9 +914,12 @@ contract Periphery is Trust, IERC3156FlashBorrower {
             userData: abi.encode(liq.amounts, liq.minBptOut),
             fromInternalBalance: false
         });
-        lpBal = ERC20(address(pool)).balanceOf(receiver);
+
+        ERC20 poolToken = ERC20(address(pool));
```
jparklev, line 918: Perhaps we could set `pool` as an `ERC20` argument for this function (wrapping the type in `_computeIssueAddLiq`) to avoid having to define this variable? not 100% it'll be cheaper, but a possible micro op

pkg/core/src/Periphery.sol
```diff
@@ -132,12 +132,14 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         PermitData calldata permit,
         SwapQuote calldata quote
     ) external payable returns (address pt, address yt) {
-        (, address stake, uint256 stakeSize) = Adapter(adapter).getStakeAndTarget();
+        (, address stakeAddress, uint256 stakeSize) = Adapter(adapter).getStakeAndTarget();
```
jparklev: It's less-than-ideal given that we'd have to give up this nice property of using the actual abstract contract as an interface, but another approach here might be to create an `Adapter` interface for this contract (like we do with `SpaceFactoryLike`) and define the return value of `stake` as an `ERC20`. Your thoughts?

pkg/core/src/Periphery.sol
jparklev: > The least disruptive approach would likely be renaming the ERC20s to end with Token and leaving the addresses with just the contract name.

Fwiw this feels good to me, and the kludginess around `ptToken` seems like an OK hit for the moment

jparklev (approved):
> he current PR both improves on gas and results in the smallest contract size.

Yay! Where's my :kurt: emoji, it would be appropriate here. Removing the larger external call bytecode blocks 👍 

> the internal balanceOf() excepted an ERC20 and caches where possible

For my own understanding, the caching you're referring to here is of the `ERC20` tokens in memory, is that right?

Left two smol thoughts on additional optimization paths, but overall the changes here LGTM (pending resolution to the renaming thread) ~

pkg/core/src/Periphery.sol
stevenvaleri: since we need to get the poolId later in the function `pool.getPoolId()` we need it as a BalancePool.

I started to re-factor to pass the poolId and poolToken as and ERC20, but it causes some upstream stack-overflow issues. I can keep trying this, but I'd prefer to include it as a separate PR.

pkg/core/src/Periphery.sol
stevenvaleri: Generally, I don't love the idea of implicit casting through the `Like` interfaces and we would still need to cast it back to an address two lines later or so.

Implemented the change and it saves three gas, so I think we should leave it as it.

jparklev: Yes, SGTM. Thanks for checking

pkg/core/src/Periphery.sol
stevenvaleri: Implemented this updated in a separate PR that will merge into this branch #355

pkg/core/src/Periphery.sol
jparklev: > and we would still need to cast it back to an address two lines later or so

Although I think we could just do `quote.sellToken != stakeAddress` for that line, right?

(not to say we should still explore this change, just to confirm)

stevenvaleri:
> 

The caching is really creating additional variables in functions to reduce duplicate code execution. In one instance rather than repetitively casting an address to an ERC20, I added a variable to do that once and store the ERC20 in a variable

In another instance rather than calling the adapter for the target twice, its only called once and saved in a variable to be used again later.

pkg/core/src/Periphery.sol
stevenvaleri: Ah - confirmed. Good catch.

I retested taking that into account. Gas savings is still minimal, and increases in the min case, so I think should skip making this update.

image

stevenvaleri:
> I tried two other iterations of this change. The current PR both improves on gas and results in the smallest contract size.
> 
> One - the internal `balanceOf()` excepted an address rather than an ERC20 Two - the internal `balanceOf()` excepted an ERC20 and does not cache possible values Three (current) - the internal `balanceOf()` excepted an ERC20 and caches where possible
> 
> image

Closing this PR - the gas assessment table is incorrect. The values for the balanceOf with caching were not calculated the correct way making the balanceOf - with an address param the best solution.

This is the correct table:
image

The caching actually adds gas by creating an extra variable, and presumably the optimizer already removes the duplicate external calls and casting.

PR #356, opened 2023-04-26
Gen 2287 internal balance of with address param

stevenvaleri:
## Motivation

Reduce Periphery Contract Size

## Solution

Move external balanceOf calls all to one internal function.

Alternative approaches tried:
+ External Libraries for Logic - too complicated and more gas
+ Import interfaces instead of full contracts - already handled by compiler
+ move `safeTransfer` to an internal function - increases size
+ have internal balanceOf use an ERC20 as an input param instead of the token address

## Implementer Checklist

All logic remains the same. Refactor to reduce contract size.

Reduced Size:
image

Corrected and Accurate Gas Changes - `balanceOf - address` is this implementation:

image


jparklev (approved):
This makes sense and actually reads a little cleaner to me, ty for circling back on this

fedealconada (approved):
lgtm!

pkg/core/src/Periphery.sol
```diff
@@ -1062,6 +1074,10 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         }
     }
 
+    function _balanceOf(address token, address user) internal returns (uint256 amt) {
+        amt = ERC20(token).balanceOf(user);
+    }
+
```
IAm0x52, line 1080: Every occurrence of native balanceOf ERC20 method replaced with _balanceOf

IAm0x52 (commented):
Fixes look good. Every occurrence of ERC20.balance of changed to internal _balanceOf() function and inputs have been correctly translated

PR #358, opened 2023-05-01
fix(Periphery): rmv adapter only restriction, send target to caller, sponsor series interface

jparklev:
## Motivation

The implementation of the `swapYTsForTarget` function only enabled adapter addresses to call it – however, this meant that it couldn't be used by auto rollers (which are themselves not adapters), and this was a primary use case for it.

Context from chat:
Screen Shot 2023-05-01 at 12 53 36 AM

Screen Shot 2023-05-01 at 12 55 28 AM

Also: ensure that there's a `sponsorSeries` function as the Auto Rollers expect (see for context https://github.com/sense-finance/sense-v1/pull/360)

## Solution

1. Remove the `OnlyAdapter` restriction, and add a `_transfer` to send the swapped Target back to the caller
2. Create a new sponsorSeries function that keeps the compatibility with AutoRollers

## Implementer Checklist

### DIFF
--- a/pkg/core/src/Periphery.sol
+++ b/pkg/core/src/Periphery.sol
@@ -18,7 +18,6 @@ import { BaseAdapter as Adapter } from "./adapters/abstract/BaseAdapter.sol";
 import { BaseFactory as AdapterFactory } from "./adapters/abstract/factories/BaseFactory.sol";
 import { Divider } from "./Divider.sol";
 
-
 interface SpaceFactoryLike {
     function create(address, uint256) external returns (address);
 
@@ -323,10 +322,14 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         uint256 ytBal
     ) external returns (uint256 amt) {
         ERC20(divider.yt(adapter, maturity)).transferFrom(msg.sender, address(this), ytBal);
-        amt = this._swapYTsForTarget(msg.sender, adapter, maturity, ytBal, block.timestamp, PermitData(
-            IPermit2.PermitTransferFrom(IPermit2.TokenPermissions(ERC20(address(0)), 0), 0, 0),
-            "0x"
-        ));
+        amt = this._swapYTsForTarget(
+            msg.sender,
+            adapter,
+            maturity,
+            ytBal,
+            block.timestamp,
+            PermitData(IPermit2.PermitTransferFrom(IPermit2.TokenPermissions(ERC20(address(0)), 0), 0, 0), "0x")
+        );
         _transfer(ERC20(Adapter(adapter).target()), msg.sender, amt);
     }
 
@@ -616,7 +619,8 @@ contract Periphery is Trust, IERC3156FlashBorrower {
             revert Errors.SwapTooSmall();
 
         // Transfer YTs into this contract if needed
-        if (sender != address(this) && msg.sender != address(this)) _transferFrom(permit, divider.yt(adapter, maturity), ytBal);
+        if (sender != address(this) && msg.sender != address(this))
+            _transferFrom(permit, divider.yt(adapter, maturity), ytBal);
 
         // Calculate target to borrow by calling AMM
         uint256 targetToBorrow;
--- a/pkg/core/src/Periphery.sol
+++ b/pkg/core/src/Periphery.sol
@@ -322,15 +322,25 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         uint256 ytBal
     ) external returns (uint256 amt) {
         ERC20(divider.yt(adapter, maturity)).transferFrom(msg.sender, address(this), ytBal);
-        amt = this._swapYTsForTarget(
+        amt = this.swapYTsForTargetHelper(
             msg.sender,
             adapter,
             maturity,
             ytBal,
-            block.timestamp,
             PermitData(IPermit2.PermitTransferFrom(IPermit2.TokenPermissions(ERC20(address(0)), 0), 0, 0), "0x")
         );
         _transfer(ERC20(Adapter(adapter).target()), msg.sender, amt);
+    }
+
+    function swapYTsForTargetHelper(
+        address sender,
+        address adapter,
+        uint256 maturity,
+        uint256 ytBal,
+        PermitData calldata permit
+    ) external returns (uint256 amt) {
+        if (msg.sender != address(this)) revert Errors.OnlyPeriphery();
+        amt = _swapYTsForTarget(sender, adapter, maturity, ytBal, block.timestamp, permit);
     }
 
     function _swapSenseToken(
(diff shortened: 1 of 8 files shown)


jparklev:
I feel that i'm missing the reason for the `OnlyAdapter` restriction introduced here https://github.com/sense-finance/sense-v1/commit/7e68b59f894cdb52a41db9e9c6532083e2a62e7f, so waiting for @fedealconada to set me straight in case there's something i've missed

pkg/core/src/Periphery.sol
```diff
@@ -321,12 +322,12 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         uint256 maturity,
         uint256 ytBal
     ) external returns (uint256 amt) {
-        if (msg.sender != adapter) revert Errors.OnlyAdapter();
-        PermitData memory permit = PermitData(
+        ERC20(divider.yt(adapter, maturity)).transferFrom(msg.sender, address(this), ytBal);
+        amt = this._swapYTsForTarget(msg.sender, adapter, maturity, ytBal, block.timestamp, PermitData(
```
jparklev: At first I didn't understand why we were doing `this._swapYTsForTarget`, but now I understand it's to enter a new context in which `PermitData` could be used as `calldata` instead of `memory`. 

While this is kind of inefficient (cheaper to just switch `_swapYTsForTarget` to accept `memory`), I understand that we'd rather keep the other functions using it cheaper rather than make this one cheaper and the others slightly more expensive. Does that sound right to everybody?

pkg/core/src/Periphery.sol
```diff
@@ -615,7 +616,7 @@ contract Periphery is Trust, IERC3156FlashBorrower {
             revert Errors.SwapTooSmall();
 
         // Transfer YTs into this contract if needed
-        if (sender != address(this)) _transferFrom(permit, divider.yt(adapter, maturity), ytBal);
+        if (sender != address(this) && msg.sender != address(this)) _transferFrom(permit, divider.yt(adapter, maturity), ytBal);
```
jparklev: Had to add this, otherwise `_transferFrom` was trying to transfer from the periphery to itself (since `_swapYTsForTarget` is being called with `this.`, making the periphery the `msg.sender`)

pkg/core/src/Periphery.sol
stevenvaleri: Yes - your assessment is accurate. Switching PermitData to memory has a down stream effect of requiring _transferFrom to also use PermitData as a memory param. Overall, this increases gas of the functions issue, removeLiquidity, sponsorSeries, swapForPTs, swapForYTs, swapPTs, swapYTs by about 1000 each. Based on gas cost, leaving the calldata param is better.

The current downside is that we have `_swapYTsForTarget` as a public function. We can consider updates here separately.

pkg/core/src/Periphery.sol
stevenvaleri: Makes sense. The transfer is already covered in the first line of `swapYTsForTarget`.

I explored removing the transferFrom in the first line of swapYTsForTarget, and relying on this transfer instead, to reduce code size, but the changes would be too substantial and costly on gas.

pkg/core/src/Periphery.sol
fedealconada: totally! @stevenvaleri 's assessment was, as far as I remember, exactly the same reasoning I had when doing this changes

fedealconada (approved):
lgtm! thanks for taking this over @jparklev ! and special thanks to @alexfertel for find this bug just in time!! 🙌🏻

stevenvaleri (approved):

pkg/core/src/Periphery.sol
```diff
@@ -321,12 +348,26 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
+    }
+
+    function swapYTsForTargetHelper(
+        address sender,
+        address adapter,
+        uint256 maturity,
+        uint256 ytBal,
+        PermitData calldata permit
+    ) external returns (uint256 amt) {
+        if (msg.sender != address(this)) revert Errors.OnlyPeriphery();
```
IAm0x52, line 369: New external wrapper for _swapYTsForTarget

pkg/core/src/Periphery.sol
```diff
@@ -126,6 +126,23 @@ contract Periphery is Trust, IERC3156FlashBorrower {
     /// @param adapter Adapter to associate with the Series
     /// @param maturity Maturity date for the Series, in units of unix time
     /// @param withPool Whether to deploy a Space pool or not (only works for unverified adapters)
```
IAm0x52, line 128: More or less a version that's for use without permits. Presumably to support existing autorollers that can't use permits

pkg/core/src/Periphery.sol
```diff
@@ -152,18 +191,6 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
-
-        // refund any remaining quote.sellToken to receiver
-        _transfer(
-            quote.sellToken,
-            msg.sender,
-            address(quote.sellToken) == ETH
-                ? address(this).balance
-                : _balanceOf(address(quote.sellToken), address(this))
-        );
-
```
IAm0x52, line 166: Above logic only useful for permits so logic is moved to permit sponsor. Still basically the same idea

pkg/core/src/Periphery.sol
```diff
@@ -608,14 +649,15 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
+    ) internal returns (uint256 tBal) {
         // Because there's some margin of error in the pricing functions here, smaller
         // swaps will be unreliable. Tokens with more than 18 decimals are not supported.
         if (ytBal * 10**(18 - ERC20(divider.yt(adapter, maturity)).decimals()) <= MIN_YT_SWAP_IN)
             revert Errors.SwapTooSmall();
 
         // Transfer YTs into this contract if needed
-        if (sender != address(this)) _transferFrom(permit, divider.yt(adapter, maturity), ytBal);
+        if (sender != address(this) && msg.sender != address(this))
+            _transferFrom(permit, divider.yt(adapter, maturity), ytBal);
```
IAm0x52, line 661: Need to confirm that transferFrom can be skipped in every context:

_swapSenseToken - msg.sender != this && sender == msg.sender (uses permit)
swapYTsForTarget (same as swapYTsForTargetHelper)
_addLiquidity - sender == this (doesn't use permit)
swapYTsForTargetHelper - msg.sender == this (doesn't use permit)

IAm0x52: _swapSenseToken - doesn't transfer beforehand so always needs permit (correct)
swapYTsForTargetHelper - transfers in beforehand so never needs permit (correct)
_addLiquidity - addLiquidity does the transfer so no permit needed (correct)

IAm0x52 (approved):
Fixes looks good. Sponsor series is now compatible with existing autorollers and swapYTsForTarget is now accessible to anyone (autorollers) instead of just adapters

PR #359, opened 2023-05-02
feat: make swapYTsForTarget helpers not succeed from external calls

stevenvaleri:
## Motivation

Make code internal intended to be supporting logic to remove unnecessary external access. Previously `_swapYTsForTarget` was an external function as a work around to reduce contract size and maintain gas efficiency of having PermitData as a `calldata` param instead of a `memory` param.

## Solution

Make _swapYTsForTarget() an internal function.

Add an additional helper function for the external `swapYTsForTarget` with an onlyThis modifier. This allows the `swapYTsForTarget` to use the external helper function to then make use of the internal `_swapYTsForTarget` function via the helper, without making PermitData a memory param in `_swapYTsForTarget`.

Since the PermitData value in swapYTsForTarget is in memory, the internal function _swapYTsForTarget cannot be called directly without updating the _swapYTsForTarget to have PermitData as a memory param.

## Implementer Checklist

Added additional test to confirm access validation via the `onlyThis` modifier.

### DIFF
--- a/pkg/core/src/Periphery.sol
+++ b/pkg/core/src/Periphery.sol
@@ -338,7 +338,7 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         uint256 maturity,
         uint256 ytBal,
         PermitData calldata permit
-    ) external returns (uint256 amt){
+    ) external returns (uint256 amt) {
         if (msg.sender != address(this)) revert Errors.OnlyPeriphery();
         amt = _swapYTsForTarget(sender, adapter, maturity, ytBal, block.timestamp, permit);
     }
--- a/pkg/core/src/Periphery.sol
+++ b/pkg/core/src/Periphery.sol
@@ -338,7 +338,8 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         uint256 maturity,
         uint256 ytBal,
         PermitData calldata permit
-    ) external onlyThis returns (uint256 amt){
+    ) external returns (uint256 amt){
+        if (msg.sender != address(this)) revert Errors.OnlyPeriphery();
         amt = _swapYTsForTarget(sender, adapter, maturity, ytBal, block.timestamp, permit);
     }
 
@@ -1103,13 +1104,6 @@ contract Periphery is Trust, IERC3156FlashBorrower {
 
     // required for refunds
     receive() external payable {}
-
-    /* ========== MODIFIERS ========== */
-
-    modifier onlyThis() {
-        if (msg.sender != address(this)) revert Errors.OnlyPeriphery();
-        _;
-    }
 
     /* ========== LOGS ========== */
 
--- a/pkg/core/src/Periphery.sol
+++ b/pkg/core/src/Periphery.sol
@@ -322,15 +322,24 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         uint256 ytBal
     ) external returns (uint256 amt) {
         ERC20(divider.yt(adapter, maturity)).transferFrom(msg.sender, address(this), ytBal);
-        amt = this._swapYTsForTarget(
+        amt = this.swapYTsForTargetHelper(
             msg.sender,
             adapter,
             maturity,
             ytBal,
-            block.timestamp,
             PermitData(IPermit2.PermitTransferFrom(IPermit2.TokenPermissions(ERC20(address(0)), 0), 0, 0), "0x")
         );
         _transfer(ERC20(Adapter(adapter).target()), msg.sender, amt);
+    }
+
+    function swapYTsForTargetHelper(
+        address sender,
+        address adapter,
+        uint256 maturity,
+        uint256 ytBal,
+        PermitData calldata permit
+    ) external onlyThis returns (uint256 amt){
+        amt = _swapYTsForTarget(sender, adapter, maturity, ytBal, block.timestamp, permit);
     }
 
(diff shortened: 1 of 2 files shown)


stevenvaleri:
There is a minor contract size increase of: .028 (KiB)

Gas Differences:
image

pkg/core/src/Periphery.sol
```diff
@@ -322,17 +322,26 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
             adapter,
             maturity,
             ytBal,
-            block.timestamp,
             PermitData(IPermit2.PermitTransferFrom(IPermit2.TokenPermissions(ERC20(address(0)), 0), 0, 0), "0x")
         );
         _transfer(ERC20(Adapter(adapter).target()), msg.sender, amt);
     }
 
+    function swapYTsForTargetHelper(
```
stevenvaleri, line 335: Open to a different function name here

pkg/core/src/Periphery.sol
```diff
@@ -1095,6 +1104,13 @@ contract Periphery is Trust, IERC3156FlashBorrower {
     // required for refunds
     receive() external payable {}
 
+    /* ========== MODIFIERS ========== */
+
+    modifier onlyThis() {
```
stevenvaleri: The modifier is the same contract size as inlined revert. I think modifier is a little cleaner, but more of an opinion here.

pkg/core/src/Periphery.sol
fedealconada: i'm tempted to say we could just use the same name: `swapYTsForTarget` since they have different params but at the same time I can see someone on etherscan seeing 2 functions with the same name and that being confusing...

pkg/core/src/Periphery.sol
fedealconada: I feel like I would have put it inline just because it's used only once, but not really opinionated

fedealconada (approved):
answered your own questions but lgtm, thanks!

pkg/core/src/Periphery.sol
jparklev: Yea, I think i'm in support of `swapYTsForTargetHelper`, just since it's so specific and helper makes it a bit clearer that it's never meant to be called externally

pkg/core/src/Periphery.sol
jparklev: Seconded @fedealconada

pkg/core/src/tests/Periphery.tm.sol
```diff
@@ -944,6 +944,25 @@ contract PeripheryMainnetTests is PeripheryTestHelper {
...
+
+        // 3. Swap 10% of bob's YTs for Target
+        vm.startPrank(bob);
+        uint256 ytBalPre = ERC20(yt).balanceOf(bob);
+        uint256 targetBalPre = mockTarget.balanceOf(bob);
+
+        Periphery.PermitData memory data = generatePermit(bobPrivKey, address(periphery), address(yt));
+
+        vm.expectRevert(Errors.OnlyPeriphery.selector);
+        periphery.swapYTsForTargetHelper(bob, address(mockAdapter), maturity, ytBalPre / 10, data);
```
jparklev, line 962: One pattern I like for access checks is fuzzing impersonation addresses, and making sure all except for the expected one fail ([e.g.](https://github.com/sense-finance/auto-roller/blob/0b4f2264a12d68739a8692ccb44034df06b0e3d6/src/test/AutoRoller.t.sol#L160)). Especially since foundry's fuzzing dictionary [is smart](https://twitter.com/transmissions11/status/1604941957444210688?s=20) and won't just chose random addresses

Not something we have to do here, so more just to share in case it's useful

jparklev (approved):
LGTM once CI is passing!

jparklev:
Also, how are you copying those gas tables in @stevenvaleri?

pkg/core/src/tests/Periphery.tm.sol
stevenvaleri: Love that!

stevenvaleri:
> Also, how are you copying those gas tables in @stevenvaleri?

I'm running `npm run gas-report` and comparing the before and after in googledrive. which is not ideal and prone to error as we've seen.

I'd love to add an action that runs and automatically displays the the gas usage and contract size diff of the PR.

pkg/core/src/tests/Periphery.tm.sol
stevenvaleri: Added

pkg/core/src/Periphery.sol
stevenvaleri: moved to inline

jparklev:
> > Also, how are you copying those gas tables in @stevenvaleri?
> 
> I'm running `npm run gas-report` and comparing the before and after in googledrive. which is not ideal and prone to error as we've seen.
> 
> I'd love to add an action that runs and automatically displays the the gas usage and contract size diff of the PR.

Got it. Sounds like a great linear ticket to queue up in quality work 😈 , how does this look: https://linear.app/sense-finance/issue/GEN-2356/create-gh-action-that-displays-contract-size-and-gas-usage-diffs-in

jparklev:
Merging to get a clean PR for Sherlock to review, hope it's ok @stevenvaleri!

PR #360, opened 2023-05-03
GEN-2358: RLV compatiblity

fedealconada:
## Motivation

Make sure that Periphery v2 is backwards compatible with RLVs.

## Solution

`AutoRoller.sol` calls the following functions of `Periphery.sol`:
- [`periphery.MIN_YT_SWAP_IN()`](https://github.com/sense-finance/sense-v1/blob/dev/pkg/core/src/Periphery.sol#L36)
- [[`periphery.swapYTsForTarget(address(adapter), maturity, excessBal);`](](https://github.com/sense-finance/sense-v1/blob/dev/pkg/core/src/Periphery.sol#L319)url)
- [`periphery.spaceFactory()`](https://github.com/sense-finance/sense-v1/blob/dev/pkg/core/src/Periphery.sol#L61)
- `periphery.sponsorSeries(address(adapter), _maturity, true);`

While the first 3 exist on the Periphery v2, `sponsorSeries` does not exist with the signature mentioned in the example above (it has been modified). Hence, similar to what we did for `swapYTsForTarget`, we need to create a new `sponsorSeries` function which keeps the compatibility.

## Implementer Checklist

### DIFF
--- a/pkg/core/src/Periphery.sol
+++ b/pkg/core/src/Periphery.sol
@@ -322,15 +322,24 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         uint256 ytBal
     ) external returns (uint256 amt) {
         ERC20(divider.yt(adapter, maturity)).transferFrom(msg.sender, address(this), ytBal);
-        amt = this._swapYTsForTarget(
+        amt = this.swapYTsForTargetHelper(
             msg.sender,
             adapter,
             maturity,
             ytBal,
-            block.timestamp,
             PermitData(IPermit2.PermitTransferFrom(IPermit2.TokenPermissions(ERC20(address(0)), 0), 0, 0), "0x")
         );
         _transfer(ERC20(Adapter(adapter).target()), msg.sender, amt);
+    }
+
+    function swapYTsForTargetHelper(
+        address sender,
+        address adapter,
+        uint256 maturity,
+        uint256 ytBal,
+        PermitData calldata permit
+    ) external onlyThis returns (uint256 amt){
+        amt = _swapYTsForTarget(sender, adapter, maturity, ytBal, block.timestamp, permit);
     }
 
     function _swapSenseToken(
@@ -612,7 +621,7 @@ contract Periphery is Trust, IERC3156FlashBorrower {
         uint256 ytBal,
         uint256 deadline,
         PermitData calldata permit
-    ) public returns (uint256 tBal) {
+    ) internal returns (uint256 tBal) {
         // Because there's some margin of error in the pricing functions here, smaller
         // swaps will be unreliable. Tokens with more than 18 decimals are not supported.
         if (ytBal * 10**(18 - ERC20(divider.yt(adapter, maturity)).decimals()) <= MIN_YT_SWAP_IN)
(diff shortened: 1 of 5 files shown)


jparklev:
Assuming this should be compared to #358, so going to update the base for easier diffing

pkg/core/src/Periphery.sol
```diff
@@ -137,6 +153,30 @@ contract Periphery is Trust, IERC3156FlashBorrower {
...
+    }
+
+    function _sponsorSeries(
+        address adapter,
+        uint256 maturity,
+        bool withPool,
+        address stake,
+        uint256 stakeSize
+    ) internal returns (address pt, address yt) {
+        ERC20(stake).transferFrom(msg.sender, address(this), stakeSize);
```
jparklev: Could this double transfer `stake` for when calling the permit `sponsorSeries`? (once from `_transferFrom(permit, address(quote.sellToken), quote.amount);` and again here?)

Maybe we can just move this line directly into the non-permit `sponsorSeries`?

jparklev (commented):
Other than the comment on the double transfer, this approach seems OK to me. Looks like https://github.com/sense-finance/auto-roller/pull/38 is passing with this, which is a positive sign

Much appreciated @fedealconada for jumping on this, and good shout suggesting the test @stevenvaleri. After this is all done, let's do a quick RCA together to make note of any things we can learn dev-wise from the periph stuff. In the meantime, we found this just in time to fit it in with the other sherlock re-review items :)

And looks like we're still in the limit 🤞 (using the currently set 100 optimizer runs)
Screen Shot 2023-05-04 at 12 20 26 AM

pkg/core/src/Periphery.sol
fedealconada: Totally! My bad, was in rush yesterday but good catch. Definitely was meant to be on the non-permit `sponsorSeries()` as you suggest :)

fedealconada:
> Assuming this should be compared to #358, so going to update the base for easier diffing

yes!

fedealconada:
> Other than the comment on the double transfer, this approach seems OK to me. Looks like [sense-finance/auto-roller#38](https://github.com/sense-finance/auto-roller/pull/38) is passing with this, which is a positive sign
> 
> Much appreciated @fedealconada for jumping on this, and good shout suggesting the test @stevenvaleri. After this is all done, let's do a quick RCA together to make note of any things we can learn dev-wise from the periph stuff. In the meantime, we found this just in time to fit it in with the other sherlock re-review items :)
> 

agree!

> And looks like we're still in the limit 🤞 (using the currently set 100 optimizer runs) Screen Shot 2023-05-04 at 12 20 26 AM

glad to see this!

stevenvaleri (approved):
Change looks good to me. We should add a test for the new sponsorSeries function. Since it shares the same logic as the other function with a test, it is non-blocking for merge, but we should add.

fedealconada:
> Change looks good to me. We should add a test for the new sponsorSeries function. Since it shares the same logic as the other function with a test, it is non-blocking for merge, but we should add.

pushed!

PR #364, opened 2023-07-20
chore: polish Periphery v2 deployment script

fedealconada:
## Motivation

Improve deployment script for Periphery V2 and add sanity checks

## Solution

Deployment script has been updated to include some sanity checks (basically, checking that we can `roll()` on the new Periphery).

Friendly reminder on how to run the deployment script in fork mode:
```
cd /sense-v1/pkg/deployments
yarn task 20230301-periphery-v2 --network hardhat
```

## Implementer Checklist

### DIFF
--- a/pkg/deployments/tasks/20230301-periphery-v2/index.js
+++ b/pkg/deployments/tasks/20230301-periphery-v2/index.js
@@ -2,9 +2,11 @@ const { task } = require("hardhat/config");
 const data = require("./input");
 
 const { SENSE_MULTISIG, CHAINS, VERIFY_CHAINS } = require("../../hardhat.addresses");
-const { verifyOnEtherscan } = require("../../hardhat.utils");
+const { verifyOnEtherscan, fund, generateTokens } = require("../../hardhat.utils");
+const { startPrank } = require("../../hardhat.utils");
 
 task("20230301-periphery-v2", "Deploys and authenticates Periphery V2").setAction(async (_, { ethers }) => {
+  const { deploy } = deployments;
   const { deployer } = await getNamedAccounts();
   const chainId = await getChainId();
   const deployerSigner = await ethers.getSigner(deployer);
@@ -13,6 +15,8 @@ task("20230301-periphery-v2", "Deploys and authenticates Periphery V2").setActio
   const { abi: peripheryAbi } = await deployments.getArtifact("Periphery");
   const { abi: rlvFactoryAbi } = await deployments.getArtifact("AutoRollerFactory");
   const { abi: rlvAbi } = await deployments.getArtifact("AutoRoller");
+  const { abi: adapterAbi } = await deployments.getArtifact("BaseAdapter");
+  const { abi: erc20Abi } = await deployments.getArtifact("solmate/src/tokens/ERC20.sol:ERC20");
 
   if (!SENSE_MULTISIG.has(chainId)) throw Error("No balancer vault found");
   const senseAdminMultisigAddress = SENSE_MULTISIG.get(chainId);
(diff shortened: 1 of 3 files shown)


pkg/deployments/tasks/20230301-periphery-v2/index.js
```diff
@@ -68,77 +76,109 @@ task("20230301-periphery-v2", "Deploys and authenticates Periphery V2").setActio
...
+      await newPeriphery.onboardAdapter(adapter, false).then(t => t.wait());
+      console.log(`- Adapter ${adapter} onboarded`);
+    } else {
+      console.log(`- Skipped adapter ${adapter} verification as it's guard is ${params[2]}`);
+    }
+  }
+
+  // Verify adapters
+  for (let adapter of verifiedDifference) {
+    console.log("\nOnboarding adapter", adapter);
```
stevenvaleri: "Verifying Adapter" instead of onboarding in this step

pkg/deployments/tasks/20230301-periphery-v2/index.js
```diff
@@ -200,30 +220,56 @@ task("20230301-periphery-v2", "Deploys and authenticates Periphery V2").setActio
...
+    console.log("\nUpdate Periphery address on existing RLVs");
+    const rlvs = [
+      ...new Set(
+        (await rlvFactory.queryFilter(rlvFactory.filters.RollerCreated(null))).map(e => e.args.autoRoller),
+      ),
+    ];
+
+    console.log("RLVs to update:", rlvs);
+    for (let rlvAddress of rlvs) {
+      const rlv = new ethers.Contract(rlvAddress, rlvAbi, multisigSigner);
```
stevenvaleri, line 242: Around here, i'm getting a message:

`Duplicate definition of ParamChanged (ParamChanged(bytes32,address), ParamChanged(bytes32,uint256))`

My abi's might be off from an old compile. I don't think it would cause an issue, curious if you have this though.

pkg/deployments/tasks/20230301-periphery-v2/index.js
```diff
@@ -200,30 +220,56 @@ task("20230301-periphery-v2", "Deploys and authenticates Periphery V2").setActio
...
+      }
+    }
 
     oldPeriphery = oldPeriphery.connect(multisigSigner);
     divider = divider.connect(multisigSigner);
 
     if (senseAdminMultisigAddress.toUpperCase() !== deployer.toUpperCase()) {
       console.log("\nUnset the multisig as an authority on the old Periphery");
-      await oldPeriphery.setIsTrusted(senseAdminMultisigAddress, false).then(t => t.wait());
+      if (await oldPeriphery.isTrusted(senseAdminMultisigAddress)) {
```
stevenvaleri: Are there risks to removing the multi-sig at this point? If something were to go wrong in this process, would there be benefit to delaying the removal of the multi-sig as admin until we confirm the migration to the new periphery has been 100% successful?

pkg/deployments/tasks/20230301-periphery-v2/index.js
```diff
@@ -248,6 +287,64 @@ task("20230301-periphery-v2", "Deploys and authenticates Periphery V2").setActio
...
+    // we skip this RLV from the sanity test
+    const firstDeposit = ethers.utils.parseEther("1");
+    let skip = false;
+    try {
+      await generateTokens(targetAddress, deployer, deployerSigner, firstDeposit);
+    } catch (e) {
+      skip = true;
+    }
+    if (skip) {
+      console.log(`- Skipping RLV ${rlv.address} as we couldn't generate target tokens for it`);
```
stevenvaleri, line 330: 3 / 8 successfully rolled, the rest are skipped for this reason. Is there another way to easily generate the tokens so that these can be tested - maybe sending from a whale's address. Likely low risk here though given there are successful rolls.

stevenvaleri (changes requested):
Overall looks good - a couple of questions / thoughts that may not need code changes.

pkg/deployments/tasks/20230301-periphery-v2/index.js
fedealconada: I have seen this before but not experiencing it right now

>  I don't think it would cause an issue
agree

pkg/deployments/tasks/20230301-periphery-v2/index.js
fedealconada: I agree this can wait!

pkg/deployments/tasks/20230301-periphery-v2/index.js
fedealconada: Yeah, that could be a solution. Or sometimes is just generating tokens for the underlying of the token wanted and then wrapping or things like that. I didn't want to spend that much time and thought that having just 2-3 working cases was fine. 
The idea was more to do an E2E test making sure that the change of Periphery was working fine with the existing RLVs.

stevenvaleri (approved):
Looks good!

PR #365, opened 2023-10-15
GEN-2786: sunset script

fedealconada:
## Motivation

Sunset V1

## Solution

The script gets all the contracts we've ever deployed from Sense's admin addresses and:
1. removes ourselves from being a rewards recipient (for contracts with `ExtractableReward.sol`
2. removes the trust to these addresses (for contracts with `Trust.sol`

Contracts with the deployer address trusted don't need to be executed via multisig.
For all the rest of the transactions, the script will output a JSON file with all the txs and its data to execute them using the Gnosis Transaction Builder (batched).


pkg/deployments/tasks/20231031-sunset/index.js
```diff
@@ -0,0 +1,404 @@
...
+    return `${acc}}`;
+  }
+
+  return `${JSON.stringify(json, stringifyReplacer)}`;
+};
+
+const calculateChecksum = batchFile => {
+  const serialized = serializeJSONObject({
+    ...batchFile,
+    meta: { ...batchFile.meta, name: null },
```
jparklev: why does the name have to be set as null?

pkg/deployments/tasks/20231031-sunset/index.js
```diff
@@ -0,0 +1,404 @@
...
+      if (isContract) {
+        contractAddresses = await Promise.all(
+          transactions
+            .filter(
+              tx =>
+                tx.isError === "0" &&
+                (tx.type === "create" || tx.type === "create2") &&
+                tx.contractAddress !== address,
+            )
+            .map(async tx => {
```
jparklev: tiny nit: doesn't need to be async if we're not using `getContractName`

pkg/deployments/tasks/20231031-sunset/index.js
```diff
@@ -0,0 +1,404 @@
...
+    const trusted = await getTrustedAddresses(contract);
+    if (trusted.sense.length > 1) {
+      throw new Error(`   * More than one trusted address: ${trusted.sense}`);
+    }
+
+    // set reward recipient to zero address
+    if (Object.values(SENSE_ADMIN_ADDRESSES).includes((await contract.rewardsRecipient()).toLowerCase())) {
+      // some contracts were never changed to be trusted by the multisig and remained trusted by the deployer
+      if (!(await contract.isTrusted(SENSE_ADMIN_ADDRESSES.multisig))) {
+        contract = contract.connect(deployerSigner);
```
jparklev: It seems like we update the connected contract address, but then it's never used (due to the `if else`)

pkg/deployments/tasks/20231031-sunset/index.js
```diff
@@ -0,0 +1,404 @@
...
+    if (includesDeployer) contract = contract.connect(deployerSigner);
+    if (isFork && includesDeployer) await stopPrank(SENSE_ADMIN_ADDRESSES.multisig); // stop pranking
+
+    for (const address of trusted.sense) {
+      if (includesDeployer && address === SENSE_ADMIN_ADDRESSES.deployer.toLowerCase()) continue;
+      if (!includesDeployer && address === SENSE_ADMIN_ADDRESSES.multisig.toLowerCase()) continue;
+
+      await contract.setIsTrusted(address, false, { gasPrice: boostedGasPrice() }).then(t => t.wait());
+
+      if (address.toLowerCase() === SENSE_ADMIN_ADDRESSES.multisig.toLowerCase()) {
```
jparklev: So this will only trigger if `address` is the `SENSE_ADMIN_ADDRESSES`, and the address list _does_ include the deployer, right? Since line 188 will `continue` if the address list doesn't include the deployer. 

I might be reading it wrong, but i'd expect this to trigger only if the address list doesn't include the deployer, since otherwise we don't need to use multisig, right?

pkg/deployments/tasks/20231031-sunset/index.js
```diff
@@ -0,0 +1,404 @@
...
+      return;
+    }
+    // if deployer is trusted, we can call setIsTrusted for all the addresses directly from this script
+    const includesDeployer = trusted.sense.includes(SENSE_ADMIN_ADDRESSES.deployer.toLowerCase());
+    if (includesDeployer) contract = contract.connect(deployerSigner);
+    if (isFork && includesDeployer) await stopPrank(SENSE_ADMIN_ADDRESSES.multisig); // stop pranking
+
+    for (const address of trusted.sense) {
+      if (includesDeployer && address === SENSE_ADMIN_ADDRESSES.deployer.toLowerCase()) continue;
+      if (!includesDeployer && address === SENSE_ADMIN_ADDRESSES.multisig.toLowerCase()) continue;
```
jparklev: does address need to be `toLowerCase` like line 192

pkg/deployments/tasks/20231031-sunset/index.js
```diff
@@ -0,0 +1,404 @@
...
+      }
+    }
+
+    // last call with corresponding remaining address
+    if (includesDeployer) {
+      await contract
+        .setIsTrusted(SENSE_ADMIN_ADDRESSES.deployer, false, { gasPrice: boostedGasPrice() })
+        .then(t => t.wait());
+    } else {
+      await contract
```
jparklev: Just to make sure I understand – this will fail on mainnet, right? Since we can't use the multisig signer. But we run it on hardhat and get the multisig txs list there? And do we also run this script on mainnet to get the deployer txs out

jparklev (commented):
A few notes & questions, mostly on the `removeTrust` paths, but overall this is a nice approach!!

Pending resolution on the RLV adapter depreciation discussion and the review comments, LGTM

pkg/deployments/tasks/20231031-sunset/index.js
```diff
@@ -0,0 +1,404 @@
...
+  // (1). Turn permissionless mode ON
+  if (!(await divider.permissionless())) {
+    await divider.setPermissionless(true, { gasPrice: boostedGasPrice() }).then(t => t.wait());
+    multisigTxs.push(
+      getTx(
+        dividerAddress,
+        "setPermissionless",
+        divider.interface.encodeFunctionData("setPermissionless", [true]),
+      ),
+    );
```
stevenvaleri: Does this call it both directly here and add to the multi-sig tx array? Seems like it's one or the other in the other scenarios.

stevenvaleri (approved):
Two high-level items that may make sense:

1. Add test scripts to confirm the action results.

2. Separate the remove rewards recipient step and the remove trust step, that way we can remove the recipients, confirm they are removed, and then remove the trust separately. I think it's quite unlikely, but maybe if there were an error in removing the reward recipient, but we remove the trust anyway, the reward recipient it could be stuck.

jparklev:
Also @fedealconada since you're off wed – once you've addressed these comments and are happy with it, feel free to start running the script and generating the multisig json :)

(since looks like we have a resolution on what to do w the RLVs in #ops-internal ~)

fedealconada:
> Two high-level items that may make sense:
> 
> 1. Add test scripts to confirm the action results.
> 2. Separate the remove rewards recipient step and the remove trust step, that way we can remove the recipients, confirm they are removed, and then remove the trust separately. I think it's quite unlikely, but maybe if there were an error in removing the reward recipient, but we remove the trust anyway, the reward recipient it could be stuck.

good point, it's always good to be safe. just a q: you are saying we run the batch of txs on Gnosis, then sort of run a sanitiy check to ensure everything is right, and then prepare another batch of txs for the trust removal?

i've now added (2) and also sanity checks but still generating one batch tx with everything in it.

pkg/deployments/tasks/20231031-sunset/index.js
fedealconada: not sure i follow you, the divider is trusted by multisig, so we need to execute this function from it by adding it to the batch

pkg/deployments/tasks/20231031-sunset/index.js
fedealconada: haha, good call, i've just fixed it so we can run it on mainnet (for the deployer ones and skipping the multisig ones)

pkg/deployments/tasks/20231031-sunset/index.js
fedealconada: i've polished all the `toLowerCase()`

pkg/deployments/tasks/20231031-sunset/index.js
fedealconada: i think this was a bit "complex" the way it as. I've simplified it by addin logic to the `getTrustedAddresses()` so it now returns the trusted addresses array but placing the multisig and deployer (if they are trusted) at the end of the array. If both are trusted, deployer address must be the last one. So, then, when iterating over it, we leave the corresponding deployer to remove trust to itself in last place.

pkg/deployments/tasks/20231031-sunset/index.js
fedealconada: it was just a copy paste to be honest

fedealconada:
> Also @fedealconada since you're off wed – once you've addressed these comments and are happy with it, feel free to start running the script and generating the multisig json :)
> 
> (since looks like we have a resolution on what to do w the RLVs in #ops-internal ~)

you mean i should already run the script to do the `deployer` actions now? just to clarify because i think @KentonPrescott said to wait until Friday (btw, I'll be on PTO but can execute this any day)

jparklev:
> you mean i should already run the script to do the `deployer` actions now?

Yes! Oh, yea I think kenton just meant that we would announce on friday, and he was wanting to get your help before you went on PTO. But! If you'll be able to execute any day it's all good!

fedealconada:
> > you mean i should already run the script to do the `deployer` actions now?
> 
> Yes! Oh, yea I think kenton just meant that we would announce on friday, and he was wanting to get your help before you went on PTO. But! If you'll be able to execute any day it's all good!

nah, i can run it today then, will do and let you know!

pkg/deployments/tasks/20231031-sunset/index.js
stevenvaleri: My confusion is on line 53 - where the divider is called directly, before adding it to the multi-sig txs. Is that only intended for the instance when we are running this on a fork?

stevenvaleri:
> just a q: you are saying we run the batch of txs on Gnosis, then sort of run a sanitiy check to ensure everything is right, and then prepare another batch of txs for the trust removal?

yes - run a batch to remove reward recipients, sanity check, run a batch to remove trust

pkg/deployments/tasks/20231031-sunset/index.js
fedealconada: ohh, yes yes, it was only to test on fork, i've already included the changes to skip if run on mainnet (otherwise it would fail)

